PR #11, opened 2017-09-19
Set units to the timeline when displayed in seconds

roperzh:
Found this while solving tour issues and being a simple fix, why not?

**What**

[Trello Card](https://trello.com/c/HbCkNxo8/425-on-timeline-at-certain-zoom-level-when-displaying-as-seconds-milliseconds-are-displayed-but-without-the-unit)

**Result**

screen shot 2017-09-19 at 10 54 46 am
screen shot 2017-09-19 at 10 54 56 am

@matthewtoast added you as a reviewer because in the card wasn't clear if we also wanted to add the units when the time is displayed in seconds. I added the unit just for consistency, but let me know if you prefer to remove the little `s`.


tsortwehttam:
This looks good to my eye (and I think it's clearer with the unit shown) - but we should defer to Zack/Taylor. I'll assign them as reviewers

taylorpoe:
Makes sense to me. I'll go ahead and merge it unless @zackbrown has concerns.

zackbrown:
go for it.

PR #14, opened 2017-09-22
Fix: disable drag of collapsed segments

taylorpoe:
This change disables collapsed transition segments from being able to be dragged — now when a user tries to "grab" and move a collapsed segment, it will simply expand the collapsed element (or if collapsed property, the property).


taylorpoe:
It appears that perhaps the `git pull` I did before adding this commit wasn't up-to-date. I'm going to close this PR and re-do it real quick.

PR #15, opened 2017-09-23
fix: disable dragging of collapsed segments

taylorpoe:
This disables the ability to drag transition segments from collapsed levels; i.e., at either property or element level. If a user attempts a drag it will instead expand the row.

### DIFF
--- a/packages/haiku-timeline/src/components/Timeline.js
+++ b/packages/haiku-timeline/src/components/Timeline.js
@@ -1669,6 +1669,7 @@ class Timeline extends React.Component {
         key={`${propertyName}-${index}`}
         axis='x'
         onStart={(dragEvent, dragData) => {
+          if (options.collapsed) return false
           this.setRowCacheActivation({ componentId, propertyName })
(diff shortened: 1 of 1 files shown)


tsortwehttam:
@taylorpoe Before merging, I have a (hopefully small?) follow-up request to this.

Can you also prevent other actions such as:

- Keyframes being created (via right-click) on collapsed rows -- _or just disable the right-click menu entirely when the row is collapsed_
- Individual keyframes from being dragged
- Keyframe endpoints from being dragged
- etc

On the collapsed rows?

taylorpoe:
Yeah will do. I'm finishing up some design writing atm, but will get to fixing this next before moving on to the project loading reticulator.

taylorpoe:
It appeared that the following were already not possible on collapsed rows:
* Keyframes being created (via right-click)
* Individual keyframes from being dragged
* Keyframe endpoints from being dragged
etc

It was possible to add a transition to a constant segment via right click; so I disabled that.

PR #16, opened 2017-09-23
Fix plumbing methods broken since ws upgrade to v3

roperzh:
In 61fa5bac the `ws` package got upgraded to v3, which comes with
a breaking change to the `connect` API. This updates our consumers
to the new API, fixing a crash when initializing the app.

More details on `ws` v3 breaking changes [here](https://github.com/websockets/ws/pull/1104)


tsortwehttam:
Thank you for this. I was just about to do this myself. :) 

The reason for this upgrade - in case future developers find this - is: Through various work we were doing since our last distro push, our project ended up with multiple versions of 'ws' installed across various sub-projects (both by virtue of 'ws' itself, as well as modules that were once-removed from 'ws', such as a legacy install of 'peerjs'). The 'ws' package, it turns out, depends on several native modules, one of which is called 'bufferutil'. Different versions of 'bufferutil' are extremely picky about what node (platform) versions they can compile on, and that pickiness was preventing our distro pipeline from being able to successfully complete the 'electron-rebuild' step without an exception. For context on that, 'electron-rebuild' is used to compile native modules against a specified electron version, which is required whenever you want to ship an electron app that includes native modules. (Without 'electron-rebuild', we wouldn't be able to use nodegit.) So, I upgraded 'ws' to 3.x, which resolved all the aforementioned issues - but which, of course, also had this breaking change, which I missed!

🌮

PR #22, opened 2017-09-26
tweak: update starter sketchfile with instructions

taylorpoe:
I'm PRing this because now because it may be holding up the deploy. I'm having trouble running mono right now, so I haven't verified this working in development, but having only replaced the file — I don't anticipate it causing any problems. Will keep trying to verify here on my end.


roperzh:
Checked out the branch and tested it, everything looks good to me.

screen shot 2017-09-26 at 6 35 31 pm

taylorpoe:
@roperzh Thanks for doing that!

PR #25, opened 2017-09-27
Auto Updates

roperzh:
**What**

[Trello Card](https://trello.com/c/0rWi3YXy/705-stabilize-auto-update)
[Specs](https://haiku.quip.com/6hYmAzGFNyCi)


roperzh:
@matthewtoast this is ready to merge. I tested it as many times as possible with different workflows, but it surely deserves extra testing with a staging build. If you think it's good to merge, feel free to go for it!

roperzh:
Added a bit of refactor and more tests.

packages/haiku-creator/src/utils/autoUpdate.js
```diff
@@ -1,63 +1,74 @@
...
+//   branch: process.env.HAIKU_RELEASE_BRANCH,
+//   platform: process.env.HAIKU_RELEASE_PLATFORM,
+//   version: process.env.HAIKU_RELEASE_VERSION
+// }
 
-  const feedURL = `${opts.server}/updates/latest?environment=${opts.environment}&branch=${opts.branch}&platform=${opts.platform}&version=${opts.version}`
+module.exports = {
+  async update (url, progressCallback, options = opts) {
+    return new Promise(async (resolve, reject) => {
+      if (process.env.HAIKU_SKIP_AUTOUPDATE !== '1') {
```
roperzh, line 27: @matthewtoast a heads up on this check for staging/prod builds. Let me know if you think of a better way to handle it.

tsortwehttam: IIRC this flag was so that we could test (run) staging builds without needing to configure the auto-update variables. I'm a bit fuzzy on the old logic, but this looks ok to me.

PR #33, opened 2017-10-05
[Needs Discussion] Skip initial builds of babel watchers

roperzh:
## What

Every time we run `yarn start` all files get re-compiled because that's what `babel --watch` does by default. This PR proposes to avoid that initial compilation by providing the `--skip-initial-build` argument by default to our watchers.

*pros*
- Faster initial time

*cons*
- If a developer makes a change _before_ starting the server, that change will not be compiled when starting the app. This applies to git pulls. A possible solution for this will be to provide another option to `yarn start` to explicitly skip the initial build or to remember to run `yarn compile` in this scenario.

### DIFF
--- a/packages/haiku-creator/package.json
+++ b/packages/haiku-creator/package.json
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "start": "electron ./lib/electron.js",
-    "watch": "babel --watch --skip-initial-build src --out-dir lib --source-maps inline",
+    "watch": "babel --watch src --out-dir lib --source-maps inline",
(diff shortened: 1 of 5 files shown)


zackbrown:
The `git pull`s issue seems high-friction to me.  Perhaps `yarn start --fast` would be a nice way to opt in to the `--skip-initial-build` flags

roperzh:
That sounds completely reasonable, I have added `yarn start -- --preset=fast` as an option to initialize the app.

I'm not sure if the implementation is the best, so I'm open to suggestions :)

tsortwehttam:
LGTM. We can refine as time goes on.

PR #34, opened 2017-10-05
Fix overlapping  between 'Back' curves and props

roperzh:
This issue was happening only with _closed_ property rows and it
was a mixture of z-index issues and missing background/positioning.

Fortunately, _open_ property rows works as expected, so it was a
matter of copying their styles.

<a href="https://trello.com/c/iR13Y8XU/780-timeline-transition-segments-got-z-index-funky-vs-properties-panel"> Timeline transition segments got z-index funky vs. properties panel</a>


taylorpoe (approved):
Lookin' z-tastic!

PR #37, opened 2017-10-11
Drag and Drop of external files

roperzh:
**What**

[Trello card](https://trello.com/c/MAss2Gvx/734-dragging-dropping-svg-file-into-creator-navigates-electron-to-that-svg-file-effectively-crashes-app)

**Comments**

- Despite the branch name, this implementation handles Sketch and SVG files.
- If the user tries to drop a file that is not supported we will just ignore that, without raising any errors. This is intentional to copycat the behavior of Sketch.
- If you drag a file in creator it will get added to the library but will not get instantiated into the stage. This is completely doable but requires extra time.


packages/haiku-serialization/src/utils/dndHelpers.js
```diff
@@ -0,0 +1,62 @@
+/*
```
roperzh, line 1: @matthewtoast heads up to this file, I needed to put a bunch of methods shared by Creator/Glass/Timeline and this is the best place I found. If you think we should place them somewhere else please let me know

tsortwehttam: This is a-ok for now. At a later date we can revisit what really belongs in haiku-serialization and what should be moved elsewhere (it has become quite a hodgepodge over time). But, for now, nbd.

PR #38, opened 2017-10-12
Sketch downloader

roperzh:
**What**

[Trello card](https://trello.com/c/UIBE4ULD/773-offer-to-download-and-install-sketch-for-user-if-they-dont-have-it)

**Notes**

We are a bit skeptical as for if this will work in production due to permissions, needs testing on staging and before rolling prod to our users.


packages/haiku-creator/src/utils/sketchUtils.js
```diff
@@ -0,0 +1,31 @@
...
+          return unzip(zipPath, installationPath, 'Sketch')
+        })
+        .then(resolve)
+        .catch(reject)
+    })
+  },
+
+  checkIfInstalled () {
+    return new Promise((resolve, reject) => {
+      fs.access('/Applications/Sketch.app', fs.constants.F_OK, (err) => {
```
roperzh, line 25: @matthewtoast we are detecting if Sketch if installed by checking if `/Applications/Sketch.app` is present. TBH I feel like this very naive, but I'm not sure if there's a better way. Let me know if you have any suggestions/comments.

tsortwehttam: LGTM.

PR #39, opened 2017-10-13
Delete items from Library

roperzh:
**What**

[Trello Card](https://trello.com/c/GWjYUYX9/249-allow-assets-to-be-removed-from-the-library)

**Comments**

Here's how it looks like (cc @taylorpoe)

delete-menu


packages/haiku-plumbing/src/Master.js
```diff
@@ -341,7 +341,6 @@ export default class Master extends EventEmitter {
 
     if (extname === '.sketch' || extname === '.svg') {
       delete this._knownDesigns[relpath]
-      this.emitDesignChange(relpath)
```
roperzh, line 344: Heads up for this one, I removed it because it was trying to "merge" a filed that was already removed, but wanted to double check if makes sense

packages/haiku-plumbing/src/Master.js
```diff
@@ -495,14 +494,35 @@ export default class Master extends EventEmitter {
...
-      if (removeErr) return done(removeErr)
-      delete this._knownDesigns[relpath]
-      return done(null, this.getAssetDirectoryInfo())
-    })
+
+    /* Remove the file and all associated assets from the in-memory registry */
+    Object.keys(this._knownDesigns)
+      .filter((path) => path.indexOf(relpath) !== -1)
+      .forEach((path) => delete this._knownDesigns[path])
+
```
roperzh, line 508: Heads up to this.

PR #42, opened 2017-10-16
Pressing "space" or "play" from final frame triggers play from beginning

roperzh:
**What**

[Trello card](https://trello.com/c/nm3PodEs/813-in-haiku-timeline-pressing-space-or-play-from-final-frame-should-trigger-play-from-beginning)

**Preview**

playback


packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -1449,17 +1466,10 @@ class Timeline extends React.Component {
...
-            let frameInfo = this.getFrameInfo()
-            this._component.getCurrentTimeline().seekAndPause(frameInfo.fri0)
-            this.setState({ isPlayerPlaying: false, currentFrame: frameInfo.fri0 })
+            this.playbackSkipBack()
           }}
           playbackSkipForward={() => {
-            let frameInfo = this.getFrameInfo()
-            this.setState({ isPlayerPlaying: false, currentFrame: frameInfo.friMax })
-            this._component.getCurrentTimeline().seekAndPause(frameInfo.friMax)
+            this.playbackSkipForward()
```
roperzh, line 1472: @matthewtoast I moved these methods from inline functions to class functions because I needed to reuse them in the class, let me know if makes sense.

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -939,7 +939,24 @@ class Timeline extends React.Component {
...
+  }
+
+  playbackSkipForward () {
+    let frameInfo = this.getFrameInfo()
+    this.setState({ isPlayerPlaying: false, currentFrame: frameInfo.friMax })
+    this._component.getCurrentTimeline().seekAndPause(frameInfo.friMax)
+  }
+
   togglePlayback () {
+    if (this.state.currentFrame >= this.state.maxFrame) this.playbackSkipBack()
```
roperzh, line 958: @matthewtoast this is the magic line that makes everything possible. At first I thought about moving this logic into a separate method like this:

```javascript
someMeagnifulName () {
  if (this.state.currentFrame >= this.state.maxFrame) this.playbackSkipBack()
  this.togglePlayback()
}
```

But I think that makes more sense to have only one function to control the playback globally. Let me know if you want to change the current implementation.

packages/haiku-timeline/src/components/Timeline.js
tsortwehttam: Makes sense - it's a lot cleaner this way.

PR #43, opened 2017-10-18
feat: Initial work to support Lottie export

stristr:
Note: no generated (`packages/haiku-*/lib/*.js`) code is included in this PR. There is some light refactoring of existing code (e.g. changing `process` to `global.process`, switching some files to use ES6 imports, ...) but nothing major.

This PR introduces three new packages: `haiku-common`, `haiku-formats`, and `haiku-testing`, all written in TypeScript linted against a lightly forked Airbnb style guide.

`haiku-common` is meant to be a reasonable dependency for any library that ships in Haiku.app except `haiku-player`. The main thing it's doing right now is providing an extremely light "experiments" layer for soft-launching features in specific environments. It's an easy upgrade to support gradual-launching features in specific environments, but for now features are either "on in environment X" or "off in environment X". The use case is in the `TopMenu` module in `haiku-creator`, and looks like this:

```
import { Experiment, experimentIsEnabled } from 'haiku-common/lib/experiments'
...
if (experimentIsEnabled(Experiment.LottieExport)) { ... }
```

This allows us to "safely" make production builds while merging the WIP Lottie export upstream.

`haiku-testing` is a test base with grand ambitions TBD. Right now, it exports a lightweight `mock` module for simplifying the tedious work of stubbing out dependencies in unit tests. Its dependency replacement is based on proxyquire and its stubs are based on sinon.

The main work of Lottie export is beginning to unfold in `haiku-formats`. The BodymovinExporter class does all the work, and its unit tests basically thrash the capabilities of destructuring assignment to demonstrate the current functionality.


packages/haiku-creator/src/TopMenu.js
```diff
@@ -68,23 +72,39 @@ class TopMenu extends EventEmitter {
...
+      })
+    }
+
     Menu.setApplicationMenu(Menu.buildFromTemplate([
       {
         label: app.getName(),
         submenu: mainMenuPieces
       },
       {
+        // TODO(xifm6diW): disable when there is no active project.
```
roperzh, line 105: I'm curious about the TODO id, does it serve a specific purpose? (probably going to copycat if so hehe)

packages/haiku-testing/src/mock.ts
```diff
@@ -0,0 +1,53 @@
...
+import * as proxyquire from 'proxyquire';
+import * as sinon from 'sinon';
+
+/**
+ * Retrieves a fresh stub.
+ */
+export const getStub = () => sinon.stub();
+
+/**
+ * Generic stub importer based on proxyquire and sinon.
```
roperzh, line 11: This was super-necessary (feel free to facepalm with [this](https://github.com/HaikuTeam/mono/blob/master/packages/haiku-creator/test/unit/autoUpdate.test.js#L74))

packages/haiku-plumbing/src/Master.js
```diff
@@ -99,6 +100,7 @@ export default class Master extends EventEmitter {
 
     this.proc.socket.on('error', (err) => {
       logger.info('[master] !!! socket error', err)
+      logger.info('[master] !!! socket error', err)
```
roperzh: Possibly duplicated?

packages/haiku-creator/src/TopMenu.js
stristr: A specific future purpose that I'm mostly playing around with right now: better task tracker/GitHub integration. Goal would be to add a linter that enforces that every TODO has a task ID, and introduce tools so that task IDs are clickable to their [canonical URLs](https://trello.com/c/xifm6diW).

packages/haiku-testing/src/mock.ts
stristr: Ah no, that's fine too :smile:

Nice thing about importStubs is it lets you mock out a list of dependencies that come in via ES6 imports. Also sinon stubs are the :bomb:. I'm going to try to see if I can do something similar to [unittest.mock.patch](https://docs.python.org/3/library/unittest.mock.html) in Python 3 using TypeScript decorators. It's harder because no JS testing framework is class-based, but TypeScript decorators are for classes only. Anyway…decided to leave this alone for now.

packages/haiku-plumbing/src/Master.js
stristr: Nice catch! (No idea how that got there.) Fixed in f433a00.

PR #44, opened 2017-10-18
Refactor autoupdates and fix tests

roperzh:
**What**

- Minor changes to properly manage error cases on Promises and make the code clearer.
- Fix tests broken since we moved from async/await to Promises.


packages/haiku-creator/src/utils/fileManipulation.js
```diff
@@ -36,16 +36,16 @@ module.exports = {
     })
   },
 
-  unzip (zipPath, destination, filename) {
+  unzipAndOpen (zipPath, destination, filename) {
+    const unzipCommand = `unzip -o -qq ${zipPath} -d ${destination}`
```
tsortwehttam: I missed this the first time around, but we should make sure to always `JSON.stringify()` any path string that appears in `.exec`, `.execSync` etc., in case that path `/Has Any/Spaces That/Could CreateProblems/`

roperzh:
Updated with comments and rebased against latest master.

PR #48, opened 2017-10-19
Disable project-specific menus if not in a project

roperzh:
**What**

[Trello card](https://trello.com/c/xifm6diW/799-disable-project-menu-options-when-there-is-no-active-project)

**Comments**
- While the Trello card mentions only the 'Project' menu, I also disabled zooming/editing features.

- Unfortunately seems like we can't disable a menu if it has a submenu, therefore we have to disable the submenu items.

**UPDATE**
Initially, I disabled copy/paste/cut, but later I realized that those are necessary for the project browser too, sorry about this. Here is the result:

result


packages/haiku-creator/src/electron.js
```diff
@@ -117,6 +117,7 @@ function createWindow () {
 
   ipcMain.on('renderer:projects-list-fetched', (ipcEvent, projectList) => {
     menuspec.projectList = projectList
+    menuspec.folder = null
```
roperzh, line 120: @matthewtoast raising attention on this.

roperzh:
@matthewtoast  I updated the main comment to reflect that we are not disabling cut/copy/paste in the project browser, sorry about that.

tsortwehttam:
👍

PR #51, opened 2017-10-19
Fix cut/copy from top menu

roperzh:
**What**

[Trello card](https://trello.com/c/D6JutqTw/844-cut-copy-from-the-top-menu-is-not-working)

**Comments**

Feels kinda bad to add that amount of code for such a simple thing, but on the bright side, the envoy boilerplate for glass gets set up.

If you think that this feature is not necessary, I'm ok deleting this and hiding the items from the menu


packages/haiku-glass/src/react/Glass.js
```diff
@@ -107,6 +107,11 @@ export class Glass extends React.Component {
       timelineChannel.on('didSeek', this.handleTimelineDidSeek.bind(this))
     })
 
+    this._component.on('envoy:glassClientReady', (glassChannel) => {
+      glassChannel.on('cut', this.handleVirtualClipboard.bind(this, 'cut'))
+      glassChannel.on('copy', this.handleVirtualClipboard.bind(this, 'copy'))
```
stristr: For a future style guide update: _usually_ prefer `() => this.handleVirtualClipboard('cut')` over `this.handleVirtualClipboard.bind(this, 'cut')`. When we compile to ES5+, there's a subtle performance advantage because we don't create a new `this`-bound function.

roperzh: understood! thanks!

stristr: Also (sorry), `() => {this.handleVirtualClipboard('cut')}` since we don't use the value returned.

PR #60, opened 2017-10-26
feat: "80%" of Lottie export.

stristr:
Please expand `bodymovinExporter.ts` for the big picture; I think GitHub hides it by default.

 - Adds support for background color (but not opacity, yet) on the wrapper div. Lottie doesn't speak "wrappers", so requires us to transclude a special rectangle during exporter bootstrapping.
 - Adds support for z-index, which works an odd way in Bodymovin ("lighter" elements need to come earlier, not later).
 - Adds support for path decomposition for paths (e.g. `M0,0 L1,1 L0,0 Z M2,2 L3,3 L2,2 Z`) with multiple closures. This is required because Bodymovin only draws singly-closed paths, so we need shim in the implicit "multiple shapes" in the midst of our recurve parse of the template.
 - Adds most of the necessary support for non-closed paths, including support for `stroke-linecap`.
 - Adds support for CSS non-colors that show up in color-able places inside SVGs (e.g. `stroke="none"`).
 - Adds support for transform-origin (which is required for scale and rotation transformations to be WYSIWYG through export; Lottie uses the After Effects default, which is `(0, 0)` for some strange reason).
 - Adds support for 2.5D rotation transformations and scale transformations. Note: there are some major limitiations to the scale transformation because although Bodymovin nominally supports separate tweening between `scale.x` and `scale.y`, the Lottie player actually ignores the transitions on `scale.y`.
 - Adds support for compound animations, mainly for the scale use case. There are lots of comments in this commit explaining the what and the why of compound animations, but the gist is Bodymovin requires some compound properties like scale to be expressed as a single tuple `[scaleX, scaleY]`. The consequences of this are…complex.
 - Adds support for layer out-points and in-points, without which nothing seems to actually show up in the Bodymovin web player.
 - Adds support for tween-less transitions, which are the de facto hack for creating animations where objects spring into existence on stage at some time `t > 0` but are not supported in After Effects at all. This is achieved by preprocessing the copy of the bytecode our class munges on and injecting 0-frame linear tweens whenever necessary. The unit test for `normalizes curves for transitions lacking tweens` should be readable and explain the expected behavior nicely.

### DIFF
--- a/packages/haiku-common/src/types/enums.ts
+++ b/packages/haiku-common/src/types/enums.ts
@@ -29,4 +29,5 @@ export enum Curve {
   EaseInOutElastic = 'easeInOutElastic',
   EaseInOutQuint = 'easeInOutQuint',
   EaseInOutSine = 'easeInOutSine',
+  Linear = 'linear',
 }
--- a/packages/haiku-formats/src/exporters/bodymovin/bodymovinEnums.ts
+++ b/packages/haiku-formats/src/exporters/bodymovin/bodymovinEnums.ts
@@ -11,6 +11,9 @@ export enum AnimationKey {
 }
 
 export enum LayerKey {
+  StartTime = 'st',
+  InPoint = 'ip',
+  OutPoint = 'op',
   Shapes = 'shapes',
   Type = 'ty',
   Transform = 'ks',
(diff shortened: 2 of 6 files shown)


tsortwehttam:
👍 👍

PR #61, opened 2017-10-27
Show the login screen on empty/invalid token

roperzh:
Before this change, if the token stored in `~/.haiku/auth` was
empty/invalid, the application was displaying an error message
prompting the user to restart the app. This was bad for two
reasons:

1. Restarting the app doesn't solve the problem.
2. This error message adds uncertainty about the stability of
Haiku.

With this change, if the login fails for a reason related to
the token the app will simply display the login screen.

Please note that this logic can be improved once we implement
token invalidation logic, but this needs efforts both in Inkstone
and the client.


zackbrown:
thanks for fixing this!

PR #63, opened 2017-11-01
Carbonite

roperzh:
**What**

Sentry provides a series of callbacks to modify the data before sending to the server. We are taking advantage of this and using one of these callbacks to trigger our carbonite action which:

- Creates a zip file containing:
  - The project folder, including designs and git history, excluding `node_modules`
  - All the logs
- Uploads the zip file into S3

Besides that, the reports now also include extra data which can be useful:
  - Experiments state
  - Project and User names
  - Link to the zip file

_Note: the zip file is only uploaded if the crash occurs when a project is open_


packages/haiku-serialization/src/utils/carbonite.js
```diff
@@ -0,0 +1,127 @@
...
+}
+
+function _zipProjectFolders ({destination, sources}) {
+  const parsedSources = sources
+    .map(source => path.relative(HOMEDIR_PATH, source))
+    .join(' ')
+
+  return new Promise((resolve, reject) => {
+    exec(
+      `tar --exclude='node_modules' -rvf ${destination} -C ${HOMEDIR_PATH} ${parsedSources}`,
```
tsortwehttam: Do we need to sanitize these paths? In case they contain spaces etc. (e.g. `/Foo Bar/Baz`)

tsortwehttam:
General question about Sentry; I noticed that in some of your Sentry tests, the stack trace only seems to show entries from the Raven library. For example, this one: https://sentry.io/haiku-systems/javascript-timeline/issues/396613049/ Were you able to find a way to get better call stack info from these errors occur in the JavaScript client?

roperzh:
@matthewtoast wasn't able to find a solution to better stack traces yet. It's a bit frustrating because according to the Sentry documentation what we are doing should be enough.

tsortwehttam:
kk, hopefully we can find a way to capture this stuff soon.

PR #64, opened 2017-11-01
feat: Support all ...Bounce and ...Elastic curves in Haiku.

stristr:
With [this neat programmable calculator](https://www.desmos.com/calculator/2hmyj2rbfl) as my lodestar, creates bezier decompositions of the easing functions we use for all ...Bounce and ...Elastic curves in Haiku.

I tested all curves manually, so this should be in pretty good shape.

### DIFF
--- a/packages/haiku-formats/src/exporters/curves.ts
+++ b/packages/haiku-formats/src/exporters/curves.ts
@@ -7,7 +7,6 @@ export type InterpolationPoints = [number, number, number, number];
 
 /**
  * Provide the interpolation points (the four arguments passed to `cubic-bezier(...)`) for a named curve.
- * TODO: Support non-bezier curves (*Bounce and *Elastic for now) by partitioning into a continuous set of beziers.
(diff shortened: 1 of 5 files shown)


packages/haiku-formats/src/exporters/curves.ts
```diff
@@ -30,15 +29,13 @@ export const getCurveInterpolationPoints = (curve: Curve|InterpolationPoints) =>
...
-    case Curve.EaseInBounce:
-    case Curve.EaseInElastic:
     case Curve.EaseInQuint:
       return [.755, .05, .855, .06];
     case Curve.EaseInSine:
       return [.47, 0, .745, .715];
 
     case Curve.EaseOutBack:
-      return [.175, .885, .32, 1.75];
+      return [.175, .885, .32, 1.275];
```
stristr, line 38: Note to reviewer: this change was intentional (see [easings.net](http://easings.net/#easeOutBack)).

PR #69, opened 2017-11-02
feat: Final full support for all features of our demo (built-in) projects for Lottie.

stristr:
A plethora of support for tricky things that show up in CheckTutorial, of all places!

In this PR:
 - Rewrite all `enum`s as `const enum`s, and fix TS compilation to transpile these enumeration values as their strings, instead of making a bunch of dictionaries we need to look up.
 - More tests for stroke and fill interpretation.
 - Full support for `stroke-dasharray`, which thankfully had an analogous equivalent in AE/Bodymovin.
 - Full support for `fill-rule` shading, which fixes the one remaining issue (O missing its donut hole) from Move.
 - Full support for `<linearGradient>` fills (not strokes, although a TODO is added and this *probably* isn't much extra work).
 - Partial, hacky support for `<radialGradient>` fills, with some known rendering deviations from what Haiku is likely to produce from a vanilla Sketch radial gradient. The main work remaining here involves unwinding [gradientTransform](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/gradientTransform). I figured we might already have a parser for SVG transforms floating around somewhere in `haiku-player`, so wanted to leave this alone until I've done some more discovery.

### DIFF
--- a/packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
+++ b/packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
@@ -1,4 +1,5 @@
 import * as difference from 'lodash/difference';
+import * as flatten from 'lodash/flatten';
 import * as mapKeys from 'lodash/mapKeys';
 
 import {Maybe} from 'haiku-common/lib/types';
@@ -15,6 +16,9 @@ import {
 } from '../curves';
 import {
   AnimationKey,
+  FillRule,
+  GradientKey,
+  GradientType,
   LayerKey,
   LayerType,
   PathKey,
@@ -32,19 +36,26 @@ import {
   linecapTransformer,
   scaleTransformer,
   linejoinTransformer,
+  fillruleTransformer,
+  dasharrayTransformer,
+  getValueReferenceMatchArray,
 } from './bodymovinTransformers';
 import {SvgInheritable} from './bodymovinTypes';
 import {
+  alwaysAbsolute,
   alwaysArray,
   compoundTimelineReducer,
   decomposeMaybeCompoundPath,
   getBodymovinVersion,
   getFixedPropertyValue,
+  getShapeDimensions,
+  initialValue,
+  initialValueOrNull,
   keyframesFromTimelineProperty,
   maybeApplyMutatorToProperty,
   pathToInterpolationTrace,
   pointsToInterpolationTrace,
-  timelineValuesAreEquivalent,
+  timelineValuesAreEquivalent
 } from './bodymovinUtils';
 
 let bodymovinVersion: Maybe<string>;
@@ -145,6 +156,16 @@ export class BodymovinExporter implements Exporter {
     }
 
     return timeline;
+  }
+
+  /**
+   * Retrieves the timeline for a specific node.
+   * @param node
+   * @param parentNode
+   * @returns {{}}
+   */
+  private timelineForNode(node: any, parentNode?: any) {
+    return this.timelineForId(node.attributes['haiku-id'], parentNode ? parentNode.attributes['haiku-id'] : undefined);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -1,4 +1,5 @@
 import * as difference from 'lodash/difference';
```
stristr, line 1: Don't ignore this file even though it's hidden by default!

stristr:
Now also featuring:
 - Limited and heavily-asterisked support for `Haiku.inject`'ed keyframe values, which are evaluated in the context of the state tree defaults during export, stubbing out all summonables to a magic class that becomes a copy of itself through deep property access, and becomes 0 when it is used in a scalar/expressiony context. (Unit tests should clarify what this means if the explanation is confusing).

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -270,8 +291,8 @@ export class BodymovinExporter implements Exporter {
 
     // Calculate the 2D center for 3D transformations.
     if (timeline.hasOwnProperty('sizeAbsolute.x') && timeline.hasOwnProperty('sizeAbsolute.y')) {
-      centerX = timeline['sizeAbsolute.x'][0].value / 2;
-      centerY = timeline['sizeAbsolute.y'][0].value / 2;
+      centerX = initialValue(timeline, 'sizeAbsolute.x') / 2;
+      centerY = initialValue(timeline, 'sizeAbsolute.y') / 2;
     }
```
tsortwehttam, line 297: Here we might want to multiply by origin.x and origin.y.

packages/haiku-formats/src/exporters/bodymovin/bodymovinTransformers.ts
```diff
@@ -18,6 +22,16 @@ export const opacityTransformer = (opacity) => 100 * opacity;
 export const scaleTransformer = (scale) => 100 * scale;
 
 /**
+ * Gets a match array for a value reference, e.g. for a property value like fill="url('#foobar')".
+ *
+ * This method will either return an array like [_, "foobar"] or null.
+ * @param {string} value
+ * @returns {RegExpMatchArray}
+ */
+export const getValueReferenceMatchArray = (value: string): RegExpMatchArray =>
```
tsortwehttam, line 31: Might be worth moving this into haiku-common since there are a couple of places in the player/serialization that use something similar to this.

packages/haiku-formats/src/exporters/bodymovin/bodymovinUtils.ts
```diff
@@ -127,6 +134,73 @@ export const alwaysArray = (maybeArray: any): any[] => {
...
+ * Gets the initial value of a timeline property.
+ *
+ * Warning: this method uses unchecked property access, assuming that the caller has already checked the timeline
+ * property exists. In cases where there's no need to check outside the context of this property, prefer
+ * `initialValueOrNull` below.
+ * @param timeline
+ * @param {string} property
+ * @returns {any}
+ */
+export const initialValue = (timeline: any, property: string): any => timeline[property][0].value;
```
tsortwehttam, line 163: In case useful, we do already have some utility functions (messy, and long in need of refactor, etc etc) that might be helpful here. And/or they may provide a somewhat more robust way of doing what you're doing here.

https://github.com/HaikuTeam/mono/blob/master/packages/haiku-bytecode/src/TimelineProperty.js#L95

For example, the method I linked ^^ lets you obtain the value for a property in a timeline from an arbitrary time, and it also lets you pass in a "host instance" (a `player/src/HaikuComponent` instance) which makes it possible for you to retrieve computed values (e.g. from expressions). Also, if no defined value is present for the timeline+time, it also handles retrieving the official fallback value from our hardcoded enumeration of fallbacks:

https://github.com/HaikuTeam/mono/blob/master/packages/haiku-player/src/properties/dom/fallbacks.ts

That fallback enumeration ^^ is also worth mentioning in the context of this PR, since there are a few places in here where it looks like you're falling back to `0`. That's probably sufficient for now, but in the future we might want to use the "official" fallbacks.

tsortwehttam: Also, I meant to add just in case: Even though a property may not have a value explicitly defined for time `0`, it still has an implicit initial value, which is the fallback value, which is what the player chooses in cases where it really needs some kind of non-null value, such as calc'ing layout.

tsortwehttam:
Since none of my comments are blockers on this, I'm going to go ahead and merge this for sake of rolling it into a staging build for QA.

PR #70, opened 2017-11-03
clean-up (haiku-cli)

stristr:
A small deembarrassification push focused on `haiku-cli`.

 - Packages TSLint rules as a module (doesn't need to be compiled, so fine to not update scripts for now).
 - Standardizes `package.json` scripts.
 - Standardizes style to match new rules from `tslint-config-haiku`.
 - Removes test cruft and replaces with a pointer to the commit where test cruft was removed.


zackbrown (approved):
lgtm!

PR #73, opened 2017-11-07
productivity: create `yarn sync`.

stristr:
A much-needed upgrade for `yarn yarn-install && yarn relink`.

 - Extends the export from `scripts/helpers/allPackages.js` so it can be passed an array of package names to resolve.
 - Moves the "guts" of `yarn-install.js`, `yarn-link.js`, and `yarn-unlink` into helpers, and makes them async.
   - Anecdotally: making the work of `yarn relink` async led to a 7x performance improvement for the standalone command in a `yarn-install`'ed project.
   - Side note: I removed the cleanup work done by `yarn yarn-unlink`, which calls `yarn unlink` on each linked module. I think unlinking the dependencies from the packages is sufficient here. `yarn unlink` from the module root seems to just unregister the module from being linkable at all, which isn't necessary because `yarn link` is idempotent.
 - Creates a new yarn command, `yarn sync`, which does the following:
   - Reads a local lockfile (`.last-sync`) if it exists.
   - Identifies **only** the packages that have a changed `yarn.lock` since the last sync.
   - Does the equivalent work of `yarn install` (but with 10 retries per dep, :smile:) and `yarn relink` for that set of packages only.

BREAKING CHANGE: We are now using `--frozen-lockfile` in `yarn sync`/`yarn yarn-install`. This means that if `yarn.lock` is out of date, both of these commands will fail. I noticed that `yarn.lock` was out of date specifically with the `yarn.lock` not being bumped up to `"@haiku/player@2.3.19"` in plumbing, so we would want to identify what process elided that detail and fix it. (Whenever `yarn.lock` is out of date for a given package, `yarn install` is the answer.)

From some reductive benchmarking, if all your deps are up to date, this should take < 30s to run the first time and < 1s to run subsequent times.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/*
 config.js
 source/*
 ~tmp
+.last-sync
--- a/README.md
+++ b/README.md
@@ -28,8 +28,7 @@ And add to `~/.ssh/config`:
 **git/yarn/packages setup**
 
     $ yarn install
-    $ yarn run yarn-install
-    $ yarn relink
+    $ yarn sync
 
 ## Development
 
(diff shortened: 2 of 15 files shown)


tsortwehttam:
> so we would want to identify what process elided that detail and fix it

Must be this - https://github.com/HaikuTeam/mono/blob/master/scripts/semver.js - which I run when cutting a new release. 

I'll add a Trello card for that.

PR #77, opened 2017-11-08
fix: `yarn sync` should link/unlink pkgname, not name.

stristr:
Not doing this causes a really "fun" bug due to not linking `@haiku/player`, leaving various copies of version 2.3.7 lying around (because `yarn.lock` is out of date ever since we stopped cutting packages to standalone repos, but there's a Trello ticket for this). The side effects of having an old version of Player in the app are…surprising. But we'll fix that with the Trello ticket, and the real fix is probably using yarn workspaces.

Fun new benchmark: sync from fresh clone and empty yarn cache = 430s.

In the unlikely event that you bricked your repo due to this bug, `yarn relink` should fix it after merging this commit.

### DIFF
--- a/scripts/helpers/yarnLink.js
+++ b/scripts/helpers/yarnLink.js
@@ -5,7 +5,7 @@ const lodash = require('lodash')
 const allPackages = require('./allPackages')()
 const log = require('./log')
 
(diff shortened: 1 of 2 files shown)


stristr:
@matthewtoast heads up, merging this now.

PR #79, opened 2017-11-10
refactor: Reboots haiku-player's `yarn develop` behavior to use webpack.

stristr:
Some foundational work for more effective Player profiling, benchmarking, etc.

 - Moves demos out of `test/` and into `demo/projects`.
 - Sets up a mostly-new approach to running the demos in `demo/server.js`. All browserify action is replaced with webpack, which allows us to write real, source-mapped bundles transpiled from our actual TypeScript (and debuggable/profileable back to our actual TypeScript) in a new "debug" mode which basically injects the bytecode and calls `HaikuDOMAdapter(bytecode)(mount)`. In the existing demos, I basically left what was there, except we now hot-compile separate bundles for `dom.js` and `react-dom.js` as source-mapped files serving on disk. **I definitely recommend you checking out the original demos to confirm nothing was inadvertently broken.**

Notes/gotchas:

 - I did *not* preserve the query-parameter ?react=false/?vanilla=false behavior. It's easy enough to add back if there's a reason we still want it.
 - I preserved `yarn develop` for launching the demo app, but rewrote `yarn watch` to `tsc --watch`. I think this is use-case-preserving, but may have missed an edge case. In particular, I'm assuming that `yarn link && yarn link @haiku/player` is only actually needed in the context of the demo app, and not needed for any part of developing/building/using the Player in the Electron app.


packages/haiku-player/develop.js
```diff
@@ -1,37 +0,0 @@
-var cp = require('child_process')
-var path = require('path')
-var opn = require('opn')
-var chokidar = require('chokidar')
-
-console.log('linking haiku player')
-cp.execSync('yarn link && yarn link @haiku/player', { stdio: 'inherit' })
```
stristr, line 7: Explicit assumption 1: ^ this step is only necessary for running the demos.

packages/haiku-player/develop.js
```diff
@@ -1,37 +0,0 @@
...
-
-console.log('compiling typescript')
-cp.execSync('yarn run compile', { stdio: 'inherit' })
-
-console.log('watching typescript')
-cp.spawn(`tsc`, [`--watch`], { stdio: 'inherit' })
-
-if (process.env.HAIKU_PLAYER_WATCH_ONLY !== '1') {
-  console.log('starting server')
```
stristr, line 16: Explicit assumption 2: HAIKU_PLAYER_WATCH_ONLY is only used in the context of running `yarn start` in mono.

packages/haiku-player/develop.js
```diff
@@ -1,37 +0,0 @@
...
-  chokidar.watch('src', {
-    ignored: /(^|[\/\\])\../
-  }).on('change', (event, path) => {
-    console.log('file changed', event)
-    try {
-      cp.execSync(`tslint -c tslint.json 'src/**/*.ts' --exclude 'src/vendor/**' --format stylish`, { stdio: 'inherit' })
-    } catch (exception) {
-      // empty
-    }
-  })
```
stristr, line 36: Explicit assumption 3: all of the above is replaced with more standardized flows—`yarn lint`, `yarn fix`, etc.

tsortwehttam:
Reviewed and this looks fine locally. All your assumptions are correct, so merging.

PR #81, opened 2017-11-10
tweak: Use last sync commit instead of last sync date in `yarn sync`.

stristr:
This should obviate any future need for `rm .yarn-sync && yarn sync` by being merge-strategy agnostic. The old implementation (based on the local date of the last sync) was implicitly assuming that every commit is rebased in order before it's merged, and that the `--after` filter in `git whatchanged` respects rebased merge dates and not original commit dates (I'm not sure that it does). By using a "since commit" type of filter instead, short of force pushing hard resets that erase history on a shared branch (please don't do this :smile:), this should be a more stable approach.

### DIFF
--- a/scripts/sync.js
+++ b/scripts/sync.js
@@ -16,17 +16,20 @@ checkYarnVersion()
 checkNodeVersion()
 
 const lastSyncFilename = path.join(global.process.cwd(), '.last-sync')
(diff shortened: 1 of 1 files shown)


stristr:
I'm gonna merge this now to open up new PR on same branch.

PR #89, opened 2017-11-16
Events ui

roperzh:
This is ready for review.

There will be probably a couple of bugs, but the most part is done and the functionality is finally here.

**Interesting bits**

- The events editor used to be a monolith file, this PR splits it into multiple components, which you can see in [components/EventHandlerEditor](https://github.com/HaikuTeam/mono/tree/events-ui/packages/haiku-glass/src/react/components/EventHandlerEditor)
- You may want to take a quick look at [HandlerManager](https://github.com/HaikuTeam/mono/pull/89/files#diff-9d3ac96cd39313832949398de3b47bf5)
- For the formatting issues: I ended up adding [yet another patch](https://github.com/HaikuTeam/mono/blob/events-ui/packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js#L198), we should fix this, but I think it's going to take a significant amount of work/discussion and it belongs to another PR.
- There are two naive checks for the root element that will need special attention when implementing multicomponents, check https://github.com/HaikuTeam/mono/pull/89/commits/ffe298d5ed1c23a70f063d108d9163684a7d243a.


packages/haiku-glass/src/react/Glass.js
```diff
@@ -1154,7 +1154,11 @@ export class Glass extends React.Component {
     }
   }
 
-  buildBoltInstance (x, y) {
+  buildBoltInstance (x, y, rotationZ) {
```
tsortwehttam: Mentioning in case it hasn't come up yet: You may also need to account for negative scaling here. I.e. scale the element so that it is flipped over. For X, Y, and X+Y together.

roperzh: Thank you! I'm looking into it now :)

roperzh:
Moving from wip -> ready for review 🎉

roperzh:
Updated main comment with some pointers/comments

packages/haiku-glass/src/react/Glass.js
```diff
@@ -242,7 +247,10 @@ export class Glass extends React.Component {
         // If we've toggled into preview mode, we have to force react to update the on-stage styles
         this.forceUpdate()
         this._component.getCurrentTimeline().togglePreviewPlayback(this.isPreviewMode())
-
+      } else if (what === 'showEventHandlersEditor') {
+        const [{elementUID, options}] = args
+        this.setLastSelectedElement(this._component.findElementByComponentId(elementUID))
```
tsortwehttam: Note to self: Method call to change in multi-component.

packages/haiku-glass/src/react/Glass.js
```diff
@@ -470,10 +479,14 @@ export class Glass extends React.Component {
...
+  saveEventHandlers (targetElement, serializedEvents) {
     let selectorName = 'haiku:' + targetElement.uid
-    this._component.upsertEventHandler(selectorName, eventName, handlerDescriptorSerialized, { from: 'glass' }, () => {
-
+    this._component.batchUpsertEventHandlers(selectorName, serializedEvents, { from: 'glass' }, () => {
+      this.props.websocket.action(
+        'eventHandlersUpdated',
+        [this.props.folder],
+        () => {}
+      )
```
tsortwehttam, line 489: Note to self: Method call to change in multi-component. (Move into modified ActiveComponent method which abstracts away the websocket hook.)

packages/haiku-glass/src/react/Glass.js
```diff
@@ -1509,14 +1561,15 @@ export class Glass extends React.Component {
...
             : ''}
 
-          {(!this.isPreviewMode() && this.state.isEventHandlerEditorOpen)
-            ? <EventHandlerEditor
-              ref='eventHandlerEditor'
+          {!this.isPreviewMode() &&
+            <EventHandlerEditor
               element={this.state.targetElement}
-              save={this.saveEventHandler.bind(this)}
+              save={this.saveEventHandlers.bind(this)}
```
tsortwehttam: Nit: We should move method bindings to the constructor.

packages/haiku-glass/src/react/Icons.js
```diff
@@ -470,3 +466,34 @@ export const PasswordIconSVG = () => (
...
+              <path d='M8.5,6 C8.224,6 8,6.18327273 8,6.40909091 L8,14.5909091 C8,14.8167273 8.224,15 8.5,15 C8.776,15 9,14.8167273 9,14.5909091 L9,6.40909091 C9,6.18327273 8.776,6 8.5,6 Z' />
+              <path d='M6.5,6 C6.224,6 6,6.18327273 6,6.40909091 L6,14.5909091 C6,14.8167273 6.224,15 6.5,15 C6.776,15 7,14.8167273 7,14.5909091 L7,6.40909091 C7,6.18327273 6.776,6 6.5,6 Z' />
+              <path d='M3.5,6 C3.224,6 3,6.18327273 3,6.40909091 L3,14.5909091 C3,14.8167273 3.224,15 3.5,15 C3.776,15 4,14.8167273 4,14.5909091 L4,6.40909091 C4,6.18327273 3.776,6 3.5,6 Z' />
+            </g>
+          </g>
+        </g>
+      </g>
+    </svg>
+  )
+}
```
tsortwehttam, line 499: Nit: We should eventually move this whole file into `haiku-ui` so we can easily share common icons.

packages/haiku-glass/src/react/components/EventHandlerEditor/Editor.js
```diff
@@ -0,0 +1,247 @@
...
+      id: this.props.id,
+      event: eventName,
+      handler: {
+        params: this.props.params,
+        body: this.state.contents,
+        type: 'FunctionExpression',
+        name: null
+      }
+    }
+  }
```
tsortwehttam, line 156: Nit/future: After I merge multi-component, I'd like to move all logic like this into the BLL.

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -0,0 +1,249 @@
...
+  _generateID (len = 3) {
+    let str = ''
+    while (str.length < len) {
+      str += ALPHABET[Math.floor(Math.random() * ALPHABET.length)]
+    }
+    return str
+  }
+}
+
+export default HandlerManager
```
tsortwehttam, line 249: Nit/future: Once I merge multi-component I'd like to move this entire file into the BLL.

packages/haiku-glass/src/react/Icons.js
roperzh: +100, I know that we are probably going to do that post-launch, but I feel we are reaching to the point where is being painful to keep separate files for the Palette and the Icons.

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Since the Glass renders on every animation frame, we implemented a custom
+   * shouldComponentUpdate function for performance reasons.
+   *
+   * We should avoid at all costs:
+   *
+   * 1- Triggering a re-render
+   * 2- Instantiating a HandlerManager
+   */
+  shouldComponentUpdate ({element, visible, options}, nextState) {
```
tsortwehttam, line 54: Note to self: When merging multi-component I will need to check whether 'element' needs to be modified here.

packages/haiku-plumbing/src/Master.js
```diff
@@ -587,6 +587,14 @@ export default class Master extends EventEmitter {
...
   }
 
+  showEventHandlersEditor ({ params }, cb) {
+    return this._component.showEventHandlersEditor.apply(this._component, params.concat(cb))
+  }
+
+  eventHandlersUpdated ({ params }, cb) {
+    return this._component.eventHandlersUpdated.apply(this._component, params.concat(cb))
+  }
+
```
tsortwehttam, line 597: Note to self: Multi-component API change.

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -1229,6 +1258,32 @@ class ActiveComponent extends BaseModel {
...
+        if (metadata.from === this.alias) {
+          this.batchedWebsocketAction('batchUpsertEventHandlers', [this.folder, selectorName, serializedEvents])
+        }
+
+        return cb()
+      })
+    })
+  }
+
   /**
```
tsortwehttam, line 1286: Note to self: All lines above need to be modified for multi-component.

packages/haiku-serialization/src/bll/Element.js
```diff
@@ -191,6 +209,15 @@ class Element extends BaseModel {
...
 
+  batchUpsertEventHandlers (serializedEvents) {
+    /* eslint-disable no-unused-vars */
+    let eventHandlers = this.getReifiedEventHandlers() // pointer to substructs[0].bytecode
+    eventHandlers = serializedEvents // eslint-disable-line
+    this.emit('update', 'element-event-handler-update')
+    return this
+    /* eslint-enable no-unused-vars */
+  }
+
```
tsortwehttam: QUESTION: What is the purpose of those variables being assigned? Can this be removed?

packages/haiku-timeline/src/components/FrameGrid.js
```diff
@@ -18,41 +22,77 @@ export default class FrameGrid extends React.Component {
...
+
+    if (what === 'timeline-frame-range' || what === 'timeline-frame-hovered') {
       this.forceUpdate()
     }
+
+    if (what === 'timeline-frame-action') {
+      setTimeout(() => {
+        this.upsertTimelineEvents()
+        this.forceUpdate()
+      }, 500)
```
tsortwehttam, line 42: QUESTION: What is the purpose of the timeout here?

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -483,6 +487,19 @@ class Timeline extends React.Component {
...
+  }
+
+  showEventHandlersEditor (elementUID, frame) {
+    this.props.websocket.action(
+      'showEventHandlersEditor',
+      [this.props.folder, elementUID, {isSimplified: Boolean(frame), frame}],
+      () => {}
+    )
+  }
+
```
tsortwehttam, line 502: Note to self: Fix for multi-component.

packages/haiku-serialization/src/bll/Element.js
roperzh: IIRC, the idea is to completely replace the pointer to `substructs[0].bytecode` with the new batch of `serializedEvents`. But, right now I'm not sure if makes sense

packages/haiku-timeline/src/components/FrameGrid.js
roperzh: My bad, I should have left a comment at lest!

The thing is, when a component listeners are updated, we trigger the `timeline-frame-action` event, but the event arrives here _before_ the component is updated in disk.

Since we are doing: `this.rootElement.getCustomEvents()` to get the listeners and display the little bolts in the timeline, we need to wait a "safe" amount of time till we can safely read all events from the component.

This is one of the things that I'm not happy about, if you have any ideas I will go for them, otherwise I will paste this as a comment.

packages/haiku-serialization/src/bll/Element.js
tsortwehttam: In this case, the variable `eventHandler` is just getting assigned to one value, and then to another value. So L214-215 aren't actually modifying anything (unless I'm missing something obvious). Is the intent to update something here?

packages/haiku-timeline/src/components/FrameGrid.js
tsortwehttam: Gotcha. This is fine to merge for now. But in the future we should try to avoid this kind of thing, since usually the result is "I solved one race condition by introducing two new race conditions" ;) (To be clear, I am guilty of this on many counts!)

Side note: We unfortunately have a significant amount of logic like this through the system. I've addressed a bit of it during multi-component, but we should think in general about how we can make logically explicit race conditions in the code, especially when those conditions are conceptually or organizationally far apart. 🏓

tsortwehttam:
Since there is a lot of code here, I'd like to also nominate @stristr to take a second look at this before giving the green light to merge (if he has time).

packages/haiku-bytecode/src/actions/batchUpsertEventHandlers.js
```diff
@@ -0,0 +1,29 @@
...
+  serializedEvents
+) {
+  // To convert legacy event handlers array to object
+  upgradeBytecodeInPlace(bytecode)
+
+  if (!bytecode.eventHandlers) {
+    bytecode.eventHandlers = {}
+  }
+
+  bytecode.eventHandlers[selectorName] = {}
```
stristr, line 16: Nit/syntactic sugar: L12-L16 can be rewritten as:

```
bytecode.eventHandlers = {
  ...bytecode.eventHandlers,
  [selectorName]: {},
}
```

stristr: (b/c `{...undefined}` is `{}`)

packages/haiku-serialization/src/utils/react2haiku.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function react2haiku (reactInstance) {
+  const children = reactInstance.props.children
+
+  return {
+    elementName: reactInstance.type,
+    attributes: Object.assign({}, reactInstance.props, {children: null}),
+    children: children ? [react2haiku(reactInstance.props.children)] : []
```
stristr: qq: is `children` an array? I would expect this to be:

```
  children: reactInstance.props.children ? reactInstance.props.children.map(react2haiku) : []
```

packages/haiku-glass/src/react/Glass.js
stristr: Better: `save={() => { this.saveEventHandlers() }}`

packages/haiku-glass/src/react/Icons.js
stristr: In case it wasn't already known (I figure it is since you LGTMed it), `haiku-ui-common` is already in master if you felt like moving it inside this PR.

stristr:
Overall LGTM! I left a few comments, mostly nits.

You may want to look carefully at the several opportunities to replace `this.foo = this.foo.bind(this)` in constructors which I noted…in some places it might not be the best call, but I think it usually fits the bill.

packages/haiku-serialization/src/utils/react2haiku.js
tsortwehttam: FWIW: I might be remembering this wrong, but -- during a deep foray into React last year, I recall noting that `children` can be any of these: a single React element, a single string, or an array of those. I believe this is because the JSX compiler may resolve `<a><b/></a>` to `React.createElement('a',React.createElement('b'))` instead of `React.createElement('a',[React.createElement('b')])`.

stristr: Interesting, I buy that (especially considering that Roberto's code appears to be working). We might want to at least note in a comment, in case there's a hidden bug we trigger sometime later, that we're not insulated against the cases where it's not a single React element.

stristr:
Style nit: I actually have a really hard time detecting the bottom boundary of the event selector menu ([screenshot](https://cl.ly/1y2e451P1p1s))—it really looks like there's a misplaced `  }` menu item down at the bottom.

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
```diff
@@ -0,0 +1,222 @@
...
+}
+
+class EventHandlerEditor extends React.PureComponent {
+  constructor (props) {
+    super(props)
+
+    this.handlerManager = null
+    this.onEditorContentChange = this.onEditorContentChange.bind(this)
+    this.onEditorEventChange = this.onEditorEventChange.bind(this)
+    this.onEditorRemoved = this.onEditorRemoved.bind(this)
```
stristr, line 41: I'm actually getting an error on this binding when I try it locally. TypeError: Invalid attempt to destructure non-iterable instance.

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
```diff
@@ -0,0 +1,123 @@
...
+    backgroundColor: Palette.DARKEST_COAL,
+    borderRadius: '50%',
+    width: '18px',
+    height: '18px',
+    textAlign: 'center',
+    cursor: 'pointer',
+    fontSize: '20px',
+    lineHeight: '18px'
+  }
+}
```
stristr, line 27: Something's up with the usability of the snippet menu ([screenshot](https://cl.ly/0m2l1n2X2m3L)). On my laptop screen, it gets cut off and I can't scroll beyond one or two items. When I do scroll, it flickers and pushes down the menu.

packages/haiku-glass/src/react/components/EventHandlerEditor/ElementTitle.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import {get} from 'lodash'
+import Palette from '../../Palette'
+import {EventsBoltIcon} from '../../Icons'
+import truncate from '../../helpers/truncate'
+
+const STYLES = {
+  container: {
```
stristr, line 8: For some reason I get CSS cursor: grab on this container. Not sure where it's coming from, but it's super weird and makes me think I should be able to drag/resize the action editor. (It's okay that I can't, but would be much nicer if I can.)

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
stristr: Additionally, if I have N > 1 event listeners, I can only use the snippet menu at all on the first one. All subsequent ones are just a noop when I click the button.

packages/haiku-bytecode/src/actions/batchUpsertEventHandlers.js
roperzh: I love it, thanks!

packages/haiku-glass/src/react/Menu.js
```diff
@@ -0,0 +1,227 @@
...
+import Radium from 'radium'
+import Palette from './Palette'
+import {DownCarrotSVG} from './Icons.js'
+
+const STYLES = {
+  wrapper: {
+    position: 'relative'
+  },
+  menu: {
+    position: 'absolute',
```
stristr, line 11: Something about the absolute positioning is wrong for N > 1 events. Specifically, I see the menu here on the second of two events: https://cl.ly/193m2M1M292a.

If there are three or more, subsequent menus are absolutely positioned off-screen on Glass!

packages/haiku-glass/src/react/components/EventHandlerEditor/ElementTitle.js
```diff
@@ -0,0 +1,76 @@
...
+  render () {
+    return (
+      <div style={STYLES.container}>
+        <h3 style={STYLES.title}>{`${this.getElementTitle()} Actions`}</h3>
+        {!this.props.hideActions &&
+          <button style={STYLES.button} onClick={this.props.onNewAction}>
+            <span>
+              <EventsBoltIcon color={Palette.PALE_GRAY} />
+            </span>
+            <span style={STYLES.buttonText}>New Action</span>
```
stristr, line 63: Nit 1: It's kind of weird that this button doesn't change at all when I hover/click on it.

Nit 2: the "new action" behavior works great for cycling through the available listener types and making sure we don't make e.g. duplicate "click" listeners. But after I have added all available actions, the button remains but does nothing. It feels like it should ideally disappear or get visibly disabled.

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
```diff
@@ -0,0 +1,123 @@
...
+    if (typeof injectable === 'function') {
+      return injectable()
+    }
+
+    let range
+    const { lineNumber, column } = this.props.editor.getPosition()
+
+    if (this.hasCursorPosition()) {
+      range = new monaco.Range(lineNumber, column, lineNumber, column)
+    } else {
```
stristr, line 72: It feels like maybe logic is missing for cases where there is selected text. My expectation as a user is that if text is selected and I add a snippet, the snippet should replace the text.

packages/haiku-timeline/src/components/EventHandlerTriggerer.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  triggerEventHandlers () {
+    this.props.onEventHandlerTriggered(this.props.element.uid)
+  }
+
+  render () {
+    return (
+      <span onClick={this.triggerEventHandlers} style={STYLES.wrapper}>
+        <Bolt color={this.getBoltColor()} />
```
stristr, line 30: Inexplicably, if I resize an element while the Bolt is present, the bolt disappears. If I deselect and then re-select, it appears. No issue with translation.

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const SNIPPET_OPTIONS = {
+  'Change State': 'this.setState({stateName: value})',
+  'Go To And Play': 'this.getDefaultTimeline.goToAndPlay(ms)',
+  'Go To And Stop': 'this.getDefaultTimeline.goToAndStop(ms)',
+  'Pause': 'this.getDefaultTimeline.pause()',
+  'Stop': 'this.getDefaultTimeline.stop()',
+  'Docs': () => {
+    shell.openExternal('https://docs.haiku.ai/using-haiku/summonables.html')
```
stristr: Any way we can shim in a separator and somehow style the docs separately? It feels weird to have "docs" as the "last snippet"—I was confused why it took me to a browser and didn't just add a snippet.

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
stristr: Final note: the first issue reported (can't see full snippet menu) only occurs with N > 1 event listeners. Everything seems fine when there's exactly one event listener.

stristr (changes requested):

packages/haiku-timeline/src/components/icons/Bolt.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+export default ({color = '#93999A'}) => (
```
stristr: Nit: we can use svgo to remove the redundant nested `<g>` elements and simplify the `transform='translate(...)'` calls by applying them to the path.

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -559,6 +576,7 @@ class Timeline extends React.Component {
             position: 'absolute',
             top: 0,
             left: 0,
+            paddingTop: '5px',
```
stristr: Nit: probably should prefer `5` to `'5px'` for consistency with the rest of this file.

packages/haiku-glass/src/react/Menu.js
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+const closeIfSelectedClass = 'js-close-on-click'
+
+class BaseMenu extends React.Component {
+  constructor (props) {
+    super(props)
+
+    this.toggleOpen = this.toggleOpen.bind(this)
```
stristr: Here and throughout: except in rare cases where we are actually creating a perf bottleneck in a component that re-renders frequently, it should save local load to use fat arrows instead of instance-bound functions. e.g. instead of doing `this.a = this.a.bind(a); ... onClick={this.a}`, just do `onClick={() => this.a()}`.

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -454,6 +458,31 @@ class ActiveComponent extends BaseModel {
...
+      {elementUID, options}
+    )
+
+    // FIXME: for some reason sometimes the `metadata` argument is missing
+    if (typeof metadata === 'function') return metadata()
+    if (typeof cb === 'function') return cb()
+  }
+
+  eventHandlersUpdated (metadata, cb) {
+    console.log(metadata, cb)
```
stristr: Probably should remove this before submit.

packages/haiku-glass/src/react/components/EventHandlerEditor/EventSelector.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  render () {
+    return (
+      <div style={STYLES.selectWrapper}>
+        <Menu
+          trigger={this.renderMenuTrigger()}
+          fixed
+        >
+          {this.renderMenuItems()}
```
stristr: Nit: `renderMenuTrigger` and `renderMenuItems` are good candidates for a ES6 [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).

If you haven't used, you would basically change:

```
a() {
  return x
}

render() {
  return <Thing a={this.a()} />
}
```

with

```
get a() {
  return x
}

render() {
  return <Thing a={this.a} />
}
```

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -0,0 +1,249 @@
...
+      // wrote by the user causing two issues:
+      // 1. If the code only contains comments, the comments are deleted
+      // 2. The format is not respected.
+      handler.body =
+        prettier.format(handler.body) || this._buildEventHandler().handler.body
+
+      result.set(event, {
+        id,
+        handler
+      })
```
stristr, line 208: Fussy nit (please don't change, but note for future): there isn't a huge advantage to using Map over a plain Object when the keys are all known to be strings.

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    if (options && options.frame !== this.props.options.frame) {
+      return true
+    }
+
+    if (visible !== this.props.visible) {
+      return true
+    }
+
```
stristr: Nit: prefer

```
if ((a) || (b)) {
  return x
}
```

if `(a)` and `(b)` are conditions that both return `x`.

packages/haiku-serialization/src/bll/Element.js
stristr: 👍, likewise confused.

packages/haiku-timeline/src/components/EventHandlerTriggerer.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+class EventHandlerTriggerer extends React.PureComponent {
+  constructor (props) {
+    super(props)
+    this.triggerEventHandlers = this.triggerEventHandlers.bind(this)
+  }
+
+  getBoltColor () {
+    return this.props.element.getDOMEvents().length ? '#4DCEDA' : '#738C8F'
```
stristr: Nit: should put these in `Palette`. (Also see in `<FrameAction />`.)

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -483,6 +487,19 @@ class Timeline extends React.Component {
...
+  showFrameActionsEditor (frame) {
+    // #FIXME in multicomponents (matthew)
+    this.showEventHandlersEditor(this.component.findElementRoots()[0].uid, frame)
+  }
+
+  showEventHandlersEditor (elementUID, frame) {
+    this.props.websocket.action(
+      'showEventHandlersEditor',
+      [this.props.folder, elementUID, {isSimplified: Boolean(frame), frame}],
+      () => {}
```
stristr, line 499: Wish list, not for now: export `() => {}` as noop in `haiku-common`, and use it instead of `() => {}`.

tsortwehttam: FWIW, some of this will be tucked away with multi-component.

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -0,0 +1,249 @@
...
+   * containing the handler and a unique ID as the value.
+   *
+   * @param {Object} element
+   * @returns {Map}
+   */
+  _getParsedAppliedHandlers (element) {
+    const result = new Map()
+    const appliedEventHandlers = element.getReifiedEventHandlers()
+
+    for (let [event, rawHandler] of Object.entries(appliedEventHandlers)) {
```
stristr: Nit: for `(const ...)`.

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -0,0 +1,249 @@
...
+   */
+  _isDOMEvent (event) {
+    return this.applicableEventHandlersList.includes(event)
+  }
+
+  /**
+   * Finds an event that hasn't been applied to the element
+   */
+  _getNextAvailableDOMEvent () {
+    for (let event of this.applicableEventHandlersList) {
```
stristr: Nit: for `(const ...)`.

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -0,0 +1,249 @@
...
+      handler
+    })
+  }
+
+  _applicableEventHandlersToList () {
+    const result = []
+
+    for (let handlerGroup of this.applicableEventHandlers) {
+      for (let {value} of handlerGroup.options) {
+        result.push(value)
```
stristr: Nit: for `(const ...)` in both loops.

packages/haiku-glass/src/react/Icons.js
roperzh: I was in doubt about moving it inside this PR because there's already too much to take and we need to figure out a plan, we have a `Icons.js` file in every webview (glass, creator, timeline) and the icons have sightly different markup between files, so I believe that it will require some extra care.

If you feel like it will be a good idea to do it in this PR, I'm in! or maybe at we can move this file and figure out later what to do with the rest?

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
roperzh: cc @taylorpoe

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
roperzh: For this particular case, i choose a `Map` because it preserves the order in which the elements are `set`.

We need to preserve order because when the user presses the `New Action` button I would expect the item to be added at the top (or bottom) of the list.

I considered using an array, but it was too much of a hassle.

Just interested: do you think that any other data structure could have fill the purpose?

packages/haiku-glass/src/react/Menu.js
roperzh: Fixed! thanks for pointing that out.

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
roperzh: This should be fixed, thanks for the heads up!

packages/haiku-glass/src/react/components/EventHandlerEditor/ElementTitle.js
roperzh: Fixed!

packages/haiku-glass/src/react/components/EventHandlerEditor/ElementTitle.js
roperzh: Nit 1: agreed 100%, will add a comment for when we unify styles, otherwise I will have to copy and paste default button styles.

Nit 2: fixed!

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
roperzh: Fixed!

packages/haiku-timeline/src/components/icons/Bolt.js
roperzh: Done!

packages/haiku-bytecode/src/actions/batchUpsertEventHandlers.js
roperzh: note: the compiler is complaining about this one, but it's valid syntax. Probably a bug, maybe we should upgrade a package?

packages/haiku-glass/src/react/components/EventHandlerEditor/Snippets.js
taylorpoe: Yeah good thought. We should use an icon here to help the user understand they're about to blast over to the web. https://cl.ly/3t273z3a2i2a. @roperzh I have this icon in `haiku-creator` in the new Dashboard UI's `icons.js`. Feel free to add a task for me to circle back in add it here to the Actions UI after our PRs, or just copy the svg here: https://github.com/HaikuTeam/mono/pull/102/files#diff-8722beaa347e20a924afe7b4fd9e5b55R71 Whatever you think makes the most sense.

roperzh: @taylorpoe I will need that icon on the publish UI too (was using the one from the icon set which is a bit different). I will create a task in asana for either one of us to pick.

packages/haiku-glass/src/react/components/EventHandlerEditor/HandlerManager.js
stristr: This makes perfect sense. Thanks for explanation!

Only possibly "better" (read: maybe more efficient) data structure is an array of objects. But I would never choose it here due to tradeoff between ergonomics and efficiency.

packages/haiku-glass/src/react/Icons.js
stristr: Let's hold off!

packages/haiku-bytecode/src/actions/batchUpsertEventHandlers.js
stristr: Oh yeah, bytecode is one of the packages that doesn't compile. I'd just leave it, with a comment to fix when we standardize to TypeScript if you're feeling ambitious!

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
stristr: @roperzh, sorry, I don't think this is correct anymore.

I meant:

```
if ((options && options.frame !== this.props.options.frame) || visible !== this.props.visible) {
  return true
}
```

roperzh: Yes, there is only one explanation for this: I am dumb 😅 

Sorry for the mental hiccup.

stristr (approved):

stristr:
Hey @roperzh, before you merge and because I think the comment seems to be missing in certain views, wanted to make sure you were at least aware of the report [here](https://github.com/HaikuTeam/mono/pull/89#discussion_r154824965) and okay with it. Basically, when I resize an element below a certain width/height, the bolt icon disappears and I'm left with just a circle and no bolt. It remains gone even if I size back to the original, unless I deselect and then reselect the element.

roperzh:
Thanks! I'm aware of that comment (thanks for pointing that out!) but I haven't found a solution yet. It is very probably related to the player, as we had a similar issue with the bolt and that was the cause. I will keep you posted

PR #95, opened 2017-11-20
Fix preview mode + playback interactions

roperzh:
This adds a few tweaks to the preview mode + timeline playback
interactions:

- Disable the timeline and the library if preview mode is on
- Enable the component to take control of the playback during
preview mode
- Fix playback issues in the timeline when switching between
modes

**Notes**

- Would be great if we can improve the response time when switching modes, feels too slow now.

percy-2

### DIFF
--- a/packages/haiku-serialization/src/bll/Timeline.js
+++ b/packages/haiku-serialization/src/bll/Timeline.js
@@ -257,6 +257,16 @@ class Timeline extends BaseModel {
     return this
   }
 
+  toggleFreezeAndGotoStart () {
+    const timelineName = this.component.getCurrentTimelineName()
+    const timelineInstances = this.component.instance._timelineInstances
+    const timelineInstance = timelineInstances[timelineName]
+    timelineInstance.isFrozen()
+      ? timelineInstance.unfreeze()
+      : timelineInstance.freeze()
+    timelineInstance.seek(0)
+  }
+
   getDurationDragStart () {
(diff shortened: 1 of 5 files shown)


packages/haiku-serialization/src/bll/Timeline.js
```diff
@@ -257,6 +257,16 @@ class Timeline extends BaseModel {
...
+  toggleFreezeAndGotoStart () {
+    const timelineName = this.component.getCurrentTimelineName()
+    const timelineInstances = this.component.instance._timelineInstances
+    const timelineInstance = timelineInstances[timelineName]
+    timelineInstance.isFrozen()
+      ? timelineInstance.unfreeze()
+      : timelineInstance.freeze()
+    timelineInstance.seek(0)
+  }
+
```
roperzh: @matthewtoast I'm sure there's a better way to do this, but after trying and failing more times than I can count I settled on this solution. If you think this logic belongs somewhere else, please let me know.

packages/haiku-serialization/src/bll/Timeline.js
```diff
@@ -257,6 +257,22 @@ class Timeline extends BaseModel {
     return this
   }
 
+  togglePreviewPlayback (isPreviewMode) {
```
roperzh, line 260: Note that while this changed, I'm still concerned about if this is the right place/way to do this.

PR #99, opened 2017-11-22
tweak: slackShout to #releases instead of #engineering-feed.

stristr:
Simply changes the channel Travis announces release on from #engineering-feed to #releases. Low-friction quick change.


roperzh:
LGTM

PR #100, opened 2017-11-29
Fix caching of attributes in the player

roperzh:
**About**

This PR is not too time consuming to review, however it needs extra effort from the reviewer to check the player demos since this could potentially break something

Asana Task: https://app.asana.com/0/479779791675933/490282277397364

**What**

For certain properties, we keep a cache to compare from in order
to render elements because most of the times it's expensive to
read those attributes from the DOM.

The problem was, the cache was not being cleared/updated when:

- The element is removed from the DOM.
- An attribute is removed from the element.

This fix adds logic to clean the cache in those scenarios

### DIFF
--- a/packages/haiku-player/src/renderers/dom/assignAttributes.ts
+++ b/packages/haiku-player/src/renderers/dom/assignAttributes.ts
@@ -67,7 +67,7 @@ function setAttribute(el, key, val, cache) {
 }
 
 export default function assignAttributes(domElement, virtualElement, component, isPatchOperation) {
-  const cache = component.cache[getFlexId(virtualElement)]
+  const cache = component.cache[getFlexId(virtualElement)];
 
   if (!isPatchOperation) {
     // Remove any attributes from the previous run that aren't present this time around
@@ -80,8 +80,8 @@ export default function assignAttributes(domElement, virtualElement, component, 
           if (newValue === null || newValue === undefined || oldValue !== newValue) {
             domElement.removeAttribute(oldKey);
 
-            if(cache[oldKey]) {
-              cache[oldKey] = null
+            if (cache[oldKey]) {
+              cache[oldKey] = null;
             }
           }
         }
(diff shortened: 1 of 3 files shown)


packages/haiku-player/src/renderers/dom/assignAttributes.ts
```diff
@@ -77,6 +79,10 @@ export default function assignAttributes(domElement, virtualElement, component,
           // Removal of old styles is handled downstream; see assignStyle()
           if (newValue === null || newValue === undefined || oldValue !== newValue) {
             domElement.removeAttribute(oldKey);
+
+            if(cache[oldKey]) {
+              cache[oldKey] = null
+            }
```
stristr: Nit: this file needs `yarn fix`.

packages/haiku-player/src/renderers/dom/renderTree.ts
```diff
@@ -75,7 +75,7 @@ export default function renderTree(
     if (!virtualChild && !domChild) {
       // empty
     } else if (!virtualChild && domChild) {
-      removeElement(domChild);
+      removeElement(domChild, virtualElement, component);
```
stristr, line 78: Perf note: `getFlexId(virtualElement)` is called three times in a removal, including L34 above (a check that should return true). It's a light operation but `renderTree` does get called once per main element per tick, so we should probably make a single variable in this scope (`const flexId = getFlexId(virtualElement);`) and use it in L34/35 above as well as `removeElement`). Unrelated to your changes, but I'm also seeing this in `updateElement.ts` and `createTagNode.ts`, so might as well clean up there as well.

stristr (commented):
LGTM, although I have not tested Player demos. If you'd like, I can test those later today.

packages/haiku-player/src/renderers/dom/assignAttributes.ts
roperzh: Fixed, sorry about that! :)

packages/haiku-player/src/renderers/dom/renderTree.ts
roperzh: understood! Do you think it's okay to include those fixes in this PR? or should I make another once this gets merged? (including `updateElement.ts` and `createTagNode.ts`)

stristr: Yeah, of course! I think it's totally okay to include tangentially related code. Others may have different opinions :smile:

stristr:
Tested Player demos and all looks good. Merge away!

PR #101, opened 2017-11-29
Perf: cache the return value of `getFlexId`

roperzh:
**About**

This should be easy to review, there's no asana ticket associated with this.

**What**

Per @stristr [comment][comment-r153825825] it is a good idea to
cache the results of this instead of calling the function multiple times.

[comment-r153825825]: https://github.com/HaikuTeam/mono/pull/100#discussion_r153825825


tsortwehttam:
LGTM.

PR #102, opened 2017-11-30
Dashboard 2

taylorpoe:
PR Ready for review 👀 

Notes:
* Thumbnails won't size appropriately for users until they republish their projects.

Probably should fix before deploying (I'm going to work on these):

* share button was removed from project cards because we don't have the publish UI yet. But we should probably find out how to get sharelink for proj if has one and pass it in for this link as an interim substitute. **UPDATE**: Doesn't seem like it's available in plumbing. More notes here: https://app.asana.com/0/479779791675933/496079697953107

Planning to do later, for expediency's sake:

* thumbs: use project bytecode in lieu of iframe to realize layout benefits; avoid no-show bug in cover sizing; avoid loading the player so many times
* modals transition in
* thumbnail isLoading state
* Icons for edit, delete, reveal, and duplicate options
* Skull icon to delete project title within confirm modal
* should shake modal if bad project name and show better error state than a toast
* Rip some of these UI pieces out into their own react components; i.e. modals & cards
* In-editor back-to-dashboard button with card icons
* Figure out how to make project creation instantaneous, or show a loader before the... other loader (reticulating splinez)
* Need to look check for errors in doLogOut plumbing call before calling clearAuth in UI
* disabled state for delete btn (it is disabled appropriately but styling does not work for some reason)


packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -834,6 +836,12 @@ export default class Plumbing extends StateObject {
     return inkstone.updates.check(authToken, options, cb)
   }
 
+  doLogOut (cb) {
+    console.log('update')
```
roperzh, line 840: Nit: probably forgotten

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
-    //   return false
-    // }
-
-    projectsList.forEach((foundProject, foundIndex) => {
-      // remove new project if navigating away before it's complete
-      if (projectIndex !== 0 && foundIndex === 0 && foundProject.isNew && !this.state.newProjectLoading) projectsList.shift()
-      if (foundIndex === projectIndex) foundProject.isActive = true
-      else foundProject.isActive = false
+    this.setState({ showDeleteModal: true, projToDelete: name, projToDeleteIndex: index }, () => {
+      this.refs.deleteInput.select()
```
tsortwehttam, line 119: We should keep an eye out for logic that depends on refs, and either add hooks to await for them to be ready, or at least wrap them in conditionals like:

```
if (this.refs.booya) {
  this.refs.booya.doSomething()
}
```

Some of the Sentry exceptions we see on a regular basis are due to race conditions where refs are not present because the React component is in the process of mounting when called.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
-    this.setState({ projectsList })
   }
 
-  unsetActiveProject () {
-    // Hacky way to unset current project
-    this.setActiveProject()
+  performDeleteProject (index) {
+    const projectsList = this.state.projectsList
+    const name = projectsList[index].projectName
+    var atProjectMax
```
tsortwehttam, line 126: Prefer `let` for variables that are re-assigned.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
+
+    return this.requestDeleteProject(name, (deleteError) => {
+      if (!deleteError) {
```
tsortwehttam, line 129: Should we display a toast if deleting a project fails?

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
-        isNew: true,
-        title: ''
-      })
-      setTimeout(() => {
-        this.refs.newProjectInput.select()
-      }, 200)
-      this.setState({ projectsList })
+  showNewProjectModal () {
+    this.setState({ showNewProjectModal: true }, () => {
+      this.refs.newProjectInput.select()
```
tsortwehttam, line 149: Another `ref` to guard.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
+    return (
+      <div style={DASH_STYLES.projectsWrapper}>
+        {this.state.projectsList.map((projectObject, index) => {
+          const project = this.state.projectsList[index]
+          const projectPath = path.join(HOMEDIR_PATH, 'projects', this.props.organizationName, project.projectName)
+          const thumbnail = projectPath + '/preview.html'
+          const hasThumb = fs.existsSync(thumbnail)
+          const standalone = projectPath + '/index.standalone.js'
+          const hasStandalone = fs.existsSync(standalone)
+          if (!fs.existsSync(projectPath) && !project.successfulSessionAdd) {
```
tsortwehttam, line 171: It makes more sense to check for the project path before checking for the files within the folder.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
+              }}>
+              <div id='thumbnail'
+                style={[
+                  DASH_STYLES.thumb,
+                  (project.isMenuActive ||
+                  project.isHovered
+                  ) && DASH_STYLES.blurred
+                ]}>
+                {(hasThumb && hasStandalone) &&
+                  <iframe src={thumbnail}/>
```
tsortwehttam, line 192: We probably want to put `scrolling="no"` on these iframes.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -93,65 +76,207 @@ class ProjectBrowser extends React.Component {
...
+                   style={[
+                    DASH_STYLES.menuOption,
+                    !!!project.isMenuActive && DASH_STYLES.gone]}>
+                    DUPLICATE
+                  </span>*/}
+                  <span key={'delete' + index}
+                    onClick={() => this.showDeleteModal(index)}
+                    style={[
+                      DASH_STYLES.menuOption,
+                      !!!project.isMenuActive && DASH_STYLES.gone]}>
```
tsortwehttam, line 236: Is there a reason you are using `!!!`? (Would not just one `!` suffice?) Or are you just a fan of dance-punk?

```
!truthy === false
!!truthy === true
!!!truthy === false
```

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -234,172 +334,130 @@ class ProjectBrowser extends React.Component {
...
+      <div style={DASH_STYLES.popover.container} onClick={this.closePopover}>
+        <div style={DASH_STYLES.popover.item}>
+          <span style={[DASH_STYLES.popover.text, DASH_STYLES.noSelect]}>{this.props.username}</span>
+        </div>
+        <div style={[DASH_STYLES.popover.item, DASH_STYLES.popover.pointer]}
+          onClick={() => this.logOut()}>
+          <span style={DASH_STYLES.popover.icon}>
+            <LogOutSVG />
+          </span>
+          <span style={DASH_STYLES.popover.text}>LOG OUT</span>
```
tsortwehttam, line 377: Perhaps `text-transform: uppercase` instead of writing it uppercase.

packages/haiku-plumbing/src/Plumbing.js
tsortwehttam, line 840: Remove plz

packages/haiku-creator/src/react/components/ProjectBrowser.js
roperzh: This logic can be changed/simplified by using the `autoFocus` attribute on the delete input, same applies for the new project input.

packages/haiku-creator/src/react/components/ProjectBrowser.js
roperzh: Same comment as above, I think we can use the `autoFocus` attribute and get rid of this.

packages/haiku-plumbing/src/ProjectFolder.js
```diff
@@ -459,7 +459,8 @@ export function buildProjectContent (_ignoredLegacyArg, projectPath, projectName
           <title>${projectNameSafe} | Preview | Haiku</title>
           <style>
             .container { margin: 0 auto; width: 100%; }
-            #mount { width: 80%; margin: 0 auto; }
+            #mount { width: 100%; height: 190px; margin: 0 auto; }
+            body { margin: 0; }
```
tsortwehttam, line 463: I know I said it was ok to just hack the sizing in anywhere, but let me walk that back, because I had a thought on a quick way to make it cleaner.

Why don't we have preview.html accept url query parameters, and set the mount sizing based on that? That way the consumer of preview.html can, for example, load `preview.html?mount-width=100%&mount-height=190px` depending on its needs at any given point.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -234,172 +334,130 @@ class ProjectBrowser extends React.Component {
...
-            style={DASH_STYLES.field}
-            readOnly
-            onChange={this.handleProjectTitleChange.bind(this, projectObject)} />
+      <div style={DASH_STYLES.overlay}
+        onClick={() => this.setState({showNewProjectModal: false})}>
+        <div style={DASH_STYLES.modal} onClick={(e) => e.stopPropagation()}>
+          <div style={DASH_STYLES.modalTitle}>Name Project To Start</div>
+          <div style={DASH_STYLES.inputTitle}>PROJECT NAME</div>
+          <input key='new-project-input'
+            ref='newProjectInput'
```
roperzh, line 397: This way of settings refs [is deprecated](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) and has some issues, we should do this instead:

```
ref={(input) => { this.newProjectInput = input }}
```

packages/haiku-plumbing/src/ProjectFolder.js
tsortwehttam: We could also have preview.html accept a postMessage event which allows the player instance on the page be controlled from the outside. This could actually be handy in a variety of different scenarios...

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -234,172 +334,130 @@ class ProjectBrowser extends React.Component {
...
-        lightScheme={content.lightScheme} />
+      <div style={DASH_STYLES.overlay}
+        onClick={() => this.setState({showDeleteModal: false, projToDelete: ''})}>
+        <div style={DASH_STYLES.modal} onClick={(e) => e.stopPropagation()}>
+          <div style={DASH_STYLES.modalTitle}>
+            Type "<span style={DASH_STYLES.projToDelete}>{this.state.projToDelete}</span>" to confirm project deletion
+          </div>
+          <div style={DASH_STYLES.inputTitle}>DELETE PROJECT</div>
+          <input key='delete-project'
+            ref='deleteInput'
```
roperzh, line 433: This way of settings refs [is deprecated](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) and has some issues, we should do this instead:

```
ref={(input) => { this.deleteInput = input }}
```

tsortwehttam:
@roperzh and I are going to divvy up these issues and address them as bugfixes in master. Merging.

PR #103, opened 2017-12-01
feat: Add haiku-ui-common module.

stristr:
Adds empty `ui-common` package which depends on `react@15.4.2`.


roperzh:
LGTM! :shipit:

PR #104, opened 2017-12-04
Feat publish ui

roperzh:
Publish UI v2.0 bare bones, up to date with latest major changes (`the-exorcism` and `mc`)


stristr:
@roperzh, is this ready for review?

Also, some details on the Inkstone snapshot response: if you poll the snapshot by sha1 (e.g. `https://inkstone.haiku.ai/v0/snapshot/c50f3deebb1f59b98f747d8f74e7970a6c1dd08b`), note the four booleans: `Locked`, `Published`, `Syndicated`, and `Errored`. This is what they mean:

`Locked`: the Git post-receive hook has been received by an Inkstone server, and the publishing process has started.
`Published`: at the very least, content is available on the CDN, so the share link page should actually have content.
`Syndicated`: For now, this just means the GIF is available on the CDN.
`Errored`: Something went wrong, either with CDN upload or with `npm publish`. Whenever an error occurs, we set `Published = false`, even if the CDN upload step succeeded and initially set `Published = true`.

Since the GIF link will be initially broken, at the very least I would recommend polling for a finite amount of time (sometimes GIF'ing fails), and if the status changes to `Errored` before the polling finishes, we should probably let the user know something went wrong.

roperzh:
@stristr Sorry for the lack of clarity, I was finishing up a refactor, but here's the current state of things (taken from commit msg):

This implements the barebones for the Publish UI V2 behind a
feature flag (`NewPublishUI`)

The progress so far:

- New UI implemented
- Moved and modularized all the UI under `haiku-ui-common`

Missing:

- Design pass to integrate with instant publish

So, in short, this is ready for review, but it's incomplete as it needs a UI pass from @taylorpoe

roperzh:
Btw, thanks for the pointers :D

packages/haiku-ui-common/src/react/ShareModal/ProjectShareDetails.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <div style={{width: '50%'}}>
+          <LinkHolster
+            isSnapshotSaveInProgress={isSnapshotSaveInProgress}
+            isProjectInfoFetchInProgress={isProjectInfoFetchInProgress}
+            linkAddress={linkAddress}
+          />
+
+          <p style={assign({textAlign: 'right'}, STYLES.info)}>
+            <span style={assign({}, STYLES.label)}>Last Published</span> Nov.
+            10, 2017
```
tsortwehttam: Is this date intended?

packages/haiku-ui-common/src/react/ShareModal/LinkHolster.tsx
```diff
@@ -0,0 +1,78 @@
...
+    linkAddress: React.PropTypes.string,
+  }
+
+  render () {
+    const {
+      isSnapshotSaveInProgress,
+      isProjectInfoFetchInProgress,
+      linkAddress
+    } = this.props
+console.log(this.props)
```
tsortwehttam: Remove ^

tsortwehttam (changes requested):

tsortwehttam:
Overall looks great. 👍  Really stoked to see shared components in haiku-ui-common 🔥 

Made two small code review notes.

packages/haiku-ui-common/src/react/ShareModal/ProjectShareDetails.tsx
stristr: …said the actress to the bishop.

tsortwehttam:
@roperzh So we don't lose track of this during LOD work, can you summarize what's still needed in this PR before this will be ready to merge?

roperzh:
@matthewtoast there's a new set of features/tweaks that need to be implemented, they are summarized [here](https://haiku.quip.com/fCAiA8upguiY#WXPACAY6v33)

Roughly, this is the current state of things:

- [x] ~Pink dot to publish project~
  - [x] ~Abstract tooltips used in the app into a shared, reusable component~
  - [ ] ~UI and logic to display/publish~
- [x] Loading bars on top of buttons & link holster
  - [x] Component to show progress
  - [x] Tooltips on buttons
- [x] Error messages on fail
  - [x] Add functionality to the `Modal` component to display messages
  - [x] Add logic to display the errors
- [x] Hide "Last Published" and "Version" during publish
- [ ] ~Add logic to show the correct "Last Published" date instead of the hard-coded text~
- [x] Add install instructions for every item
- [x] Remove unused plumbing logic to fetch project info
  
EDIT: ~striketrough~ will not be implemented anymore

tsortwehttam (approved):

zackbrown:
a few requests:

1. can we get some visual feedback when the "copy to clipboard" action occurs?  it's nice to have confirmation that it succeeded (and a little confusing not to have it)
2. For the GIF screen, It'd be great to be able to `right-click > Save image as...` or `right-click > Copy image` — exactly how the browser does it.  would it be possible/easy to handle this natively (i.e. browser image RCM) using electron?  otherwise, if too much effort, we can revisit later.  just feels like a nice natural way to interact with the image
3. Not a huge deal, but it'd be nice to scale the GIF down when it's too big (maybe like a CSS max-height or max-width?), see what happens here
image

zackbrown:
tried to sneak in a little fix, wasn't synced with origin, borked it; heads-up that I'm force-pushing back to the commit you left it at, @roperzh

roperzh:
@zackbrown 👍 

> can we get some visual feedback when the "copy to clipboard" action occurs? it's nice to have confirmation that it succeeded (and a little confusing not to have it)

Does this work for you? Originally we were changing the title of the popover, but that is no longer possible with the new layout.



> For the GIF screen, It'd be great to be able to right-click > Save image as... or right-click > Copy image — exactly how the browser does it. would it be possible/easy to handle this natively (i.e. browser image RCM) using electron? otherwise, if too much effort, we can revisit later. just feels like a nice natural way to interact with the image

After looking into this:

- `right-click > Save image as` is possible but I think it needs some discussion, what happens after that? we need to let the user choose a destination
- `right-click > Copy` I have a prototype but needs a bit of work, I will try to polish it over the weekend.

> Not a huge deal, but it'd be nice to scale the GIF down when it's too big (maybe like a CSS max-height or max-width?), see what happens here

Good catch! done

Thanks for the feedback.

tsortwehttam:
Have reviewed locally. Fixed a couple of issues with the share snippets. Going to merge this.

I moved the RCM idea to its own Asana task: https://app.asana.com/0/539750334128224/555214777981591

PR #105, opened 2017-12-04
fix: use setAttribute() during DOM render instead of setting className directly.

stristr:
Quick fix for https://app.asana.com/0/479779791675933/486086768243940/f.


tsortwehttam:
LGTM.

PR #106, opened 2017-12-04
Run `electron.remote.app.relaunch()` on update

roperzh:
`electron.remote.app.relaunch()` was removed in bf328fa3b as part of a
refactor of the autoupdater. My rationale for removing  it and using the
shell `open` command was to cover users that have launched Haiku before
the autoupdate from a folder that is not `/Applications`, since `relaunch()`
uses by default `process.execPath`.

Turns out that I am dumb enough to not read the full documentation,
`relaunch()` accepts a `execPath` option.

**Notes**

- There are a lot of hard-coded macOs-specific paths in this functionality,
my instinct says that it's good for now, but we should probably take a look
in the future.
- We still need to call `electron.remote.app.exit()` after `relaunch()`, per
the docs:

> Note that this method does not quit the app when executed,
> you have to call app.quit or app.exit after calling app.relaunch to
> make the app restart.

### DIFF
--- a/packages/haiku-creator/src/utils/autoUpdate.js
+++ b/packages/haiku-creator/src/utils/autoUpdate.js
@@ -2,7 +2,7 @@ const qs = require('qs')
 const os = require('os')
 const electron = require('electron')
 const fetch = require('node-fetch')
-const {download, unzipAndOpen} = require('./fileManipulation')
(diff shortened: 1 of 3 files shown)


tsortwehttam:
LGTM.

PR #107, opened 2017-12-04
Use the right data type to ping inkstone

roperzh:
I fixed the source of the problem, but there is nothing to worry about this, we are using this endpoint as a backdoor to report to mixpanel, the autoupdater still uses squirrel-express as our single source of truth.

More info on why we did this in this quip document: https://haiku.quip.com/Hdy2AnRRJQtY


tsortwehttam:
LGTM.

PR #108, opened 2017-12-04
Add logic to deselect selected items on multiselect

roperzh:
This should not require a _lot_ of time to review.

Asana Ticket: https://app.asana.com/0/479779791675933/488080510412730

### DIFF
--- a/packages/haiku-creator/src/utils/autoUpdate.js
+++ b/packages/haiku-creator/src/utils/autoUpdate.js
@@ -27,7 +27,6 @@ module.exports = {
           return reject(Error('Missing release/autoupdate environment variables'))
         }
 
-
         const tempPath = os.tmpdir()
         const zipPath = `${tempPath}/haiku.zip`
         // FIXME: these paths are macOS specific, it's ok for now, but
--- a/packages/haiku-serialization/src/bll/Keyframe.js
+++ b/packages/haiku-serialization/src/bll/Keyframe.js
@@ -72,6 +72,31 @@ class Keyframe extends BaseModel {
     return this
   }
 
+  deselectAndDeactivate () {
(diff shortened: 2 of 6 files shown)


tsortwehttam:
Because this logic can have weird interactions, it's worth double checking if the following still works:

- Adding many curves
- Dragging many keyframes at the same time
- Dragging many segments at the same time
- Right-clicking to change a curve
- Right clicking to add a keyframe
- (etc)

If you've tested and this behavior still looks good after this change, then LGTM.

roperzh:
I will do a double check, but just to know what is the best way to go, when you say

> Because this logic can have weird interactions

You are meaning the way the logic is implemented? or the fact that an element can be deselected on multiselection?

If you think this approach can cause problems down the road I can try something else 👯

tsortwehttam:
I don't mean your logic specifically - I only mean the overall behavior in general related to select/deselect.

(It's just worth checking a few related interactions to make sure those still work as expected. Sometimes an innocuous change actually ends up affecting other things weirdly.)

tsortwehttam:
Just to clarify: This code LGTM and is OK to merge, if you've tested for regressions in the related behavior ^^.

roperzh:
> (It's just worth checking a few related interactions to make sure those still work as expected. Sometimes an innocuous change actually ends up affecting other things weirdly.)

This was spot on, I have found a bug, I will circle back and ping you for a second review.

Thanks!

roperzh:
This is ready for another review, please note that I added a good chunk of code.

I triple-tested everything, but If anyone wants to give it a try before merging, go for it!

tsortwehttam:
LGTM; let's bring this into master. We can circle back to address the couple of hacks you commented .

PR #109, opened 2017-12-05
Timeline bugfixes

tsortwehttam:
WIP; not ready for merge yet.

Just a series of quick bug-fixes that shouldn't require deep code review.

- Create a keyframe 0 when keyframes are created/deleted > 0
- Create a keyframe 0 when the keyframe is dragged
- When the keyframe 0 is deleted, ensure the new keyframe 0 value is the default value
- Ensure the UI updates properly when interior keyframes are created/deleted
- Fix bug causing keyframe curves to be removed on disk but not in UI upon keyframe deletion
- Fix keyframe UI render bugs after pasting an element due to non-unique-at-location React key

UX notes:

- Instead of just "always showing a fake keyframe 0" logic, I'm creating one on-demand. This makes the UI a lot less cluttered. (There is no keyframe 0 unless there really 'needs' to be one due to a user explicitly editing a timeline row.)

- Keyframe drag logic is tricky, and so, on short notice/deadline, the best I can do to accommodate creating keyframe 0 when a keyframe previously-at-0 is dragged, is to do so once the dragging action is complete (debounce fires). The effect is that when the user drags keyframes, only once the action is are complete does keyframe 0 appear. In practice I think this is fine, maybe even the desired UI. But the side effect is that, if the playhead is sitting at time 0, this may result in an on-stage element's property changing dramatically _after_ the movement is done. (For example, since position.x's default value is 0, suppose they dragged property.x's previously-at-0 keyframe which had a value of 456.)

### DIFF
--- a/packages/haiku-bytecode/src/actions/deleteKeyframe.js
+++ b/packages/haiku-bytecode/src/actions/deleteKeyframe.js
@@ -30,14 +30,13 @@ module.exports = function deleteKeyframe (bytecode, componentId, timelineName, p
   }
 
   var prev = list[curr.index - 1]
+  var next = list[curr.index + 1]
 
   // First delete our keyframe
   delete property[keyframeMs]
 
-  // Remove the curve from the previous keyframe since it has nothing to attach to now.
-  // In case of a middle keyframe (between two transitions) assume the user will create a new
-  // transition manually in this case
-  if (prev) {
+  // Remove the curve from the previous keyframe if it has no subsequent keyframe to attach to
+  if (prev && !next) {
     property[prev.start] = {}
     property[prev.start].value = prev.value
     if (prev.edited) property[prev.start].edited = true
--- a/packages/haiku-serialization/src/bll/Row.js
+++ b/packages/haiku-serialization/src/bll/Row.js
@@ -275,6 +275,31 @@ class Row extends BaseModel {
     this.emit('update', 'row-deleted')
   }
 
+  hasZerothKeyframe () {
+    return !!this.getZerothKeyframe()
+  }
+
+  getZerothKeyframe () {
+    return this.getKeyframes().filter((keyframe) => {
+      return keyframe.getMs() < 1
+    })[0]
+  }
+
+  ensureZerothKeyframe (metadata) {
+    if (!this.hasZerothKeyframe()) {
+      this.createKeyframe(this.getFallbackValue(), 0, metadata)
+    }
+  }
+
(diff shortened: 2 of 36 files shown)


tsortwehttam:
I didn't intend for the 2.3.37 updates to go into this PR, but ¯\_(ツ)_/¯

tsortwehttam:
@stristr Ready for review.

packages/haiku-bytecode/src/actions/deleteKeyframe.js
```diff
@@ -30,14 +30,13 @@ module.exports = function deleteKeyframe (bytecode, componentId, timelineName, p
...
 
   // First delete our keyframe
   delete property[keyframeMs]
 
-  // Remove the curve from the previous keyframe since it has nothing to attach to now.
-  // In case of a middle keyframe (between two transitions) assume the user will create a new
-  // transition manually in this case
-  if (prev) {
+  // Remove the curve from the previous keyframe if it has no subsequent keyframe to attach to
+  if (prev && !next) {
```
stristr, line 39: Nice catch, makes sense.

packages/haiku-serialization/src/bll/Row.js
```diff
@@ -275,6 +275,31 @@ class Row extends BaseModel {
...
   }
 
+  hasZerothKeyframe () {
+    return !!this.getZerothKeyframe()
+  }
+
+  getZerothKeyframe () {
+    return this.getKeyframes().filter((keyframe) => {
+      return keyframe.getMs() < 1
+    })[0]
```
stristr, line 285: Although already basically clear, I think `['0']` is clearest we are accessing a property of an object and not a (possibly missing) offset in an array.

packages/haiku-serialization/src/bll/Row.js
```diff
@@ -514,6 +549,15 @@ class Row extends BaseModel {
     return this.property && this.property.name
   }
 
+  getFallbackValue () {
+    if (this.property) {
+      if (this.property.fallback !== undefined) {
+        return this.property.fallback
+      }
+    }
+    return 1 // Possibly safer and more obvious than 0?
```
stristr, line 558: Yes. This is a "this shouldn't happen" situation too, right?

stristr:
LGTM!

packages/haiku-serialization/src/bll/Row.js
tsortwehttam: Correct.

PR #110, opened 2017-12-05
fix: ensure {sizing: "cover"} always applies scaling.

stristr:
This is a fast fix for {sizing: "cover"} that seems to work and should cover edge cases where you'd get a `null` scaling factor causing items to disappear at certain zoom levels/resolutions, mostly where the animation and container width/height are extremely close to each other.

Note: I removed the `~~(scaleDiffX * elementWidth) >= containerWidth`-type conditions because `scaleDiffX` is *defined* as `containerWidth / elementWidth`. I think most edge cases I saw were from cases where this was coming out false purely due to a rounding error.


tsortwehttam:
LGTM.

PR #111, opened 2017-12-06
"Faster" publish

stristr:
Reviewer(s) should ideally run this locally against a local copy of the API changes in [Inkstone #8](https://github.com/HaikuTeam/inkstone/pull/8), but at the very least a very thorough read will be needed.

I had to restart mono so many times that I decided to cherry-pick my (technically still WIP but working) `yarn start` speed-up into here. I'll revert it before merging this if it's not done and everything else is good to go.

This PR is focused on the `saveProject` cluster of functionality, and strips out all the "wait for share link" functionality in favor of shimming in a new, auth-required "create snapshot for project and sha" call as soon as everything has been committed. Because both async call series involved (the one in `MasterGitProject` and the one in `Master`) were looking for the "share info" previously returned at the end of waiting, I opted to leave that basic functionality in place and defer the part where we actually push to a silent and unmonitored process. The idea is that literally every part of waiting, including finding out if publishing worked, will now happen on the share page instead of inside the app—and in the parlance of [Inkstone #8](https://github.com/HaikuTeam/inkstone/pull/8), a Snapshot that was updated five minutes ago but isn't "locked" probably broke on local push.

Also, please note that there is a breaking change in `@haiku/sdk-inkstone` in the *removal* of `snapshot.awaitSnapshotLink`, now called `snapshot.getSnapshotLink`. I tried to leave the API as stable as possible other than that (e.g. it basically does the same thing, except without "awaiting").

### DIFF
--- a/packages/haiku-plumbing/dev/gulp.js
+++ b/packages/haiku-plumbing/dev/gulp.js
@@ -6,8 +6,8 @@ var gulp = require('gulp')
 var watch = require('gulp-watch')
 var path = require('path')
 var exec = require('child_process').exec
-var execSync = require('child_process').execSync
-var spawn = require('child_process').spawn
+
+const argv = require('yargs').argv
 
 var ROOT = path.join(__dirname, '..')
 var SOURCE = './src/'
@@ -31,33 +31,11 @@ function cmd(command, cb) {
   proc.stdout.pipe(process.stdout);  
 }
 
-function scmd(command, args, cb) {
-  console.log('[gulp] Spawning `' + command + ' '+ args.join(' ') +'`')
-  var proc = spawn(command, args, { cwd: ROOT })
-  proc.stdout.pipe(process.stdout);
-  return cb(null, proc)
-}
-
 function copyFile(relpath, cb) {
   console.log('[gulp] Copying `' + relpath)
   var abssrc = path.join(ROOT, SOURCE, relpath)
   var absdest = path.join(ROOT, DEST, relpath)
   return fse.copy(abssrc, absdest, cb)
-}
-
-var SAY_VOICE = process.env.SAY_VOICE || 'Alex'
-var SAY_SPEED = process.env.SAY_SPEED || 300
-function sayAudibly(message) {
-  try {
-    if (!message) return void (0)
-    message = message.split('\n')[1]
-    if (!message) return void (0)
-    message = message.split(/\s+/).slice(0, 5)
-    message = 'Air Roar! ' + message
-    execSync('say -v ' + JSON.stringify(SAY_VOICE) + ' -r ' + SAY_SPEED + ' ' + JSON.stringify(message))
-  } catch (exception) {
-    // empty
-  }
 }
 
 function compileAll(cb) {
@@ -106,24 +84,22 @@ function watchFiles(cb) {
   })
 }
 
-function flagsToArgs(flags) {
-  var args = []
-  each(flags, (value, key) => {
-    if (value === true) return args.push('--' + key)
-    else args.push('--' + key + '=' + value)
-  })
-  return args
-}
-
 gulp.task('compile', function _compile(done) {
   return compileAll(done)
 })
 
 gulp.task('watch', function _watch(done) {
+  const watch = () => watchFiles(function _watch(err) {
+    if (err) return done(err)
+  })
+
+  if (argv['skip-initial-build']) {
+    console.log('[gulp] Skipping initial build.')
+    return watch()
+  }
+
   return compileAll(function _compile(err) {
     if (err) return done(err)
-    return watchFiles(function _watch(err) {
-      if (err) return done(err)
-    })
+    return watch()
   })
 })
--- a/packages/haiku-plumbing/src/Master.js
+++ b/packages/haiku-plumbing/src/Master.js
@@ -432,7 +432,7 @@ export default class Master extends EventEmitter {
   fetchProjectInfo ({ params: [projectName, haikuUsername, haikuPassword, fetchOptions = {}] }, cb) {
     return this._git.fetchFolderState('fetch-info', fetchOptions, (err) => {
       if (err) return cb(err)
-      return this._git.getCurrentShareInfo(2000, cb)
+      return this._git.getCurrentShareInfo(cb)
     })
   }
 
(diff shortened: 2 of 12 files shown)


roperzh:
Hello! I was going to review/test this, but I noticed the title has changed to "WIP".

It is safe to pick the current state of things? I want to integrate with the new publish ui asap in order to avoid conflicts.

tsortwehttam:
@roperzh Since Sasha will be offline for a part of today I'll answer: I think it's probably fine to pick this in to your branch, although as far as I know there's still a few hours of work left on it. 👍

stristr:
Eh, heads up @roperzh, this branch has been rebased to master. There actually hasn't been any pending work on it, except for low-urgency improvements to "faster `yarn start`".

stristr:
Ready for review!

packages/haiku-plumbing/dev/gulp.js
```diff
@@ -31,35 +31,13 @@ function cmd(command, cb) {
...
-    message = message.split('\n')[1]
-    if (!message) return void (0)
-    message = message.split(/\s+/).slice(0, 5)
-    message = 'Air Roar! ' + message
-    execSync('say -v ' + JSON.stringify(SAY_VOICE) + ' -r ' + SAY_SPEED + ' ' + JSON.stringify(message))
-  } catch (exception) {
-    // empty
-  }
-}
-
```
tsortwehttam, line 62: Nooooooooooooo

packages/haiku-plumbing/src/Master.js
```diff
@@ -1045,33 +1045,39 @@ export default class Master extends EventEmitter {
...
 
-      (cb) => {
-        return this._git.commitProjectIfChanged('Populated content', cb)
-      },
-
       // Now do all of the git/share/publish/fs operations required for the real save
       (cb) => {
-        logger.info('[master] project save: committing, pushing, publishing')
-        return this._git.saveProject(saveOptions, cb)
+        logger.info('[master] project save: creating snapshot')
```
tsortwehttam, line 1065: Nit: It's a bit safer to `return asyncThing(cb)` than to just call `asyncThing(cb)`, because it makes it less likely that the statement will be misread as a standard procedural call. It makes it more obvious that we're passing a continuation around and that the async flow of control has jumped elsewhere, i.e. "don't add further statements after this guy, add them at the next step that receives the continuation"

tsortwehttam:
This looks good so far from a code POV, but I'm going to do some tire-kicking locally too.

packages/haiku-sdk-inkstone/src/index.ts
```diff
@@ -40,7 +41,7 @@ export namespace inkstone {
     _.extend(_inkstoneConfig, newVals)
 
     //ease SSL restrictions for dev
-    if (newVals.baseUrl.indexOf("https://localhost") === 0) {
+    if (newVals.baseUrl && newVals.baseUrl.indexOf("https://localhost") === 0) {
```
roperzh, line 44: Nit: we can use [`String.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) here for more clarity:

```javascript
if (newVals.baseUrl && newVals.baseUrl.includes("https://localhost")) {
  //
}

File: scripts/start.js [hunk 4]
```diff
@@ -251,127 +146,98 @@ function setup () {
...
-    // Only set the mock envoy variable if we are in an env where it will work
-    process.env.MOCK_ENVOY = inputs.mockEnvoy
-    if (inputs.devChoice === 'glass') {
-      instructions.unshift(['haiku-glass', ['yarn', 'start'], null, 5000])
-      instructions.unshift(['haiku-glass', ['yarn', 'run', 'watch', ...watchOptions], null, 5000])
-    } else if (inputs.devChoice === 'timeline') {
-      instructions.unshift(['haiku-timeline', ['yarn', 'start'], null, 5000])
-      instructions.unshift(['haiku-timeline', ['yarn', 'run', 'watch', ...watchOptions], null, 5000])
-    }
-  }
```
tsortwehttam, line 314: Removing the `devChoice` switch breaks my workflow. I need to be able to:

- Develop the entire app, starting at the dashboard
- Develop the entire app, starting up within a known fixture project (or a blank project)
- Develop glass _in isolation, without starting up plumbing etc_ 
- Develop timeline _in isolation, without starting up plumbing etc_

Those last two items are what this seems to break. Previously, I was able to just:

```
$ yarn start glass
```

Which would launch glass (only the glass Electron app), automatically displaying a fixture project.

Can we reinstate that?

stristr: Sorry for not clearly stating: this is known and on the list of reasons I noted this `yarn start` refactor is still WIP in the description. The other main reason is that `.last-compile` doesn't get updated by watchers, so you end up recompiling packages that don't need it without a choice to opt out over multiple starts. If I don't get around to fixing these before this PR is done, I'll revert it from the merge. Meanwhile, I'm just keeping that WIP commit in this branch because it's handy during development.

packages/haiku-plumbing/dev/gulp.js
stristr: 🤣

packages/haiku-sdk-inkstone/src/index.ts
stristr: Hmmm, I didn't really look at the code (just updated it to not crash on configs that don't override `baseUrl`), but [String.prototype.startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) would be the best, right? (Technically `'foo https://localhost'.includes('https://localhost')`….)

But either way, thanks for pointing out this line, because now that I think about it, it really should be better (e.g., why not support https://127.0.0.1?). I think I'll opt for moving it to a higher scope rather than redefining `request` when we get a new config…something like

```
 const request = requestLib.defaults({
   strictSSL: !isDevelopmentMode()
 })
```

Just would need to figure out how to do `isDevelopmentMode()` across all the various app- and browser-targeted code that uses `@haiku/sdk-inkstone`, but not too bad. I'll make a deeply backlogged ticket for this.

stristr: (In [Asana](https://app.asana.com/0/480796620059195/499239569411979/f).)

scripts/start.js
stristr: Also: I should have said, "Ready for review, but please don't merge yourself!" :-)

(In addition to breaking devChoices, we need to deploy sumi-e and share page updates before this goes in a release. [Nothing would *break* if we didn't, but the problem with the CDN caching 404s for a long time would get suddenly much worse.])

tsortwehttam: 👍

tsortwehttam:
LGTM, but will leave it to @stristr to merge this.

(We can circle back later on the tooling issue I mentioned.)

stristr:
All downstream dependencies are deployed, so shipping this.

PR #113, opened 2017-12-06
fix: Ensure correct first rendering on Dashboard v2 thumbnails.

stristr:
Tiny fix to load the Dashboard v2 previews directly from bytecode instead of iframe'ing in preview.html.

Three main advantages:
 - Existing projects benefit from the performance optimizations in Player we recently introduced.
 - Existing projects have `sizing: 'cover'`.
 - Only one global Haiku animation frame, so probably a much lower overall footprint regardless of perf optimizations.


tsortwehttam:
LGTM and checked locally and reviewed with team; merging.

PR #114, opened 2017-12-06
Throttle carbonite uploads

roperzh:


packages/haiku-serialization/src/utils/carbonite.js
```diff
@@ -100,7 +109,11 @@ function crashReport (orgName, projectName, projectPath) {
     )
     .then(() => _getPreSignedURL(zipName))
     .then(AWS3URL => _upload(AWS3URL, zipPath))
-    .catch(error => console.log(error))
+    .catch(error => {
+      console.log(error)
+      // If the upload has failed, allow upcoming errors to be uploaded
+      lastUploadTime = null
```
stristr: Actually, should we?

The main thing @matthewtoast was worried about was looping on `tar` eating system resources. I think it's better to not reset `lastUploadTime` on errors (i.e. leave as it was).

stristr (changes requested):

packages/haiku-serialization/src/utils/carbonite.js
tsortwehttam: Agreed - error or not, I believe we should still throttle.

roperzh: Done!

PR #115, opened 2017-12-06
tweak: Put dashboard previews in edit mode, to suppress possible eventHandler and summonable weirdness.

stristr:
API already provides this option and edit mode already has been tested, so a cursory review is fine.


tsortwehttam:
LGTM, and merging.

PR #117, opened 2017-12-06
WIP: Tour fixes following Dashboard 2 changes

tsortwehttam:


roperzh:
I pushed a couple of changes fixing the broken step and the position of the popup, so the tour is working now.

There's a known issue product of using absolute positioning (a necessary evil due to having three webviews) which is causing the popup to go off of the screen on certain screen configurations:

screen shot 2017-12-06 at 23 14 01

I have a couple of possible fixes in mind, but I think we should implement a well rounded solution for the tour since it's going to hit us back in the long term. I will leave the decision to when to implement this to Matthew. 

It will also need a better copy.

stristr:
Please hold off on merge, I'm all over this file working on [Dashboard many renders * much fs = perf crater](https://app.asana.com/0/479779791675933/497228471686486/f). I will try to cherry-pick these commits in when done.

roperzh:
@stristr understood, if you have any difficulties moving the changes to your branch I can apply them again once you merge.

stristr:
Replaced by #118.

PR #118, opened 2017-12-07
WIP: Dashboard 2 fixes

stristr:
Replaces #117.

The work of [Dashboard many renders * much fs = perf crater](https://app.asana.com/0/479779791675933/497228471686486/f) should be complete inside of here. Tour fixes were conflicting so were pulled in here.

From @roperzh:

I pushed a couple of changes fixing the broken step and the position of the popup, so the tour is working now.

There's a known issue product of using absolute positioning (a necessary evil due to having three webviews) which is causing the popup to go off of the screen on certain screen configurations:

screen shot 2017-12-06 at 23 14 01

I have a couple of possible fixes in mind, but I think we should implement a well rounded solution for the tour since it's going to hit us back in the long term. I will leave the decision to when to implement this to Matthew. 

It will also need a better copy.


packages/haiku-sdk-creator/src/tour/TourHandler.ts
```diff
@@ -178,13 +170,6 @@ export class TourHandler implements Tour {
     }
   }
 
-  private requestSelectProject() {
-    this.server.emit(TOUR_CHANNEL, <EnvoyEvent> {
-      payload: {},
-      name: 'tour:requestSelectProject',
-    });
-  }
-
```
tsortwehttam, line 187: @roperzh  Before I ran out of energy last night looking at this, I saw an issue related to this and sort of gave up. :) When you have a chance can you explain what was going on here, just in case useful in the future?

tsortwehttam:
LGTM, merging.

packages/haiku-sdk-creator/src/tour/TourHandler.ts
roperzh: @matthewtoast yes sorry for the obscurity of this piece of code, at the very least the naming could have been better!

In the old dashboard ui, in order to open a project you'd have to:

1. Select the project from the list on the left
2. Click on the "Open Project" button

This logic is used to do step `1`, and ensure the tour project is selected.

PR #119, opened 2017-12-07
Fixes roberto dec7

roperzh:
- Fixes the styling of multiple UI buttons (tour, autoupdater, sketch downloader, etc), we should get back to those and update with the new black buttons. I decided to keep the pink ones because there are some styling considerations to be addressed.

- Fixes issues with the autoupdater


tsortwehttam:
LGTM and merging.

PR #120, opened 2017-12-07
Fix issues on external assets dropping in Creator

roperzh:
- [x] Remove react-specific hacks used in dndHelpers.js
- [x] Fix an issue where the project folder was `undefined` when
dragging assets into creator


tsortwehttam:
LGTM, merging.

PR #122, opened 2017-12-07
Build fixes

tsortwehttam:
Should be a quick review.

Ultimately the way to verify is to run a build and explore the contents of the built app to ensure the exclusions work and that the app still functions as expected.


stristr (approved):
LGTM.

PR #124, opened 2017-12-08
fix: Hack into Node's module loader to bypass require() in dash thumbs.

stristr:
Heavy on the hack but should be a quick and readable CL.

So, turns out we were still using the version of player that was `npm install`'ed in the local project folder in dashboard thumbs because we were using `require()`. (See [Project dashboard thumbnails not appearing](https://app.asana.com/0/479779791675933/497587971544080/f) for background.) ~~This is probably at the root of the spinning fans we experienced during testing, so that's fixed here too!~~*

There's a better/simpler way, and as a bonus we don't even have to mess around with the `require()` cache—just compile the module directly using the mostly undocumented and private module API!

Although this might seem scary, it really seems to be fine, and produces the intended effect.

*Actually, shouldn't mean much for perf. We were only using the project's player for (maybe) its `Haiku.inject`.

### DIFF
--- a/packages/haiku-creator/src/react/components/ProjectPreview.js
+++ b/packages/haiku-creator/src/react/components/ProjectPreview.js
@@ -1,19 +1,46 @@
+import fs from 'fs'
+import Module from 'module'
 import React, { PropTypes } from 'react'
+
 import HaikuDOMAdapter from '@haiku/player/dom'
 import {InteractionMode} from '@haiku/player/lib/helpers/interactionModes'
+
+const renderMissingLocalProjectMessage = (projectName) => {
+  switch (projectName) {
+    case 'CheckTutorial':
+      return (
+        <p>Click to load tutorial project</p>
+      )
+    case 'Move':
+    case 'Moto':
+      return (
+        <p>Click to load sample project</p>
+      )
+    default:
+      // TODO: Do we want to display a message or anything else if the project isn't
+      // already present locally?
+      return <p />
+  }
+}
 
 class ProjectPreview extends React.Component {
   constructor (props) {
     super(props)
     this.bytecode = null
     this.mount = null
+    // TODO: Try to get the bytecode from CDN or eager clone if not yet available.
+    this.hasBytecode = fs.existsSync(props.bytecodePath)
+    if (!this.hasBytecode) {
+      return
+    }
+
     try {
-      if (require.cache.hasOwnProperty(props.bytecodePath)) {
-        delete require.cache[props.bytecodePath]
-      }
-      this.bytecode = require(props.bytecodePath)
+      const bytecode = new Module('', module.parent)
+      bytecode.paths = Module._nodeModulePaths(global.process.cwd())
+      bytecode._compile(fs.readFileSync(props.bytecodePath).toString(), '')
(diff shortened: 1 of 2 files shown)


packages/haiku-creator/src/react/components/ProjectPreview.js
```diff
@@ -1,19 +1,46 @@
...
+      return
+    }
+
     try {
-      if (require.cache.hasOwnProperty(props.bytecodePath)) {
-        delete require.cache[props.bytecodePath]
-      }
-      this.bytecode = require(props.bytecodePath)
+      const bytecode = new Module('', module.parent)
+      bytecode.paths = Module._nodeModulePaths(global.process.cwd())
```
tsortwehttam, line 39: In prod, what is `cwd` going to be, and is that going to affect how this works?

packages/haiku-creator/src/react/components/ProjectPreview.js
```diff
@@ -1,19 +1,46 @@
...
+      bytecode._compile(fs.readFileSync(props.bytecodePath).toString(), '')
```
tsortwehttam, line 40: Can you explain a bit about what this is really doing? It was my impression that this is basically what `require()` does under the hood anyway, so what advantage do we get from doing these steps ourselves (vs just clearing the require cache and calling require)?

packages/haiku-creator/src/react/components/ProjectPreview.js
tsortwehttam: I.e. should this be `__dirname` or `path.join(path_to_app_root)` or something?

tsortwehttam:
Going to merge this to get it in for testing in a prod build but would still be interested in the answers to the questions above ^^

packages/haiku-creator/src/react/components/ProjectPreview.js
stristr: Not 100% sure of the directory structure semantics but I suspect the packed asar virtualizes a filesystem but never actually creates it. (This is something I'd like to learn more about when we start taking on app size aggressively.) Anyway, `global.process.cwd()` *should* resolve to the virtual equivalent of "the directory where `node_modules` sits", and I guess we'll find out if I'm wrong when the release goes up.

If it happens not to work, I would try `path.dirname(__dirname)` first.

packages/haiku-creator/src/react/components/ProjectPreview.js
stristr: Fun story! So the `require()` module loader is a little too smart about looking for module roots, so the `var Haiku = require("@haiku/player");` was actually checking the project folder's `node_modules` for the player it uses to (maybe) `Haiku.inject` stuff.

This was fine on local dev, because the `npm install` step seems to be working there, but broken on projects that had only been opened in a release, because on releases (separate issue, I didn't really look into it) `npm install` seems to not be working. This was at the heart of Nad not seeing previews and you not seeing previews update when you tested on the release.

So, this code is basically doing what `require()` does, except bypassing all of its smart module root detection and telling the module loader to look for `@haiku/player` in the same place the React app does. Works great on local dev, and I *thought* it felt a little speedier from not loading N extra Haiku objects.

PR #125, opened 2017-12-08
Fix: project card doesn't show if proj name is "loader" 

roperzh:
This should only take a couple of minutes to review.

Asana ticket: [Dash bug: project card doesn't show if proj name is "loader" ](https://app.asana.com/0/479779791675933/498221072732921)

### DIFF
--- a/packages/haiku-creator/public/stylesheets/global.css
+++ b/packages/haiku-creator/public/stylesheets/global.css
@@ -129,14 +129,6 @@ iframe {
   -webkit-transition: opacity .25s ease-in-out;
 }
 
-#loader {
-  position:absolute;
-  top:50%;
-  left:50%;
-  transform: translateX(-50%) translateY(-50%);
-  text-align:center;
-}
-
 .frame {
   -webkit-user-select: none;
   -webkit-app-region: drag;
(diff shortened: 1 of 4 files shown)


stristr:
LGTM and I don't see any things that actually use an id of "loader", but should get a 👍 from somebody who has been with the code since the beginning to be sure.

packages/haiku-creator/public/stylesheets/global.css
```diff
@@ -129,14 +129,6 @@ iframe {
   -webkit-transition: opacity .25s ease-in-out;
 }
 
-#loader {
```
roperzh, line 132: @matthewtoast any bad feelings about removing this? I traced it back to https://github.com/HaikuTeam/glass/commit/d643796baefbcd05804a5b4b2b818950d7513a45#diff-1804b38ac46271db28843ef316579fe6 but couldn't find where this was applied.

After a quick test, every loader in the app seems to be working fine

tsortwehttam:
@roperzh That is a good change, but we should go a step further and namespace this:

https://github.com/HaikuTeam/mono/blob/master/packages/haiku-creator/src/react/components/ProjectThumbnail.js#L24

So the ids end up like `haiku-project-thumbnail-FooBar` and `haiku-project-thumbnail-Meow` - which will prevent future accidental collisions.

roperzh:
@matthewtoast I did it in https://github.com/HaikuTeam/mono/pull/125/commits/0d0d785e94a0d8f57757675ad3da9e47115c5552 do you want to change the prefix? If yes I'm in!

tsortwehttam:
Nope that's great. Missed that. LGTM.

PR #126, opened 2017-12-08
Fixtures for template projects

stristr:
Medium-small PR, but feel free to ignore the copy/pasted bytecode.

After discussing with @zackbrown last night: since first touch is so important and we didn't want to treat the template projects **too** differently in dashboard 2, we can for now store preview bytecode for the template projects directly in the app so they're available as a fallback. This is extensible if needed, but not meant to be a permanent solution or anything—just for getting dashboard 2 out the door without corrupting the essential first touch with ugly, inconsistent, or missing previews.

I also opportunistically cleaned up the brittle @haiku/zack2-checktutorial dependency by using another fixture for that.


stristr:
Now featuring "play dashboard 2 Haiku on hover".

tsortwehttam:
LGTM.

PR #127, opened 2017-12-09
WIP: 250MB

tsortwehttam:
This is just the result of some Friday night hacking. The objective was to reduce the app footprint "somewhat". With these changes, Haiku.app is at ~250MB. (As opposed to 350 or 400 or 700 (!))

Although it seems to work fine (I've built the app locally and run it through some basic QA testing without any problem), it's pretty hacky, so it warrants review. (Note: If the verdict is "This is too hacky" or "I have a much better solution cooking that will land before New Years Day") then we can just simply toss this in the bin; this an after-hours project and didn't take a lot of time.)

Background: In Haiku.app, the entry point is `source/plumbing`. And plumbing holds all of the `haiku-*` subpackages. Previously, each `haiku-*` subpackage had its own node_modules folder.

What this PR does is move all dependencies up to source/plumbing, or, which has the effect of deduping them. So instead of multiple lodash copies across the app, we have just one. And then there are a a few hacks to work around some side-effects of this.

Risks:

- Suppose `haiku-foo` depends on `meow@1.2.3` and `haiku-bar` depends on `meow@4.5.6`. That's not accounted for at all by this. This may already be a problem when developing locally with mono, i.e. we may already "secretly" have this problem, so ¯\_(ツ)_/¯

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ config.js
 source/*
 ~tmp
 .last-sync
+.last-compile
--- a/build.sh
+++ b/build.sh
@@ -5,7 +5,6 @@ node ./scripts/distro-configure.js --non-interactive # 0m
 node ./scripts/distro-download-secrets.js
 yarn relink
 node ./scripts/distro-prepare.js # 3m
-node ./scripts/distro-electron-rebuild.js # 5m
 node ./scripts/distro-uglify-sources.js # 1m
 node ./scripts/distro-build.js # 8m
 node ./scripts/distro-upload.js # 2m
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "mac": {
       "category": "public.app-category.developer-tools"
     },
-    "asar": true,
+    "asar": false,
     "asarUnpack": [
       "index.js",
       "config.js",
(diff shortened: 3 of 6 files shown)


stristr:
I'm diggin' it! I'm going to test it locally carefully just to make sure it works like it apparently should, will merge this weekend if no issues.

roperzh:
I can also help to test over the weekend if needed (will need a build from Travis, afaik I can't build locally).

Wondering if we can also do tree shaking on the resulting `node_modules` folder

edit: grammar

stristr:
LGTM with (major!) caveats below.

Recap of weekend testing: no major alarms, and agree with the one stated risk—but I do find it a little disturbing since we're only running deps this way in releases, and it's the nature of the beast that we spend way less time running releases than running local dev. I am concerned that even exhaustive testing of a release might not produce an obscure, edge-case bug from incompatible versions of sub-sub-dependencies.

I didn't quite fire up the oven yet, but I would like to seriously look into getting "a much better solution cooking that will land before New Years Day" in the form of Yarn workspaces (h/t to @roperzh for originally suggesting it). The idea is essentially to do what you're doing here, except letting yarn resolve potentially conflicting versions and having local and release resolve deps in exactly the same way. Given the sheer amount of third party code in our project, I'm not convinced that a package requiring `^1.0.x` _wouldn't_ crash in an obscure edge case on `1.0.[x+1]`, so it's not like having yarn resolve conflicts is necessarily better than rolling the dice on whichever version wins based on the order of rsyncs here…but still, I would feel a _lot_ better knowing that the deps we use in dev are the same as the deps we use in releases.

Long story short: if we decide to put this in the next release, we should test the holy hell out of it (i.e. the next release shouldn't be today).

tsortwehttam:
Going to merge this and we can revert if we see further problems.

stristr:
Note to self: I'm coming around to almost wanting to revert this after spending an unexpected day tracing inability to build after pushing a new semver from the same codebase to the fact that we're trying to rsync our own node dependencies instead of letting `npm`/`yarn` do their thing.

Think I have it figured out, but probably time to cut this loose if the next hack doesn't hack it.

PR #128, opened 2017-12-10
Fix "return" inside event handler function breaks events UI

stristr:
Slightly complex review, but short.

I caught a bad bug in events UI due to how we prettify the handler code before showing it to the user. tl;dr, if we try to pass a function body through babylon intact, we'll get a compiler error if the function uses the keyword `return` anywhere—`return` and `return x` are not valid expressions outside of a function's scope.

Without fixing this, the UI loops on the syntax error it catches while trying to prettier the function body. Unfortunately, if we want to beautify the user's code using a formatter that depends on AST, we need to make it look like a complete function, not a list of expressions.

This PR provides a relatively painless solution: wrap the function body in `() => {` ... `}`, then format it, then return the resulting outdented function body.

Please see [User is allowed to save event handlers with syntax errors](https://app.asana.com/0/479779791675933/499273135535577/f) and [Monaco syntax errors don't display until additional keypress](https://app.asana.com/0/479779791675933/499273135535576/f) for related problems we should fix.

### DIFF
--- a/packages/haiku-common/config/experiments.json
+++ b/packages/haiku-common/config/experiments.json
@@ -1,14 +1,5 @@
 {
   "LottieExportInGlobalMenu": {
     "development": true
-  },
-  "LottieExportOnPublish": {
-    "development": true,
-    "production": true
-  },
-  "PreviewMode": {
-    "test": true,
-    "development": true,
-    "production": true
   }
 }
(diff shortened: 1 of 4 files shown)


stristr:
(Note, this PR also includes a chore commit to retire successful experiments.)

tsortwehttam:
Not sure if you found this, but I believe Babylon has a `allowReturnOutsideFunction` option. Whether we can use it or not without causing other problems would of course be an open question, but I wanted to mention it.

This looks ok to me, but @roperzh should do a more thorough review since I know he was battling with code formatting at one point.

roperzh:
LGTM! Thanks for handling this case

PR #129, opened 2017-12-11
Force users to solve syntax errors in events ui

roperzh:
This adds logic to avoid saving event handlers with syntax errors
by checking if there are editors with syntax errors and scrolling
to them if the user has pressed on 'save'

Asana task: https://app.asana.com/0/479779791675933/499273135535577

### DIFF
--- a/packages/haiku-glass/src/react/components/EventHandlerEditor/EditorActions.js
+++ b/packages/haiku-glass/src/react/components/EventHandlerEditor/EditorActions.js
@@ -34,6 +34,7 @@ class EditorActions extends React.PureComponent {
         <button
           onClick={this.props.onSave}
           style={{...STYLES.button, ...STYLES.doneButton}}
+          title={this.props.title}
         >
           Done
         </button>
--- a/packages/haiku-glass/src/react/components/EventHandlerEditor/SyntaxEvaluator.js
+++ b/packages/haiku-glass/src/react/components/EventHandlerEditor/SyntaxEvaluator.js
@@ -4,7 +4,7 @@ import {EVALUATOR_STATES} from './constants'
 import Palette from '../../Palette'
 import HaikuMode from '../../modes/haiku.js'
 
-class SyntaxEvaluator extends React.Component {
+class SyntaxEvaluator extends React.PureComponent {
   constructor (props) {
     super(props)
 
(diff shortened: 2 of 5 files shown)


packages/haiku-glass/src/react/components/EventHandlerEditor/SyntaxEvaluator.js
```diff
@@ -4,7 +4,7 @@ import {EVALUATOR_STATES} from './constants'
 import Palette from '../../Palette'
 import HaikuMode from '../../modes/haiku.js'
 
-class SyntaxEvaluator extends React.Component {
+class SyntaxEvaluator extends React.PureComponent {
```
stristr, line 7: :smile: 👍

packages/haiku-glass/src/react/components/EventHandlerEditor/EditorActions.js
```diff
@@ -34,6 +34,7 @@ class EditorActions extends React.PureComponent {
         <button
           onClick={this.props.onSave}
           style={{...STYLES.button, ...STYLES.doneButton}}
+          title={this.props.title}
```
stristr, line 37: Nit: for extra protection, could also add `disabled={this.props.title !== ''}`

stristr: (Actually n/m, disabling would prevent the click listener from scrolling to the error if it exists.)

So if anything might want to put a style, like `cursor: not-allowed` here.

packages/haiku-glass/src/react/components/EventHandlerEditor/index.js
```diff
@@ -200,7 +230,11 @@ class EventHandlerEditor extends React.PureComponent {
         />
 
         <div style={STYLES.outer}>
-          <div style={STYLES.editorsWrapper} className='haiku-scroll'>
+          <div
+            style={STYLES.editorsWrapper}
+            className='haiku-scroll'
+            ref={(el) => (this.wrapper = el)}
```
stristr: Nit: lint error here (h/t to @matthewtoast who fixed it when I did this last week). Should be wrapped in braces, not parens, i.e. don't return the result of an assignment.

stristr:
LGTM!

roperzh:
Thanks for the review!

PR #135, opened 2017-12-13
Fix hide handlers editor

roperzh:

Asana ticket: https://app.asana.com/0/479779791675933/501102857337893

**Concerns**

Matthew added this conditional a few days ago, and his commit states he had to do it because his fan was spinning like crazy when opening a project. Well, this disables the performance enhancement at all in preview mode...

The performance looks good to me, but I think the rationale stated above is still valid.

If anyone has better ideas, please shout


tsortwehttam:
LGTM - we'll try it and see how it handles 🌗

PR #136, opened 2017-12-13
Move Palette, Menu, and a few shared helpers into ui-common

tsortwehttam:
This moves some common UI code into ui-common, where it belongs. This should not take long to review, as there are no behavior changes here.

Motivation:

Help me build parts of the multi-component UI that leverage shared utils.

Summary:

- Move all Palette.js files into ui-common, consolidate the colors, and verify the colors look ok
- Move all JSX SVG icons into ui-common, and verify they look ok
- Move some additional UI helper utilities into ui-common, and verify they work
- Convert all of the above into TypeScript
- Add minor modification to the tsconfig.json and tslint.json

Notes:

- There are currently two lint errors that need to be fixed in ui-common (these happened many times over which is why I haven't fixed them yet):

```
import-name    Misnamed import. Import should be named 'react' but found 'React'
variable-name  variable name must be in lowerCamelCase or UPPER_CASE
```

- Please note the changes to tsconfig.json and tslint.json which I did in order to expedite getting these changes together without having to wrangle for too long on tooling.

I would like to merge this ASAP and circle back and fix any code convention issues later.


tsortwehttam:
We may also want @taylorpoe to do a once-over just to double check that things are looking ok.

The Palette now lives in https://github.com/HaikuTeam/mono/blob/6df07a2196762172608e469e200d111f1cff5145/packages/haiku-ui-common/src/Palette.ts

packages/haiku-creator/src/electron.js
```diff
@@ -171,7 +171,7 @@ function createWindow () {
   })
 
   // Uncomment me to automatically open the tools
-  // browserWindow.openDevTools()
+  browserWindow.openDevTools()
```
roperzh: Double checking if we want to leave this un-commented.

tsortwehttam: Nice catch, and fixed

roperzh:
The code LGTM! I am pulling and yarn-installing as I type this in order to do QA

packages/haiku-ui-common/tslint.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": [
     "tslint-config-haiku"
-  ]
+  ],
+  "rules": {
+    "max-line-length": 180
```
roperzh, line 6: meta: any thoughts on including prettier into our workflow? We can stop worrying for things like this or brackets and similar things.

packages/haiku-ui-common/tslint.json
```diff
@@ -1,5 +1,8 @@
...
+  }
```
stristr, line 7: Would prefer a:

/* tslint-disable:max-line-length */ at the top of files it's too much work to reformat right now.

roperzh:
Things are looking great for me, ~the only bug I found is the user icon on the dashboard, seems like it  has an incorrect tone of white.~

Edit: nvm, seems like the color is the same in master, sorry!

packages/haiku-ui-common/src/helpers/doPositionsMatch.ts
```diff
@@ -0,0 +1,7 @@
+function doPositionsMatch(p1, p2) {
+  if (!p1) { return false; }
+  if (!p2) { return false; }
+  return ~~p1.x === ~~p2.x && ~~p1.y === ~~p2.y;
```
stristr, line 4: Nit: `if (!p1 || !p2)`

stristr: Re-nitted: `return p1 && p2 && ~~p1.x === ~~p2.x && ~~p1.y === ~~p2.y;`

packages/haiku-ui-common/src/helpers/ensureEq.ts
```diff
@@ -0,0 +1,10 @@
+import retToEq from './retToEq';
+import * as EXPR_SIGNS from './ExprSigns';
+
+export default function ensureEq(str) {
+  let fixed = retToEq(str);
+  if (fixed.slice(0, 1) !== EXPR_SIGNS.EQ) {
+    fixed = EXPR_SIGNS.EQ + fixed;
+  }
+  return fixed;
+}
```
stristr, line 10: Nit (and apologies for commenting on not-new code ported to TS): `const` rewrite would be:

```
export default function ensureEq(str) {
  const fixed = retToEq(str);
  if (fixed.slice(0, 1) !== EXPR_SIGNS.EQ) {
    return EXPR_SIGNS.EQ + fixed;
  }
  return fixed;
}
```

packages/haiku-ui-common/src/helpers/doesValueImplyExpression.ts
```diff
@@ -0,0 +1,10 @@
+import leftTrim from './leftTrim';
+import * as EXPR_SIGNS from './ExprSigns';
```
stristr, line 2: Since we're using allowSyntheticDefaultImports this could just be `import ExprSigns from './ExprSigns';`

Would prefer it for internal consistency, but not a huge deal.

packages/haiku-timeline/src/components/ExpressionInput.js
```diff
@@ -6,7 +6,7 @@ import stripindent from 'strip-indent'
 import marshalParams from '@haiku/player/lib/reflection/marshalParams'
 import parseExpression from 'haiku-serialization/src/ast/parseExpression'
 import mod from 'haiku-serialization/src/bll/helpers/mod'
-import Palette from './DefaultPalette'
+import Palette from 'haiku-ui-common/lib/Palette'
 import AutoCompleter from './AutoCompleter'
 import * as EXPR_SIGNS from './helpers/ExprSigns'
```
stristr: Hi, noted that `./helpers/ExprSigns` was removed, but this import is still here. Does this still compile?

packages/haiku-ui-common/src/helpers/ensureRet.ts
```diff
@@ -0,0 +1,10 @@
+import eqToRet from './eqToRet';
+import * as EXPR_SIGNS from './ExprSigns';
+
+export default function ensureRet(str) {
+  let fixed = eqToRet(str);
+  if (fixed.slice(0, 6) !== EXPR_SIGNS.RET) {
+    fixed = (EXPR_SIGNS.RET + ' ') + fixed;
+  }
+  return fixed;
+}
```
stristr, line 10: (`const`-able like above)

packages/haiku-ui-common/src/helpers/ensureEq.ts
roperzh: Yet another way, just for fun ¯\_(ツ)_/¯ 

```js
export default function ensureEq(str) {
  const fixed = retToEq(str);
  return fixed.slice(0, 1) !== EXPR_SIGNS.EQ
    ? EXPR_SIGNS.EQ + fixed
    : fixed
}
```

packages/haiku-ui-common/src/Palette.ts
```diff
@@ -1,42 +1,41 @@
...
   PALE_GRAY: '#d1d1d1',
-  BLACK: '#0C0C0C',
+  ROCK: '#C0CCCD', // Making ROCK a bit colder
+  ROCK_MUTED: '#C0CCCD',
   SPECIAL_COAL: '#122022',
-  DARKEST_COAL: '#0F171A'
-}
+  STAGE_GRAY: '#ebebeb',
+  SUNSTONE: 'rgb(254,254,254)', // This is our "white", use sparingly
+};
```
stristr, line 41: Likely perf payoff to change things like these to `const enum` instead of an object. With our compiler options, `const enums` becomes strings during compile. We use these colors a lot.

packages/haiku-ui-common/src/react/icons/Bolt.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+import Palette from './../../Palette';
```
stristr, line 2: Nit: all `from './../` can be changed to `from '../`?

packages/haiku-ui-common/src/react/Menu.tsx
```diff
@@ -1,33 +1,33 @@
...
   wrapper: {
-    position: 'relative'
+    position: 'relative',
   },
   menu: {
     position: 'absolute',
     display: 'none',
-    background: Palette.SPECIAL_COAL,
-    color: Palette.PALE_GRAY,
+    background: Palette['SPECIAL_COAL'],
```
stristr: `Palette.SPECIAL_COAL` should be fine here and throughout.

packages/haiku-ui-common/src/react/icons/CheckmarkEmptyIconSVG.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
+import ButtonIconSVG from './ButtonIconSVG';
+
+export default ({color = '#636E71'}) => (
+  <ButtonIconSVG>
+    <g id="54" stroke={color} transform="translate(0.000000, -4.000000)" fill={color} />
```
stristr, line 6: Nit: pretty sure this `<g />` does nothing.

packages/haiku-ui-common/src/react/icons/CheckmarkIconSVG.tsx
```diff
@@ -1,10 +1,10 @@
-import React from 'react'
-import ButtonIconSVG from './ButtonIconSVG'
+import * as React from 'react';
```
stristr, line 1: Here and throughout: might be worthwhile to do an svgo pass if these are truly common UI components. The transform on the `<g />` is relatively inexpensive but can be cascaded down to the path for faster rendering.

packages/haiku-ui-common/src/helpers/doesValueImplyExpression.ts
stristr: (Same comment re: allowSyntheticDefaultImports for all cases of `import * as React from 'react';`…we should either disable that option or actually use these "synthetic" default imports when ergonomic.)

stristr:
General note: the idea of `haiku-ui-common` was for it to be a reasonable drop-in UI dependency in non-Haiku.app projects (e.g. adding the "reticulating splines" loader to the `share-page` project). In light of this, I saw a few things in here that probably make more sense for `haiku-sdk-creator`, including:

 * Globals.ts
 * Mixpanel.ts
 * most of the stuff in `src/helpers`

packages/haiku-ui-common/tslint.json
stristr: (`yarn fix` would do that for you now in this project.)

roperzh: But it won't format long lines into a shorter form, right?

packages/haiku-ui-common/src/react/Menu.tsx
```diff
@@ -180,35 +187,34 @@ class BaseSubMenu extends React.Component {
...
 }
 
 const BaseMenuItem = ({children, data, disabled, onClick, style}) => {
   return (
     <li
       className={CLOSE_IF_SELECTED_CLASS}
-      onClick={event => {
-        if (!disabled) onClick(event, data)
+      onClick={(event) => {
+        if (!disabled) { onClick(event, data); }
```
taylorpoe: Aren't semicolons in jsx bad practice?

taylorpoe (approved):

packages/haiku-ui-common/tslint.json
stristr: Yeah, I guess I don't think about that because my IDE does all that sort of stuff for me…I'm guessing VSCode does too with a certain configuration and could look into it if you'd like. It comes out to the same thing as prettier (I don't actually format code I type, a "system" does it for me), but without prettier's "opinions". Probably worthwhile to figure out how to get "house style" into whatever IDEs everyone is using.

Plan is to move basically all style nits into CI very soon, so I think that will help the most!

taylorpoe:
LGTM ...other than the semicolons.

packages/haiku-ui-common/tslint.json
roperzh: Thanks for the answer! makes sense :)

packages/haiku-ui-common/src/react/Menu.tsx
stristr: Haha, no, not really…there's work underway to align style throughout our projects to Airbnb's style guide, which is basically the most comprehensive React style guide there is.

It's an endless debate and nobody's "right", but I think Airbnb's [explanation](https://github.com/airbnb/javascript#semicolons) is pretty solid overall. Especially about "These rules will become more complicated as new features become a part of JavaScript."

Case in point for me is straight from [NPM's style guide](https://docs.npmjs.com/misc/coding-style#semicolons), which (IMO) creates a self-defeating argument by encouraging you to write `;[a, b, c].forEach(...)`

packages/haiku-ui-common/tslint.json
taylorpoe: > "house style"  into whatever IDEs everyone is using

🎉 I'd switch from Sublime to VScode if we could get this going.

tsortwehttam:
Made a series of fixes. HT @stristr for noticing the missing imports. (It was loading from previously compiled code, which is why I missed it.) I plan to merge this circa 4pm ET unless anybody raises a red flag before then.

roperzh:
Hello! are there any news on this? I wanted to check as I'm facing an scenario where would be useful to share components.

I'm able to help with any requested changes if needed

stristr:
I vote we fix merge conflicts and merge this.

PR #137, opened 2017-12-13
tweak: for now, untravis our build process

stristr:
As discussed on Slack: for now, we're going to move flaky builds to our East Coast data center.


stristr:
Merging this.

PR #138, opened 2017-12-14
Fix tour positioning

roperzh:
**About**

This is a medium-sized pr, should take around 10 minutes to review

Asana ticket: https://app.asana.com/0/479779791675933/498466015939025

**What**

- Tooltip.js (needs a better name) accepts a prop that instructs the component where to place the tooltip (`top`, `bottom`, `left`, `right`), this adds logic to detect if the tooltip is going to be off of the screen and auto-position the tooltip accordingly. Ideally we should iterate over this and accept an array of preferred positions instead of determining by guessing.
- Add logic to re-render the tour on scroll
- Add logic to auto-scroll to CheckTutorial if the project is below the fold

2017-12-14-12_27_45-compressor


stristr:
LGTM with _minor_ tweak in 76a0f3e. Merging!

PR #139, opened 2017-12-14
tweak: scale up sketch-instructions file

taylorpoe:
Should be a quick review. Just scaled up the instructions in our default sketchfile.

As @stristr mentioned last week, the instructions were too small — they had to be zoomed in to be read.

These assets (heart, diamond, logo) — which end up being users' play-around assets by default — are now larger in Haiku. The byproduct of enlarging them is that users will now be less likely to encounter the egregious blurring on scale-up. (It's quite a bad first-touch if the user scales one of these assets 5x up and then publishes and sees the raster-gunk in Safari.)


stristr:
LGTM, merging!

PR #140, opened 2017-12-14
hack: Disable timeline shift key behaviors behind a feature flag.

stristr:
This change is a fast hack to disable *all* shift-key timeline behaviors (yes, even shift-click to deselect). It's the fastest known working way to get out a build that doesn't have major multiselect bugs. Since the feature flag is easy enough to flip, I'm going to propose rolling a release with this now (i.e. checking it in) so @zackbrown and @taylorpoe can QA a build with this stuff disabled. It should be straightforward to kick off a new build if we find the fix. Meanwhile, I left shift key enabled on dev for attempted fixin'.

### DIFF
--- a/packages/haiku-common/config/experiments.json
+++ b/packages/haiku-common/config/experiments.json
@@ -1,5 +1,8 @@
 {
   "LottieExportInGlobalMenu": {
     "development": true
+  },
+  "TimelineShiftKeyBehaviors": {
+    "development": true
   }
 }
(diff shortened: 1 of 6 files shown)


tsortwehttam (changes requested):
Do we need to add a similar conditional here? https://github.com/HaikuTeam/mono/blob/700273bf44e24e2222a6ba5fb039b6c9740cc277/packages/haiku-timeline/src/components/Timeline.js#L810

stristr:
Shouldn't have to, because `Globals.isShiftKeyDown` is disabled at a higher level by the experiment flag. `Timeline.js` just had its own state thing with shift key going on.

tsortwehttam (approved):

tsortwehttam:
LGTM.

PR #141, opened 2017-12-15
Add logic to fix multiselect

roperzh:


tsortwehttam:
LGTM, merging. There are some small issues to clean up, but on the whole this is looking great.

PR #146, opened 2017-12-15
Separate the concepts of `selected` from `active`

roperzh:
Most of our problems come from the mix of those two used everywhere,
after this, we are using `selected` only to deal with curves.

It surely needs better naming, but this is a start.

Rules:

- If a keyframe has a curve, and it's selected, the curve is
highlighted
- If a keyframe has a curve and it's active, only the keyframe
is highlighted

I used `selected` because of legacy compatibility with drag&drop,
but we must change that


tsortwehttam:
LGTM. 👍

PR #147, opened 2017-12-15
Enable the "Next" button in all tutorial steps

roperzh:
As the title says, there are only two "special" steps:

- When the user is asked to open a project, here I included a bit
of extra logic to open the project programatically (hack)
- When the user is asked to publish the project, here I kept the
"Next" button hidden, because the step that follows depends on the
popup modal, if we think we absolutely need the "Next" button here
we can do it, but with a extra bit of logic (we need to talk to
envoy)

Asana ticket: https://app.asana.com/0/479779791675933/504610834323577


stristr:
LGTM!

PR #148, opened 2017-12-16
Fix layout issues with mana-ification of raster data

stristr:
This PR requires ~~a reviewer who deeply groks affine transformations and careful testing of Player demos~~ careful testing of object instantiation on stage.

 * [Asana ticket](https://app.asana.com/0/479779791675933/488080510412729/f)
 * [Asana ticket](https://app.asana.com/0/479779791675933/488686568445425/f)

**Background:**

After a pretty deeply false start (see below—tl;dr my understanding of the issue turned out to be deeply flawed), it turned out all our raster layouts were broken due to one small oversight in the the process of converting mana layout when instantiating an object on stage. Specifically, we blindly strip out the `<x, y>` node attributes during this process, which is a good idea to do on everything _except_ `<image>` nodes. The unique behavior of `<image>` nodes is that `<x, y>` _defines a coordinate system_ about which all transforms are applied.

~~The vendor library we use for decomposing a sequence of SVG transformations into `scale.x`-type properties uses [a specific ordering](https://github.com/mattdesl/mat4-decompose) for its decomposition. The ordering as it applies to work we do (since we don't support computations based on skew or perspective [yet]) is: 1. Translation, 2. Scale, 3. Rotation. This happens to be incompatible with the ordering we use to compute layout matrices, which is: 1. Rotation, 2. Scale, 3. Translation.~~

~~This doesn't really matter *unless* we have a case of where an origin-repositioning transform (scale or rotation) is composed with any other transform, at which point wacky things start to happen because origin-repositioning transforms are generally not commutative with other transforms. Put more simply, `translate(10, 20) rotate(30)` is not isomorphic to `rotate(30) translate(10, 20)`, and we parse complex transforms into the former but actually implement the latter.~~ Note from the future: these operations actually *do* commute as long as there is no transform-origin set.

**Implementation:**

~~To avoid backward-incompatible changes to how the Player calcs layout and any other side effects, this PR instead parses transform sequences into an order actually compatible with the rest of our code.~~ In the process, I refactored out all the dead/useless code dealing with skew and perspective, since it doesn't factor into our computed layout at this time.

~~Player demos should be tested, as well as the documented broken cases in Asana:~~
The following should be carefully tested:
 - Rotate raster slice in Sketch, then instantiate on stage. Confirm bounding box perfectly contains the rotated element.
 - Flip raster slice in Sketch, then instantiate on stage. Confirm orientation on Glass matches what you see in Sketch.

### DIFF
--- a/packages/haiku-player/src/Layout3D.ts
+++ b/packages/haiku-player/src/Layout3D.ts
@@ -145,7 +145,7 @@ function multiplyArrayOfMatrices(arrayOfMatrices: number[][]): number[] {
   return product;
 }
 
-function computeLayout(layoutSpec, currentMatrix, parentMatrix, parentsizeAbsoluteIn) {
+function computeLayout(layoutSpec, currentMatrix, parentsizeAbsoluteIn) {
   const parentsizeAbsolute = parentsizeAbsoluteIn || {x: 0, y: 0, z: 0};
 
   if (parentsizeAbsolute.z === undefined || parentsizeAbsolute.z === null) {
@@ -153,16 +153,11 @@ function computeLayout(layoutSpec, currentMatrix, parentMatrix, parentsizeAbsolu
   }
 
   const size = computeSize(layoutSpec, layoutSpec.sizeMode, parentsizeAbsolute);
-
-  const outputNodepad = {} as any;
-  const matrix = computeMatrix(outputNodepad, layoutSpec, currentMatrix, size, parentMatrix, parentsizeAbsolute);
-
-  outputNodepad.size = size;
-  outputNodepad.matrix = matrix;
-  outputNodepad.shown = layoutSpec.shown;
-  outputNodepad.opacity = layoutSpec.opacity;
-
-  return outputNodepad;
+  return {
+    ...layoutSpec,
+    size,
+    matrix: computeMatrix(layoutSpec, currentMatrix, size, parentsizeAbsolute),
+  };
 }
 
 export default {
--- a/packages/haiku-player/src/helpers/isMatrixableTransformArray.ts
+++ b/packages/haiku-player/src/helpers/isMatrixableTransformArray.ts
@@ -1,0 +1,14 @@
+import {transformValueIsEssentiallyInt} from './transformValueIsEssentiallyInt';
+
+export const isMatrixableTransformArray = (transform) => {
+  return transformValueIsEssentiallyInt(transform[2], 0) &&
+    transformValueIsEssentiallyInt(transform[3], 0) &&
+    transformValueIsEssentiallyInt(transform[6], 0) &&
+    transformValueIsEssentiallyInt(transform[7], 0) &&
+    transformValueIsEssentiallyInt(transform[8], 0) &&
+    transformValueIsEssentiallyInt(transform[9], 0) &&
+    transformValueIsEssentiallyInt(transform[10], 1) &&
+    transformValueIsEssentiallyInt(transform[11], 0) &&
+    transformValueIsEssentiallyInt(transform[14], 0) &&
+    transformValueIsEssentiallyInt(transform[15], 1);
+};
--- a/packages/haiku-player/src/layout/convertManaLayout.ts
+++ b/packages/haiku-player/src/layout/convertManaLayout.ts
@@ -158,9 +158,13 @@ export default function convertManaLayout(mana) {
       // Strip off the old value which is no longer needed
       delete attributes.transform;
 
-      // If the x/y attributes are present, they can interfere with the transform, so we strip them off
-      delete attributes.x;
-      delete attributes.y;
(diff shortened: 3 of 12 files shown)


packages/haiku-player/src/helpers/isMatrixableTransformArray.ts
```diff
@@ -0,0 +1,14 @@
...
+    transformValueIsEssentiallyInt(transform[3], 0) &&
+    transformValueIsEssentiallyInt(transform[6], 0) &&
+    transformValueIsEssentiallyInt(transform[7], 0) &&
+    transformValueIsEssentiallyInt(transform[8], 0) &&
+    transformValueIsEssentiallyInt(transform[9], 0) &&
+    transformValueIsEssentiallyInt(transform[10], 1) &&
+    transformValueIsEssentiallyInt(transform[11], 0) &&
+    transformValueIsEssentiallyInt(transform[14], 0) &&
+    transformValueIsEssentiallyInt(transform[15], 1);
+};
```
stristr, line 14: Note: this is unused as of right now. It's potentially throwaway work in service of a backlogged item to use `matrix()` instead of `matrix3d()` if possible.

packages/haiku-player/test/unit/layout.parseCssTransformString.test.js
```diff
@@ -38,14 +38,7 @@ test('layout.parseCssTransformString', function (t) {
...
       {
-        'translation.x': 0.2,
-        'translation.y': -0.1,
-        'translation.z': 14211.27,
-        'rotation.y': 1.57,
-        'rotation.z': 1.57,
-        'scale.x': 0,
-        'scale.y': 0,
-        'scale.z': 0
+        "scale.x": 0
```
stristr, line 84: Note: this test's outcome changed due to an optimization I added while I was deep in the thick of things (if any scale factor is 0, bail early with a "vanishing" layout).

stristr:
Ready for review (the approach changed a lot over the weekend).

packages/haiku-player/src/Layout3D.ts
```diff
@@ -145,24 +145,19 @@ function multiplyArrayOfMatrices(arrayOfMatrices: number[][]): number[] {
...
-  outputNodepad.matrix = matrix;
-  outputNodepad.shown = layoutSpec.shown;
-  outputNodepad.opacity = layoutSpec.opacity;
-
-  return outputNodepad;
+  return {
+    ...layoutSpec,
+    size,
+    matrix: computeMatrix(layoutSpec, currentMatrix, size, parentsizeAbsolute),
+  };
```
tsortwehttam, line 160: Future: Since we create so many of these objects, would we see a perf gain (prevent GC pauses) by using an object pool?

packages/haiku-player/src/layout/convertManaLayout.ts
```diff
@@ -158,9 +158,13 @@ export default function convertManaLayout(mana) {
...
-      // If the x/y attributes are present, they can interfere with the transform, so we strip them off
-      delete attributes.x;
-      delete attributes.y;
+      // If the x/y attributes are present, they can interfere with the transform, so we strip them off except in
+      // the case of <image>. <image> has the special behavior that x and y position the image relative to the
+      // origin *before* any transformations are applied.
+      if (node.elementName !== 'image') {
+        delete attributes.x;
+        delete attributes.y;
+      }
```
tsortwehttam, line 167: Great finding. 👍

packages/haiku-player/test/unit/layout.parseCssTransformString.test.js
tsortwehttam: It might be a good idea to add back a similar test that doesn't return the 'vanishing' layout, so we can continue to verify that the parsing works correctly for e.g. `matrix3d()` and such.

tsortwehttam:
LGTM.

packages/haiku-player/src/Layout3D.ts
stristr: Definitely think GC dumps are still a problem, but lower priority compared to making patch rendering do way less work (e.g. getting patches to be proper patches vs. rerendering an entire mana tree if the topmost `<svg>` node has an animation). From my profiling on haiku.ai, "major" GC dumps take about 2ms, so we're guaranteed to be looking at ≥ 14.667ms of rendering work when we actually get into a situation where we need to skip frames.

packages/haiku-player/test/unit/layout.parseCssTransformString.test.js
stristr: Added back proper unit tests for all the *3d functions (rotate3d, translate3d, scale3d, matrix3d). In the process, realized matrix3d was broken (wasn't doing unit conversion), so fixed that too!

PR #149, opened 2017-12-18
Zack/tracking fix

zackbrown:
accompanies:  https://github.com/HaikuTeam/inkstone/pull/10

purpose:  gets periodic update checking (and thus usage reporting) working

along the way:  introduces a functional example of using an envoy-backed, singleton BLL entity (User) as well as an example of a DAL entity (Registry.)

### DIFF
--- a/packages/haiku-cli/package.json
+++ b/packages/haiku-cli/package.json
@@ -31,7 +31,8 @@
     "cli-color": "^1.1.0",
     "dedent": "^0.6.0",
     "fs-extra": "^4.0.2",
-    
+    "@haiku/sdk-client": "HaikuTeam/sdk-client.git",
+    "@haiku/sdk-inkstone": "HaikuTeam/sdk-inkstone.git",
     "hasbin": "^1.2.3",
     "inquirer": "^2.0.0",
     "lodash": "^4.17.4",
--- a/packages/haiku-cli/package.json
+++ b/packages/haiku-cli/package.json
@@ -31,8 +31,7 @@
     "cli-color": "^1.1.0",
     "dedent": "^0.6.0",
     "fs-extra": "^4.0.2",
-    "@haiku/sdk-client": "HaikuTeam/sdk-client.git",
-    "@haiku/sdk-inkstone": "HaikuTeam/sdk-inkstone.git",
+    
     "hasbin": "^1.2.3",
     "inquirer": "^2.0.0",
     "lodash": "^4.17.4",
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -21,6 +21,7 @@ import Tour from './components/Tour/Tour'
 import AutoUpdater from './components/AutoUpdater'
 import EnvoyClient from 'haiku-sdk-creator/lib/envoy/EnvoyClient'
 import { EXPORTER_CHANNEL, ExporterFormat } from 'haiku-sdk-creator/lib/exporter'
+import { USER_CHANNEL, User } from 'haiku-sdk-creator/lib/bll/user'
 import { GLASS_CHANNEL } from 'haiku-sdk-creator/lib/glass'
 import { isPreviewMode } from '@haiku/player/lib/helpers/interactionModes'
 import Palette from './components/Palette.js'
@@ -301,6 +302,18 @@ export default class Creator extends React.Component {
         })
       })
     })
+
+    this.envoy.get(USER_CHANNEL).then(
+      /**
+       * @param {User} user
+       */
+      (user) => {
+        this.user = user
+        
+        //kick off initial report
+        this.onActivityReport(true)
+      }
+    )
 
     this.envoy.get('tour').then((tourChannel) => {
       this.tourChannel = tourChannel
@@ -628,12 +641,7 @@ export default class Creator extends React.Component {
 
   onActivityReport (userWasActive) {
     if (userWasActive) {
-      return this.props.websocket.request(
-        {method: 'checkInkstoneUpdates', params: [{}]},
-        (err) => {
-          console.log('[creator] ping to Inkstone for updates finished', err)
-        }
-      )
+      this.user.reportActivity()
     }
 
     this.setState({
--- a/packages/haiku-sdk-creator/src/bll/User.ts
+++ b/packages/haiku-sdk-creator/src/bll/User.ts
@@ -1,0 +1,40 @@
+import { client as sdkClient } from '@haiku/sdk-client';
+import { inkstone } from '@haiku/sdk-inkstone';
+import { MaybeAsync } from '../envoy';
+
+import { Registry } from '../dal/Registry';
+
+export interface User {
+  reportActivity: MaybeAsync<() => void>;
+  setConfig: MaybeAsync<(key: string, value: string) => void>;
+  getConfig: MaybeAsync<(key: string) => string>;
+  getAuthToken: MaybeAsync<() => string>;
+}
+
+export const USER_CHANNEL = 'user';
+
+export class UserHandler implements User {
+
+  reportActivity() {
+    const authToken = sdkClient.config.getAuthToken();
+    //faked query string to keep squirrel happy.  as this is not genuinely an 'update' check, this shouldn't matter
+    return inkstone.updates.check(authToken, '?platform=mac&environment=production&branch=master', () => { });
+  }
+
+  setConfig(key: string, value: string) {
+    Registry.setConfig(key, value);
+  }
+
+  getConfig(key): string {
+    return Registry.getConfig(key);
+  }
+
+  getAuthToken(): string {
+    return sdkClient.config.getAuthToken();
+  }
+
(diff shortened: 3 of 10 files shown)


packages/haiku-creator/src/react/Creator.js
```diff
@@ -302,6 +303,18 @@ export default class Creator extends React.Component {
       })
     })
 
+    this.envoy.get(USER_CHANNEL).then(
+      /**
+       * @param {User} user
+       */
```
zackbrown, line 309: TIL that you can annotate TypeScript types _inside JS_ and gain auto-complete.  See screenshot:  
image

Not as good as full-on TS, but should make working with BLL much more discoverable and usable on the front-end, without requiring a TS re-write.

packages/haiku-sdk-creator/src/bll/User.ts
```diff
@@ -0,0 +1,40 @@
...
+  reportActivity() {
+    const authToken = sdkClient.config.getAuthToken();
+    //faked query string to keep squirrel happy.  as this is not genuinely an 'update' check, this shouldn't matter
+    return inkstone.updates.check(authToken, '?platform=mac&environment=production&branch=master', () => { });
+  }
+
+  setConfig(key: string, value: string) {
+    Registry.setConfig(key, value);
+  }
+
```
zackbrown, line 27: For a while, @taylorpoe has been requesting an easy way to store/retrieve whether a user wants to see frames or ms by default in the timeline.  Now that can be done easily without leaving react code, using `this.user.setConfig('defaultTimelineUnits', 'ms')` — supports arbitrary key/values (this behavior is backed by the Registry, elsewhere in this PR.)

taylorpoe: NIce!

packages/haiku-cli/package.json
```diff
@@ -31,8 +31,7 @@
     "cli-color": "^1.1.0",
     "dedent": "^0.6.0",
     "fs-extra": "^4.0.2",
-    "@haiku/sdk-client": "HaikuTeam/sdk-client.git",
-    "@haiku/sdk-inkstone": "HaikuTeam/sdk-inkstone.git",
+
```
stristr: (I think this was a cut that should have been a copy.)

packages/haiku-sdk-creator/src/dal/Registry.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!fs.existsSync(REGISTRY_PATH)) {
+      fs.writeFileSync(REGISTRY_PATH, '{}');
+    }
+    return JSON.parse(fs.readFileSync(REGISTRY_PATH).toString());
+  }
+
+  static _setFileContents(contents: any) {
+    Registry._ensureHomeFolder();
+    return fs.writeFileSync(REGISTRY_PATH, JSON.stringify(contents));
+  }
```
stristr: Nit: preferred way to make "private static" methods would be making functions and not exporting them. `const ensureHomeFolder = () => { ... };`

packages/haiku-sdk-creator/yarn.lock
```diff
@@ -4546,6 +4562,33 @@ request@2, request@^2.74.0, request@^2.81.0:
     tunnel-agent "^0.6.0"
     uuid "^3.1.0"
 
+request@^2.79.0:
+  version "2.83.0"
```
stristr, line 4566: Warning, this is the request version that started having SSL issues on share-page. I think it was a side effect of how it interacts with webpack, but you might want to check to make sure.

packages/haiku-sdk-creator/src/dal/Registry.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as mkdirp from 'mkdirp';
+import { client as sdkClient } from '@haiku/sdk-client';
+
+const REGISTRY_PATH = path.join(os.homedir(), '.haiku', 'registry.json');
+
+// Purpose:  Set and retrieve locally persisted settings.
+//           Intended to supersede (and deprecate) haiku-serialization/*/HaikuHomeDir and most of haiku-sdk-client
+// Use-case: Easily store and retrieve whether the user chooses to view their timeline in ms or frames —
+//           and specifically, remove friction around UI devs wanting to persist evolving data
+export class Registry {
```
stristr, line 30: Drive-by: `localStorage` is great for storing configs which are frequently accessed, and (assuming no difficult persistence problems—we'd probably want to flush it to disk when the app is closing gracefully and read it from disk when the app boots up) feels more robust than filesystem.

Plus, its explicit, object-oriented API makes it way easier to write unit tests for (`Registry.setStore(mockLocalStorage);` vs. creating a mock filesystem).

packages/haiku-cli/package.json
zackbrown: ah, good catch.

packages/haiku-sdk-creator/src/dal/Registry.ts
zackbrown: good call; I suppose as written they aren't quite private

packages/haiku-sdk-creator/src/dal/Registry.ts
zackbrown: That's a good idea — if/when this needs to be developed further, we could explore a node implementation of localStorage as the backing here.  Like this perhaps:  https://www.npmjs.com/package/node-localstorage .  

Though to be clear, since config needs to transcend webviews, browser-side localStorage isn't an option here.

packages/haiku-sdk-creator/src/dal/Registry.ts
```diff
@@ -16,31 +33,14 @@ export class Registry {
...
-  static _setFileContents(contents: any) {
-    Registry._ensureHomeFolder();
-    return fs.writeFileSync(REGISTRY_PATH, JSON.stringify(contents));
-  }
-
   static setConfig(key: string, value: string) {
-    const config = Registry._getFileContents();
+    const config = _getFileContents();
     config[key] = value;
     Registry._setFileContents(config);
```
stristr: Now just `_setFileContents(config);`

zackbrown: 👍  fixed

stristr:
LGTM, although you didn't ask me :smile:

stristr:
(leaving open in case you'd like @matthewtoast to look before merging)

packages/haiku-sdk-creator/src/dal/Registry.ts
stristr: Ah, wasn't even thinking about non-browser contexts. Okay then.

tsortwehttam:
LGTM.

Only one nit: I think we should start following the haiku-tslint rules to a T (and stop adding exceptions). I am guilty of this many times but intend to stop as my New Year's resolution.

packages/haiku-sdk-creator/src/dal/Registry.ts
stristr: On second glance, I realized you might have meant "be shared between webviews" instead of "work in a non-browser context"—can you clarify? Because localStorage is already shared between Creator, Glass, and Timeline. (It's like Electron IPC except fast :smile:.)

zackbrown: Well, both — yes, needs to be shared between webviews, and yes, needs to work in a non-browser context (i.e. DAL needs to be node.)  

Or are we not agreed on the three-tier UI/BLL/DAL approach?  Direct UI <-> localStorage persistence would short-circuit that design and lose its benefits. (e.g. since it'd be strongly coupled to UI logic, would be painful to extricate in the future; also, would isolate config read/write to the UI layer, whereas we should be able to read/write config anywhere in the app, e.g. especially other BLL entities.)  Happy to talk more about this.

stristr: Gotcha—no, we're agreed on three-tier architecture, just wasn't remembering "DAL needs to be node" (ultimately because there is no client-side WebSocket server). Will continue to think about this is a current and not an absolute constraint…"config read synchronously from disk every time it's accessed" does feel like a ticking time bomb of a sort. Not putting up any roadblocks here, though.

zackbrown: extra ticking time-bomb in that this logic doesn't handle exceptions gracefully.  eyes on this if we start seeing total app crashes (once we're actually using Registry, that is—currently nothing is actually using it.)

packages/haiku-sdk-creator/yarn.lock
zackbrown: I didn't intentionally change this — though i did have to run a `yarn install` at one point while tangling with tooling

stristr: We'll see the error in the forthcoming build if it exists. I'm 99% sure the whole request lib version is a red herring and the real problem was a new version of webpack which commandeers HTTP requests for unknown reasons.

PR #151, opened 2017-12-20
save sketch starter file from Sketch42

taylorpoe:
Resaved our Sketch starter file from a copy of Sketch Version 42.


stristr:
LGTM!

PR #152, opened 2017-12-20
feat: Add a new player config option for strictSizing.

stristr:
Should be really quick review for @matthewtoast 😄 

Background: the last Player perf pass introduced the concept of caching the internal representation of the mount's bounding box between full flush renders if there isn't a sizing mode enabled, i.e. if we don't actually care about the size of the mount's bounding box for our rendering logic. This resulted in a huge performance improvement for patch rendering when there is no sizing mode, but did nothing for cases where there is. (Calculating a bounding box on every RAF is pretty expensive.)

This PR introduces a default-on config-based option (`strictSizing`) which, if disabled, tells the DOM renderer to recalculate the bounding box only on `window.resize`. It feels safer to go with defaults tracking current behavior and experiment with this selectively, but plan would be to disable `strictSizing` on Dashboard 2, share page, and haiku.ai, all of which use sizing modes to achieve the desired effect.

Note: since component-initialized window event listeners are the domain of `HaikuDOMRenderer`, it  felt wisest to refactor the DOM renderer to accept config.

### DIFF
--- a/packages/haiku-creator/src/react/components/ProjectPreview.js
+++ b/packages/haiku-creator/src/react/components/ProjectPreview.js
@@ -39,6 +39,7 @@ class ProjectPreview extends React.Component {
           this.mount,
           {
             sizing: 'cover',
+            strictSizing: false,
             loop: true,
             interactionMode: InteractionMode.EDIT,
             autoplay: false
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -3,6 +3,7 @@ import lodash from 'lodash'
 import Radium from 'radium'
 import HaikuDOMRenderer from '@haiku/player/lib/renderers/dom'
 import HaikuContext from '@haiku/player/lib/HaikuContext'
+import Config from '@haiku/player/lib/Config'
 import ActiveComponent from 'haiku-serialization/src/bll/ActiveComponent'
 import Element from 'haiku-serialization/src/bll/Element'
 import react2haiku from 'haiku-serialization/src/utils/react2haiku'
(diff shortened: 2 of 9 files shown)


stristr:
Note: please do NOT merge so we can cut a clean build from master tonight with the fix from #151 in. I would like this to push a semver release into this branch after we get tomorrow's release out so we can rebuild @taylorpoe's blog post project using 2.3.50. The perf gains are palpable (main motivation for doing this PR now vs. later).

tsortwehttam:
Code LGTM. Other thoughts:

- Is `'resize'` the only event we'd want here? I'm wondering if we might also want to listen for `'orientationchange'` -- and possibly other events that, while not strictly being a "resize," do change the view dimensions.

- Should we consider calling this option `resizeWithWindow` (or something like that) that makes it more explicit what the effect will be? `strictSizing` is fine but it leaves one guessing about the semantics. No strong opinion here; just mentioning it.

- Vaguely wondering if we might want to tie this logic to _element_ resizing as opposed to window resizing. Or would that defeat the purpose? When it comes to layouts and responsive design, one tends to care about the container size as opposed to the page size. We could listen to the mount element with something like this https://github.com/sdecima/javascript-detect-element-resize to vendor in. Just a thought.

stristr:
1. Good call on `orientationchange`. I will add.
2. Yes, I struggled with what to call this too and agree we should rename, but was thinking the API should be `true` if we're always checking the container for its dimensions, and `mountSizeMightChangeDynamically` seemed like a bad name for an option…since this is part of the config options API, though, I think we should really like the name we give it. Any ideas about how to frame it more succinctly?
3. Yes, exactly, then we wouldn't need a config at all and could make this perform optimally in all cases—but I was waiting for [ResizeListener](https://developers.google.com/web/updates/2016/10/resizeobserver) to have [better support](https://caniuse.com/#search=ResizeObserver) vs. hacking it through a polyfill. I've looked into the implementation details of that and similar libraries, and they are precariously balanced on hacks and hopes…so that was thinking behind having an "element might be resized by something other than traditional responsive layout" flag.

stristr:
1. Added `orientationchange` listener as wisely noted.
2. Came up with a less awful name, `alwaysComputeSizing`, for what this option does.
3. Upped the ante with our rendering tests, so I could actually write tests for these beasts! It was pretty tricky figuring out how to trick jsdom into giving our objects layout…ch-ch-ch-ch-check out 855d65f if you're curious what the trick was!

stristr:
Gonna rebase and merge this now.

PR #153, opened 2017-12-20
List touch events as available event listeners

roperzh:
The player already handles all the logic to attach event handlers,
this only lists touch events as available, making them appear on
the events ui.

This should be a straightforward review

[Asana ticket](https://app.asana.com/0/482906470227387/505957877019449)


tsortwehttam:
LGTM.

PR #155, opened 2017-12-20
Add simple window dragging for the tour

roperzh:
As the title says, it is using the same react library as the
timeline for the dragging (`react-draggable`) for consistency
purposes, but we may consider other, more robust options.

Sasha suggested `react-dnd` which has an awesome API, maybe we
can go with that as our standard when doing the refactor of the
timeline multiselect.

[Asana task](https://app.asana.com/0/482906470227387/504284557602347)


stristr (commented):
LGTM! Merging.

PR #157, opened 2017-12-23
Use frames by default in timeline API.

stristr:
WIP, please don't merge! Don't even look!

Pending:
 - After `mc` is merged, teach plumbing to upgradeBytecodeInPlace and persist the result to disk on project start. (Note: at this point we can remove all the places we call it manually in the app.)
 - Update first semver this is supported in from 2.3.62 to whatever the actual version is right before merge, and probably make a compatibility config to store this sort of thing in.


stristr:
I'm closing this, will reopen a new one.

PR #160, opened 2017-12-27
Support two-finger trackpad scroll to pan stage; also support cmd+0 to reset stage pan & zoom

zackbrown:


packages/haiku-glass/src/react/Glass.js
```diff
@@ -382,6 +382,33 @@ export class Glass extends React.Component {
       false
     )
 
+    document.addEventListener('mousewheel', (evt) => {
+      //on mac, this is triggered by a two-finger pan
+      if (!this.getActiveComponent()) {
+        return
+      }
+
+      var artboard = this.getActiveComponent().getArtboard()
```
roperzh: nit: use `let` instead of `var`

packages/haiku-glass/src/react/Glass.js
```diff
@@ -382,6 +382,33 @@ export class Glass extends React.Component {
...
+      const deltaX = evt.wheelDeltaX,
+            deltaY = evt.wheelDeltaY
+
+      //HACK:  If you zoom and pan in quick succession, getZoom() will sometimes return 0 (or specifically, will return some value Y such that `1 / Y == Infinity`.)
+      //       [probably a FS-race related bug; should be fixed by decoupling FS read/write from an in-mem representation]
+      //       as a result, the logic that compensates for pan 'sensitivity' based on zoom goes haywire.
+      //
+      //       This null-coalesce (`|| SCROLL_PAN_COEFFICIENT`) works around this bug.
+      //       Ideally, Artboard#getZoom() should return a reliable value.
+      var scale = SCROLL_PAN_COEFFICIENT / ((artboard.getZoom() ^ 2) || SCROLL_PAN_COEFFICIENT)
```
roperzh: nit: use `let` instead of `var`

packages/haiku-glass/src/react/Glass.js
```diff
@@ -382,6 +382,33 @@ export class Glass extends React.Component {
       false
     )
 
+    document.addEventListener('mousewheel', (evt) => {
+      //on mac, this is triggered by a two-finger pan
+      if (!this.getActiveComponent()) {
```
roperzh: Maybe worth caching this function call into a variable since it's called again later on?

```
const activeComponent = this.getActiveComponent()
```

roperzh (commented):
LGTM code wide, I have tested the `cmd+0` functionality, but I don't have a trackpad to test the rest of the functionality.

Please note that the code needs an styling pass (`yarn lint-all`)

zackbrown:
pushed a lint fix — also, please _do_ test scroll wheel + shift+scroll wheel for stage panning, if you have that hardware instead of the trackpad.  It should work the same way, but I haven't been able to test as I'm not traveling with a mouse.

zackbrown:
(specifically:  should act about the same as Sketch for zooming + panning with trackpad AND mouse scrollwheel)

packages/haiku-ui-common/src/Globals.ts
```diff
@@ -35,9 +35,7 @@ window.addEventListener('keyup', (keyupEvent) => {
 });
 
 window.addEventListener('keydown', (keydownEvent) => {
-  if (keydownEvent.which === 16 && experimentIsEnabled(Experiment.TimelineShiftKeyBehaviors)) {
```
stristr, line 38: Assuming the intent is to retire the experiment, remaining steps are:
 - remove the similar block in `haiku-timeline/src/components/Timeline.js`
 - (remove unused imports here and in that file)
 - remove the experiment config and enum (see `haiku-common/config/experiments.json` and `haiku-common/src/experiments/index.ts`)

zackbrown: The intent was a logic-fix:  this experiment flag broke global Shift.  

Due to that brokenness, I set out to retire the experiment thoroughly, then discovered that this logic check seemed to be redundant with the other place this flag is checked (https://github.com/HaikuTeam/mono/blob/master/packages/haiku-timeline/src/components/Timeline.js#L394 — note it also checks whether Shift is down.)  So removing this line appeared to be a minimal and viable fix.

stristr: Sadly, it was not redundant…the Timeline stuff uses a combination of `Globals.isShiftKeyDown` and `this.state.isShiftKeyDown` all over, so it did actually need to be in both places. I think previously `Globals.js` was in `haiku-timeline` and then got ported over in `mc`. I think it would be best to just retire the experiment for good at this point.

zackbrown: mm, that's unfortunate.  pushed 5431fc8, which retires the experiment.

stristr: 👍 thanks!

PR #161, opened 2017-12-27
The exorcism

stristr:
a.k.a. migration to [Yarn workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/) and related cleanup. If you are making changes to any build code or package dependencies, please review this first to make ensure your changes are compatible.

*Note: this PR does not actually contain ~2K commits. This is an artifact of a methodology change with our subtree pulls during semver bumps—and ironically will result in fewer autocommits after this is merged. Meanwhile, this branch will have to be merged manually and can't be rebased.*

This PR includes an effort to align dependency versions across all code that touches `Haiku.app`, including mono itself, so that `packages/haiku-*/node_modules` and `packages/@haiku/*/node_modules` are always empty, and one `yarn install --production` provides tight, correct, and consistent `node_modules` in mono root in a single place. I made the unconventional decision to develop `mono` (not `mono/packages`) as the workspace root early on, with all the stuff we use for building shunted into `devDependencies`, with the basic goal of ensuring no surprises related to different versions of packages running around. We still need to hoist dependencies from the root into `haiku-plumbing` when we build for release due to ASAR's discomfort with symlinks in general but we are virtually guaranteed singleton dependencies with this project structure.

**Major workflow changes:**
 - All open source modules are now in `packages/@haiku`.
 - There is no more `yarn [yarn-link|yarn-unlink|relink|sync]`, only `yarn install`.
 - Introduction of `yarn doctor` with the basic goal of telling you why your build isn't working.
 - Any hardcoded references to `/path/to/local/node_modules/haiku-timeline/index.html`-type code should be replaced with `require.resolve('haiku-timeline/index.html')`-type code. We shouldn't have to think about where things are, but they will almost always be in `<rootDir>/node_modules` during development and `<rootDir>/packages/haiku-plumbing/node_modules` in a release.
 - Use latest yarn (1.3.2) and latest LTS Node (8.9.3).
 - Move `haiku-fs-extra` local so we don't have any Git-bound dependencies remaining in the project.
 - Stop doing `--squash` subtree work.

Current state:

- [x] development working (`yarn start`, `yarn go`, `yarn compile-all`, `yarn lint-all`, `yarn test-all`, etc.)
- [x] semver and npm publish working (`yarn semver`, `yarn push`)
- [x] build working locally
- [x] build working on Jenkins
- [x] update mono docs

Tabled nice-to-haves:

- [ ] migrate Player build to use `rollup.js`
- [ ] migrate all development and dev builds to use webpack with single watcher configs and tree-shook vendor build


stristr:
Ready for review 😄. I'm going to recommend doing a fresh clone (`cd /tmp && git clone git@github.com:HaikuTeam/mono.git -b the-exorcism && cd the-exorcism && yarn install && yarn setup && yarn go`), doing some quick sanity checks for any new breakage, and merging as soon as possible if that all checks out. As noted in PR description, this is un-rebaseable due to the new approach to semver bumps (no-squash/preserve history), so the sooner we merge this, the better.

Any trailing PRs started after mc was merged and before this may need some extra TLC due to the exhaustive alignment of deps across projects and tweaks required to get everything in working order.

tsortwehttam:
LGTM. 🔥 stoked to finally have a more sane setup for the packages.

Note: I only reviewed as far as cloning and starting up development with the main flavors, and then verifying that a change to a symlinked module would reflect correctly after a refresh.

If necessary, we should let folks know that a fresh clone might be the preferred thing to do if developing after this change.

package.json
```diff
@@ -1,155 +1,130 @@
...
+      "!**/{.DS_Store,.git,.hg,.svn,CVS,RCS,SCCS,__pycache__,thumbs.db,.gitignore,.gitattributes,.editorconfig,.flowconfig,.yarn-metadata.json,.idea,appveyor.yml,.travis.yml,circle.yml,npm-debug.log,yarn-debug.log,.nyc_output,yarn.lock,.yarn-integrity,tslint*.json,tsconfig.json}",
+      "!**/@haiku/player/demo",
+      "!**/node_modules/aws-sdk",
+      "!**/node_modules/eslint*",
+      "!**/node_modules/monaco-editor/dev",
+      "!**/node_modules/mout",
+      "!**/node_modules/node-pre-gyp",
+      "!**/node_modules/rocambole-indent",
+      "!**/node_modules/sinon",
+      "!**/node_modules/table",
```
stristr, line 106: @matthewtoast, I'd like to take a careful look at L100-107. Specifically, which of these are only here because something that could be in `devDependencies` isn't? I have my eye on `aws-sdk` in particular (ignoring cases where we selectively remove a chunk of something like `monaco-editor`).

The actual solution is a proper tree shaking build, but there could be a more straightforward for-now solution….

tsortwehttam: I can't recall offhand anymore where these come from.

I think it might be a dep-of-a-dep-has-dev-deps-in-its-deps issue. I looked at this semi-closely the last time I was futzing with the distro process, but I've forgotten.

FWIW, possibly tangential: The module that is part of our build that seems to add the most egregious amount of dep-cruft seems to be `npm`. We originally used that to `npm install` in a user's project folder but that might not be needed any more since we changed the process to just copy the source code over.

PR #162, opened 2017-12-28
Ensure 'content' property gets written correctly to tspan elements

tsortwehttam:
OK to merge.

**What:** Add necessary whitelist entries so that `<tspan>` text content gets written correctly as content when its text content has been hoisted to timeline property as part of the serialization pipeline. This bug was a side-effect of a change made around multi-component/JIT properties to allow text content to be controlled via the timeline by the user.

**Review:** Should be short. Note this includes changes to `haiku-player` which we will need to figure out how to merge into #161.


stristr:
LGTM, merging.

stristr:
(Merges to #161 without issue. Shouldn't really have problems except for updates to deleted files like `yarn.lock`s.)

PR #163, opened 2017-12-29
Enable JIT properties; consolidate menus; fix issues

tsortwehttam:
Ready to merge -- but there may still be some random outstanding issues to fix.

**What:**

- Enable the JIT property menus with some basic addable properties:
  - CSS style properties
  - SVG attributes
  - Placeholder
  - I _excluded_ size mode, align, origin, mount, etc. because we need to discuss those
- Refactored (most of) the context menus to use a shared harness and a native menu
  - _Later_, I will circle back and replace with a React menu that matches our aesthetic consistently
- Instead of bolt icon => Action Editor, we now have gear icon => full element menu
- Fix the issue where clicking said icon ^^ didn't work
- Let player accept either a number (child index) or string (key mapping) for `controlFlow.placeholder`, which is a little step toward a usable placeholder UI
- Minor bug fixes and touch-ups as I worked

**To the reviewer:**

- Should be a medium review.


stristr:
LGTM, merging.

PR #164, opened 2017-12-30
fix: always write out dom-embed.js.

stristr:
Teensy review. Noticed this significant regression when testing embed code—we need to always write out `dom-embed.js` in the project folder for the embed code to be checking for the right version of `@haiku/player`.

[Asana ticket](https://app.asana.com/0/480796620059175/511466430513270/f)


tsortwehttam:
LGTM, merging.

PR #165, opened 2017-12-30
Fix legacy of short-lived strictSizing API

stristr:
Tiny fix: after renaming the `strictSizing` API to `alwaysComputeSizing`, some stuff got lost in the shuffle. This should restore expected behavior.


tsortwehttam:
LGTM, merging.

PR #166, opened 2017-12-30
fix: un-bork new project init

stristr:
Noticed that we're not always going to have props to read children from, and this in particular seems to break new project init.


stristr:
Actually n/m, this is fixed and I was using old code.

PR #168, opened 2018-01-02
Go back to original playhead when existing preview

roperzh:
**Info**

- This PR should only take a couple of minutes to review
- [Asana Task](https://app.asana.com/0/482906470227387/505957877019447)

**Summary**

Since the internal playhead state of the `Timeline` model doesn't change when switching to preview mode, we can reset the playhead using `Timeline#getCurrentFrame` and `Timeline#getCurrentMs`


tsortwehttam:
LGTM, merging.

PR #169, opened 2018-01-02
Primitives Instantiation (incomplete but worth merging)

tsortwehttam:
OK to merge.

**What:**

- Adds (experiment-hidden) logic for instantiating primitive designs as primitive components
- Various fixes to supporting, tangential, and ancillary logic for this feature

**To the reviewer:**

- Should be a medium review.
- The primitive instantiation behavior is working "in essence," but it is both _unfinished_ and _hidden behind an experiment flag_. (But don't expect it to work right if you enable the experiment flag.) There were some tricky problems surfaced through this work that need to be resolved before I can finish the feature, but...
- I want to merge these changes now, since they will be required in the near future, and will end up fixing some edge cases to the feature and some behavior that we will depend on for multi-component as well.

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -614,20 +614,21 @@ class ActiveComponent extends BaseModel {
    * @param cb {Function}
    */
   instantiateReference (subcomponent, identifier, modpath, coords, overrides, metadata, cb) {
-    const fullpath = (modpath[0] === '.')
-      ? path.join(this.project.getFolder(), modpath) // Expected to be ./*
-      : modpath // Expected to be @haiku/*
+    let fullpath
 
-    // const relpath = (modpath[0] === '.')
-    //   ? path.relative(this.getSceneCodeFolder(), fullpath)
-    //   : modpath
+    const isExternalModule = modpath[0] !== '.'
 
-    // const fulldir = path.dirname(fullpath)
+    if (!isExternalModule) {
+      fullpath = path.join(this.project.getFolder(), modpath) // Expected to be ./*
+    } else {
+      fullpath = modpath
+    }
 
     // This assumes that the file has already been written to the file system or
     // stored inside the module require.cache via an earlier hook
     const mod = ModuleWrapper.upsert({
       uid: fullpath,
+      isExternalModule,
       file: this.project.upsertFile({
         relpath: modpath,
         folder: this.project.getFolder()
@@ -664,13 +665,13 @@ class ActiveComponent extends BaseModel {
    * @description Given an identifier and the bytecode of some primitive, instantiate
    * it as a reference to that primitive instead of the whole primitive bytecode
    */
-  instantiatePrimitive (identifier, primitive, coords, metadata, cb) {
+  instantiatePrimitive (primitive, coords, overrides, metadata, cb) {
     return this.instantiateReference(
       null,
-      identifier,
-      primitive.metadata.relpath,
+      primitive.getClassName(),
+      primitive.getRequirePath(),
       coords,
-      {},
+      overrides,
       metadata,
       cb
     )
@@ -723,6 +724,8 @@ class ActiveComponent extends BaseModel {
         mana,
         coords
       )
+
+      Bytecode.applyOverrides(overrides, timelinesObject, timelineName, `haiku:${componentId}`, timelineTime)
 
       template.children.push(mana)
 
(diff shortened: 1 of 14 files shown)


packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -957,12 +968,53 @@ class ActiveComponent extends BaseModel {
...
+        const haikuId = node.attributes[HAIKU_ID_ATTRIBUTE]
+
+        const timelineObj = (
+          bytecode.timelines &&
+          bytecode.timelines[timelineName] &&
+          bytecode.timelines[timelineName][`haiku:${haikuId}`]
+        )
+
+        if (timelineObj) {
+          for (const propertyName in timelineObj) {
```
roperzh, line 990: _safety check_: Is there any chance of iterating over unwanted inherited properties? If yes consider using `for (const propertyName of Object.entries(timelineObj))`

packages/haiku-serialization/src/bll/Bytecode.js
```diff
@@ -304,6 +304,33 @@ Bytecode.pasteBytecode = (destination, pasted, translation) => {
...
 
+Bytecode.applyOverrides = (overrides, timelinesObject, timelineName, timelineSelector, timelineTime) => {
+  if (overrides && Object.keys(overrides).length > 0) {
+    if (!timelinesObject[timelineName]) timelinesObject[timelineName] = {}
+    if (!timelinesObject[timelineName][timelineSelector]) timelinesObject[timelineName][timelineSelector] = {}
+
+    for (const propertyName in overrides) {
+      if (!timelinesObject[timelineName][timelineSelector][propertyName]) timelinesObject[timelineName][timelineSelector][propertyName] = {}
+      if (!timelinesObject[timelineName][timelineSelector][propertyName][timelineTime]) timelinesObject[timelineName][timelineSelector][propertyName][timelineTime] = {}
+      if (!timelinesObject[timelineName][timelineSelector][propertyName][timelineTime].edited) {
```
roperzh, line 315: I wonder if we can make this all this safety logic more clear and less verbose with `Object.assign`

packages/haiku-serialization/test/bll/02_ActiveComponent.test.js
```diff
@@ -331,3 +333,47 @@ const PATH_SVG_2 = `
       </g>
   </svg>
 `
+
+const PATH_SVG_3 = `
```
roperzh, line 337: _extreme nit_: this and subsequent paths can use an `svgo` pass.

packages/haiku-timeline/src/components/ExpressionInput.js
```diff
@@ -216,7 +228,13 @@ export default class ExpressionInput extends React.Component {
       // Assume that we already stored warnings about this function in the evaluator state from a change action
       return false
     } else {
-      let observedType = typeof committable
+      let observedType
+      if (Array.isArray(committable)) {
+        observedType = 'array'
+      } else {
+        observedType = typeof committable
```
roperzh, line 235: _safety check_: maybe it doesn't matter in this context, but a heads up that we can end with misleading values for `observedType` (for example `typeof NaN === "number"` is true)

roperzh (approved):
The code LGTM, I also have tested the app with the feature disabled and nothing seems to be out of place, I was able to do basic animations, etc.

I haven't tested with the flag enabled as this is a work in progress.

packages/haiku-serialization/src/bll/ActiveComponent.js
tsortwehttam: Good catch; in practice this sort of logic is used widely so I'm in favor of circling back later to write a utility and then refactoring to use that.

packages/haiku-serialization/src/bll/Bytecode.js
tsortwehttam: Good point. This is another case where a utility would be ideal (as opposed to Object.assign (which isn't deep and which doesn't respect the `edited` property which is necessary for detecting appropriate properties to merge during a design merge)).

packages/haiku-serialization/test/bll/02_ActiveComponent.test.js
tsortwehttam: In this case it's better _not_ to do the `svgo` pass on the test fixtures, since that allows us to check that the svgo pass works correctly.

packages/haiku-timeline/src/components/ExpressionInput.js
tsortwehttam: In this case a safe `isNumeric` check already occurs upstream, so this value should never be `NaN`.

packages/haiku-serialization/test/bll/02_ActiveComponent.test.js
roperzh: woot! I am dumb, sorry about that :)

packages/haiku-timeline/src/components/ExpressionInput.js
roperzh: 👍 sgtm, just to be clear there are other weird scenarios like `typeof null === "object"`, `NaN` was one example.

PR #170, opened 2018-01-03
Ui Tweaks

taylorpoe:
**Has been reviewed/approved:**
* Updates various small icon, coloring, & layout details
* Updates css transition group package from the old deprecated package to the suggested one [instructions on the package we were using](https://www.npmjs.com/package/react-addons-css-transition-group)
* Disallow project creation if name is blank

**Still needs review:**
* Only show the actions bolt icon next to collapsed element rows names if it _has_ actions attached. 
* Fixes actions icons to change to blue immedately when elements are given actions
* put ticker line above collapsed rows
* Fixes click area of frame events to not respond to gauge numbers being clicked
* Brings ticker bulb up on top of numbers to reduce overall gauge height — had to put the numbers back in the bulb because the bulb now covers the numbers.


stristr:
Deps management LGTM, have not reviewed code.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -240,6 +245,7 @@ class ProjectBrowser extends React.Component {
 
   handleNewProjectGo () {
     const raw = this.newProjectInput.value
+    if (!raw) return false
```
tsortwehttam: What if the string is `"    "` or `"\n   "`?

packages/haiku-creator/public/stylesheets/global.css
```diff
@@ -264,8 +264,8 @@ input::-webkit-input-placeholder {
 }
 
 .toast-enter {
-  opacity: .1;
-  transform: translateY(-400px);
+  opacity: .01;
```
roperzh: _extreme nit_: `0.01` reads much more clearly for me, but whatever we choose should be consistent across the file.

packages/haiku-creator/src/react/components/ProjectBrowser.js
roperzh: _nit_: would prefer a more explicit name like `rawProjectName` or similar

roperzh (changes requested):
LGTM, the only necessary change to make is a lint pass (`yarn lint-all`), as there are unused variables and other lint issues

packages/haiku-creator/src/react/components/ProjectBrowser.js
tsortwehttam: That bad variable name is my doing and I agree it should be changed.

taylorpoe: Spaces and slashes are both ignored as the user is typing; so they won't ever end up here.

taylorpoe: updated

packages/haiku-creator/public/stylesheets/global.css
taylorpoe: updated

roperzh (approved):

roperzh:
:shipit:

taylorpoe:
Oops, I incorrectly thought this was already merged when I saw that the PR was approved. I have a few new commits here that will need to be reviewed/approved. I will outline the full scope in the PR message.

tsortwehttam:
LGTM.

roperzh:
LGTM.

PR #171, opened 2018-01-03
Enable stage panning in preview mode

roperzh:
**Info**

- Small review
- [Asana ticket](https://app.asana.com/0/482906470227387/488080510412732)

**About**

As the title says, this PR allows panning on preview mode, while keeping all other mouse related interactions disabled, it also adds a bit of logic to disable all context menus on glass during preview mode.

2018-01-03-11_09_39-compressor


tsortwehttam:
LGTM.

Nit: Remember to prefix commit messages with `feat:`, `fix:`, etc.

roperzh:
@matthewtoast thanks! I will keep that in mind, do we have rules or it's more a common sense thing?

PR #172, opened 2018-01-03
Lottie fixes

stristr:
Some very minor Lottie fixes I noticed while working on different issues. Just a few minutes to review.

1. lottie-web 5.x doesn't like stroke-dasharray represented as an empty array (`[]`). This changes behavior to only put in a dash array if a nontrivial one exists.
2. Something in `mc` made the virtual timeline property `__max` show up during iteration, which essentially breaks all export. I worked around this by making all __-prefixed timeline IDs get skipped, in case we want to add more "concrete virtual properties" later.


tsortwehttam:
LGTM.

PR #173, opened 2018-01-04
Finishing touches on CI

stristr:
Should be a pretty quick review with some possible scope creep for @matthewtoast based on findings.

This PR presents machine-readable lint and test output for CI through new package-level commands `yarn lint-report` and `yarn test-report`. While I was in there, I cleaned up house style for our remaining TypeScript modules and hacked around some leaked handles in some tests. Projects using the standard linter were migrated to use the ESLint equivalent in order to get suitable output. Finally, I cleared out some heavy unused dependencies in plumbing and moved a small number of actually-used devDependencies to their proper place. TBD what this will do to app size, but it should have some impact.

In terms of possible scope creep: glass, serialization, timeline, and websockets all have leaked handles somewhere in their tests. Rather than spend time on this, I shut them with a hack (look for files called `99_Exit.test.js`). I think it would be good to investigate with e.g. `leaked-handles` to figure out what's actually happening here. Might even be a clue toward some of our big perf challenges. Finally, plumbing tests aren't passing at all, I suspect due to some of the mc refactoring…and we'll probably have some leaked handles there too. I left it out of the test reporting for now so that "instability" in CI is meaningful, but we should fix.


roperzh:
I know this is on Matthew to review, but in the spirit of bringing extra help I can confirm that:

- Both commands work for me
- Both commands generated proper `.xml` and `.tap` files with the right contents

stristr:
Actually only assigned to @matthewtoast for visibility due to the new issues I noted, so gonna merge this and circle back in Asana with the (potentially backlog) tasks that arose from this work.

PR #175, opened 2018-01-04
Fix: set preview mode background to root element

roperzh:
> Difficulty: easy review
> [Asana ticket](https://app.asana.com/0/482906470227387/505957877019445)

When preview mode is activated, we are setting a white background
(vs the default gray) to reinforce the idea that something is
happening.

This sets the background to the root component of glass, avoiding
the illusion at certain stage sizes that there's a strange border
around the stage.


tsortwehttam:
LGTM.

stristr:
Merging this (failed health check is from grandfathering in).

PR #177, opened 2018-01-04
React 16

stristr:
The reviewer should plan to spend some time on sanity checking the entire app and checking Player demos, but this actually turned out to be a relatively simple update with a small amount of code to review. Most of the changes here are upgrading demos to use ES6 classes (`React.reactClass` is deprecated after React 15) and rejiggering the way we handle refs to play nicely with both React 16 and React 15.

On that note, I was able to do this in a way that is backward compatible to the last version of React we nominally supported, so I rerolled the peer dependency to be an `||` statement without issue.


tsortwehttam:
LGTM.

Did quick local review:

- Player demos look good
- App behavior looks good after a quick smoke test

Minor tangential note I realized when testing this:

- Our node version is now "officially" 7.9.0 for Electron alignment, but the code in `@haiku/player/demo/server.js` requires a higher version of node (8.9.3?) otherwise you get a syntax error when running `$ yarn develop`

roperzh:
Player demos and app behavior LGTM too.

I'm wondering if is worth to have at least player demos for React 15.x too, since it will be "officially supported"

PR #179, opened 2018-01-05
Various UI tweaks

taylorpoe:
Tackles these cards:
* [Timeline should remember whether user chose frame or ms](https://app.asana.com/0/480796620059175/505957877019442)
* [Remove commenting option until fully supported](https://app.asana.com/0/480796620059175/512476954907134)
* [fix: when created keyframes on collapsed properties (e.g. on Position itself), user is presented with an error toast](https://app.asana.com/0/480796620059175/506363302552360)

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -1594,27 +1594,27 @@ export class Glass extends React.Component {
     const sourcePath = selectedElement && selectedElement.staticTemplateNode && selectedElement.staticTemplateNode.attributes && selectedElement.staticTemplateNode.attributes['source']
     const sketchAssetPath = sourcePath && sourcePath.split(/\.sketch\.contents/)[0].concat('.sketch')
 
-    items.push({
-      label: (this.state.doShowComments)
-        ? 'Hide Comments'
-        : 'Show Comments',
-      enabled: this.state.comments && this.state.comments.length > 0,
-      onClick: () => {
-        this.setState({ doShowComments: !this.state.doShowComments })
-      }
-    })
+    // items.push({
+    //   label: (this.state.doShowComments)
+    //     ? 'Hide Comments'
+    //     : 'Show Comments',
+    //   enabled: this.state.comments && this.state.comments.length > 0,
+    //   onClick: () => {
+    //     this.setState({ doShowComments: !this.state.doShowComments })
+    //   }
+    // })
 
-    items.push({
-      label: 'Add Comment',
-      onClick: () => {
-        this._comments.build({
-          x: menu.getLastX(),
-          y: menu.getLastY()
-        })
+    // items.push({
+    //   label: 'Add Comment',
+    //   onClick: () => {
+    //     this._comments.build({
+    //       x: menu.getLastX(),
+    //       y: menu.getLastY()
+    //     })
 
-        this.setState({ comments: this._comments.comments, doShowComments: true })
-      }
-    })
+    //     this.setState({ comments: this._comments.comments, doShowComments: true })
+    //   }
+    // })
 
     items.push({ type: 'separator' })
 
--- a/packages/haiku-serialization/src/bll/Project.js
+++ b/packages/haiku-serialization/src/bll/Project.js
@@ -320,7 +320,7 @@ class Project extends BaseModel {
   emitHook (...args) {
     const method = args.shift()
     const metadata = args.pop()
-    if (metadata.from === this.getAlias()) {
+    if (metadata && metadata.from === this.getAlias()) {
       this.emit.apply(this, ['update', method].concat(args))
     } else {
       // Otherwise we received an update and may need to update ourselves
@@ -332,7 +332,7 @@ class Project extends BaseModel {
     const method = args.shift()
     const metadata = args.pop()
     // If we originated the action, notify all other views
-    if (metadata.from === this.getAlias()) {
+    if (metadata && metadata.from === this.getAlias()) {
       log.info(`[project (${this.getAlias()})] method hook: ${method}`)
       this.batchedWebsocketAction(
         method,
--- a/packages/haiku-timeline/src/components/ComponentHeadingRow.js
+++ b/packages/haiku-timeline/src/components/ComponentHeadingRow.js
@@ -114,10 +114,10 @@ export default class ComponentHeadingRow extends React.Component {
               }}>
               {(this.props.row.isExpanded() || hasAttachedActions)
                 ? <EventHandlerTriggerer
-                    element={this.props.row.element}
-                    row={this.props.row}
-                    boltColor={boltColor}
-                    onEventHandlerTriggered={this.props.onEventHandlerTriggered}
+                  element={this.props.row.element}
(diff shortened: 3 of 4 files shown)


packages/haiku-serialization/src/bll/Project.js
```diff
@@ -320,7 +320,7 @@ class Project extends BaseModel {
   emitHook (...args) {
     const method = args.shift()
     const metadata = args.pop()
-    if (metadata.from === this.getAlias()) {
+    if (metadata && metadata.from === this.getAlias()) {
```
taylorpoe, line 323: Spooky that sometimes `metadata` isn't present?

packages/haiku-glass/src/react/Glass.js
```diff
@@ -1594,27 +1594,27 @@ export class Glass extends React.Component {
...
+    //   label: 'Add Comment',
+    //   onClick: () => {
+    //     this._comments.build({
+    //       x: menu.getLastX(),
+    //       y: menu.getLastY()
+    //     })
+
+    //     this.setState({ comments: this._comments.comments, doShowComments: true })
+    //   }
+    // })
```
stristr, line 1617: Would you mind putting this behind a feature flag instead? Steps would be:

1. Add a new flag [here](https://github.com/HaikuTeam/mono/blob/master/packages/haiku-common/config/experiments.json), like:
```
"CommentsUi": {
  "development": false,
  "production": false
}
```

2. Also add the experiment to the enum [here](https://github.com/HaikuTeam/mono/blob/master/packages/haiku-common/src/experiments/index.ts#L16).
3. Instead of commenting out code, guard it with `if (experimentIsEnabled(Experiment.CommentsUi)) { ... }`.  There are some example of this using `experimentIsEnabled(Experiment.MultiComponentFeatures)` throughout the code base.

The main benefit of this approach is it gives an explicit API for the symbolic act of commenting out code. Also, a less annoying way of setting up the experiment is Coming Soon™.

packages/haiku-serialization/src/bll/Project.js
stristr: Yes…but looks like this isn't due to a change in this PR, correct? I filed a [ticket](https://app.asana.com/0/480796620059175/514892342431432/f) because this is probably a bug.

stristr:
Feature and code LGTM, except for request above to use an experiment. There also is a minor merge conflict on the relevant lint.

Also a small nit: do we really want `timeDisplayMode` to be a global setting? I sort of think of it as a per-project setting…i.e. it's conceivable that I would want one project to be in seconds and another in frames. You could use `this.project.getFolder()` as a unique ID for hashing the different values if you think this is worthwhile, e.g.

```
  // getter
  user.getConfig('timeDisplayModes').then(
    (timeDisplayModes) => {
      if (timeDisplayModes && timeDisplayModes[this.project.getFolder()]) {
        this.setState({timeDisplayMode: timeDisplayModes[this.project.getFolder()]})
      }
    }
  );

  // setter
  const timeDisplayModes = this.user.getConfig('timeDisplayModes') || {};
  timeDisplayModes[this.project.getFolder()] = mode;
  this.user.setConfig('timeDisplayModes', timeDisplayModes);
```

packages/haiku-serialization/src/bll/Project.js
taylorpoe: No, it's not due to this PR. The issue is also noted on this other [ticket by Zack](https://app.asana.com/0/480796620059175/506363302552360). So my "fix" here takes care of the symptom, but perhaps there's something ickier happening when `metadata` isn't present. Let me know if you'd like me to leave this check-in, or let the error ride until you look into that other ticket you filed.

packages/haiku-glass/src/react/Glass.js
taylorpoe: Yeah makes sense. I'll make the change.

packages/haiku-serialization/src/bll/Project.js
stristr: Yeah, I think in this case since we're already tracking it, it's good to suppress the error and helps us get a release out sooner. I haven't noticed any awful side effects from this with some light testing.

stristr:
After a little more reflection on my 2:00 AM drive-by, I still like the idea but also think it's useful knowing what the user's last setting was to use as a default for new projects. My final, more complex answer (untested, might need some tweaking), is:

```
  // getter
  user.getConfig('timeDisplayModes').then(
    (timeDisplayModes) => {
      if (timeDisplayModes && timeDisplayModes[this.project.getFolder()]) {
        this.setState({timeDisplayMode: timeDisplayModes[this.project.getFolder()]})
      } else {
        user.getConfig('defaultTimeDisplayMode').then((defaultTimeDisplayMode) => {
          timeDisplayMode && this.setState({defaultTimeDisplayMode})
        })
      }
    }
  );

  // setter
  const timeDisplayModes = this.user.getConfig('timeDisplayModes').then((timeDisplayModes) => {
    this.user.setConfig(
      'timeDisplayModes',
      {
        ...timeDisplayModes,
       [this.project.getFolder()]: mode 
      }
    )
    this.user.setConfig('defaultTimeDisplayMode', mode)
  })
```

taylorpoe:
@stristr 
> Also a small nit: do we really want `timeDisplayMode` to be a global setting?

In some form or another, I think we do. I agree that ultimately per-project is ideal, but I wouldn't want to start there at the expense of not also doing this default setting. IMO the default is the biggest win here — as most users _(hypothesis)_ are going to be _either_ millisecond users or frames users and won't likely have any variation between the two between in their projects.

So if we had the setting per project, the user (depending on their preference) still might need to change the setting once for every new project. I'd like to first save them from needing to do this; so what I've added here is a hack for the average user to at least get their default set, so when they create a new project they won't have to change it every time. The way forward, in my mind, is to leave it like this for now and then when we build the settings page (on the dashboard), have a Time Display Mode Default setting to be able to be set globally from there, then in each project change the toggle to save the setting for each project individually in the way that you've outlined in your comment.

Agree/disagree?

taylorpoe:
Looks like we were both writing long-ish comments at the same time. Mine is not a response to your new one. Reading through now.

taylorpoe:
@stristr Ok, I dig your new solution as it accomplishes the "default" win here — which I'm most interested in — and is not nearly as hacky as I outlined in my "first step" summary. I'll make the changes and drop a note here when the commits are in. Thanks for the solution!

stristr:
Thanks! I was having the same thoughts as you (a little late to the party but I get it). 

Note: I edited the comment above to consistently call the setting `defaultTimeDisplayMode`, but actual name is your call.

stristr:
(Also edited to use the promise of `this.user.getConfig(...)` in the setter, my bad.)

tsortwehttam:
CPR to this thread. 💓 What's still needed to get this LGTM'd and merged?

taylorpoe:
This PR is no longer relevant. Keeping open only for service as a reference for its replacement: #192. Should #192 be accepted, this PR can be ignored and closed.

roperzh:
Closing in favor of #192

PR #180, opened 2018-01-05
Implement x/y-constrained dragging (shift) & duplicate dragging (alt)

zackbrown:


packages/haiku-creator/src/react/Creator.js
```diff
@@ -352,7 +352,7 @@ export default class Creator extends React.Component {
        */
       (user) => {
         this.user = user
-
+
```
stristr: Thanks, this line of whitespace helped me notice that the `"standard-jsx"` linter I switched this project over to doesn't actually include the `"standard"` linter rules we're supposed to have in these projects. Kind of moot because I plan going to replace it with `eslint-config-haiku` soon anyway, but for now I'll fix it here and commit.

stristr:
Very very nice! As noted above, a lint error I noticed by eye helped me fix the correct application of the Standard style to the React projects—now that this is in place, there are three lint errors requiring manual resolution (you can see them [here](https://ci.haiku.ai/job/Health/job/Haiku/16/checkstyleResult/) in Hai-CI, or with `yarn lint-all`).

Some minor UX feedback:
 - As a design tool user I expect when shift-drag locks onto an x-axis, the y-axis is locked out (and conversely). When I'm trying to edge something due left a couple of pixels, the feature is a little hard to use because if my mouse strays north of dead center mid-drag, I switch into vertical drag mode without wanting to.
 - Whether there's a good way to upsert a copy of an element without stealing from the paste API, we should probably fix the "append random string to ID" methodology because it's so unusual. Sketch copies layer `X` to `X copy`, `X copy 2`, etc

Neither of the above is a dealbreaker, but would be nice to fix if time permits.

zackbrown:
I trapped the lint and re-rebased off master — so this should be good to merge unless anyone notices other issues.

re: the UX points you mentioned:

 - **Shift-dragging over small distances:**  I studied both AI and Sketch's approaches to shift-dragging when designing the UX here — they handle it pretty differently; Sketch accumulates the "greater delta" of X or Y axis, and keeps on you that axis until you exceed that delta on the other axis.  I see how this addresses the small nudge case (you can go "farther along an axis", then drag it back and nudge precisely,) but it also felt sloppy to me that you can't switch axes mid-drag (admittedly, probably related to my AI background.)  AI handles it very similarly to my implementation here, insofar as small deltas are hard to do unless you zoom, but you have full control over whether you want to jump to horizontal or vertical mode (they also support 45-degree increments, which wouldn't be too tough to add.)  Arrow-key nudging works well for precise, vertical/horizontal nudges.

 - **"Element name Copy"** this is a good point, and is probably the better UX.  I don't think it should block shipping this feature — mostly, I want to shift my eng efforts to focus on some of the user account stuff — but I filed a task to revisit this during Mercury.

stristr:
Mergin'!

PR #181, opened 2018-01-05
fix: bypass the 'request' module's strictSSL in browser.

stristr:
[Asana ticket](https://app.asana.com/0/480796620059175/501816230772807/f)

A fast review that merits some explanation. Due to some unaccountable weirdness with the version of webpack resolved in `share-page`, the strictSSL check in the `'request'` Node library (polyfill for [Request](https://caniuse.com/#feat=fetch)) is failing on a fresh `yarn install` due to some sideloaded library invisibly rewriting the response object mid-flight. There's really no way to disable "strict SSL" in any modern browser, and functionally zero risk of a man-in-the-middle attack that would be caught by this polyfill but not the browser itself, so simply turning it off seemed like the most reliable workaround.

If somebody were to actually rewrite `Object.prototype.toString` to beat the "are we in a browser?" check I used in the Node context, the worst case scenario is they'd use our software to spoof secured requests to inkstone instead of writing their own.

### DIFF
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -21,7 +21,11 @@ const ENDPOINTS = {
 };
 
 let request = requestLib.defaults({
-  strictSSL: true,
+  // Delegate to the browser to handle "strictSSL" if we're in a browser context. We have seen some false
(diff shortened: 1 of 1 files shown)


roperzh:
Super interesting way to deal with this! LGTM :shipit:

PR #182, opened 2018-01-06
Fixes II

tsortwehttam:
OK to merge. Short review:

What:

- Fix for metadata being undefined https://app.asana.com/0/515552372320151/514892342431432
- Clone project if necessary https://app.asana.com/0/515552372320151/487676779427754

To reviewer:

We need to clone the project if the folder doesn't already exist -- but only if that project _isn't_ a newly project (only clone if there's actual remote content we need to grab). So, rather than go down a rabbit hole trying to solve that race, what I'm doing is just creating the project folder at the same time that the project is newly created, so it exists as of that next should-clone check. Fine for an LOD hack I think.


stristr:
LGTM!

PR #184, opened 2018-01-06
Fix issue causing large project init to hang for several minutes

tsortwehttam:
OK to merge. Short review.

What:

- Fixes https://app.asana.com/0/481696547044295/510407326169535
- Includes addition of some fixtures I used to test this bug which will be useful for the future

To the reviewer:

- This is the main thing: https://github.com/HaikuTeam/mono/compare/hung-project-fix?expand=1#diff-7433d312b35db871f4e4161899095cfeR2263
- The large project isn't the most pleasant thing to edit even despite this. Just noting this to make it clear that I'm just fixing this specific bug and future issues will be addressed separately.


stristr:
Perf **much** improved, LGTM, merging.

PR #185, opened 2018-01-06
Fix leaky player

stristr:
[Asana task](https://app.asana.com/0/480796620059175/515552372320152/f)

Turned out to be a one-liner to fix a major memory leak.

783ac163 introduced a perf regression with [this overzealous use of the object spread notation](https://github.com/HaikuTeam/mono/commit/783ac163f331e79ea3e2d2c8b6402aeaf54b45c1#diff-821c3c0323b1418629401b82f32760d9R157) in a super-obscure way. From tick to tick, layout is patched from the cached `_flatManaTree` if it changed. (The relatively new trick where we cache the flat mana tree vs. recalculating it on every tick, which gives a net perf boost, is actually what made this leak possible!) The problem is that by patching `cachedNode.layout` through vanities, we end up holding on to `catchedNode.layout.computed` from the previous tick. So the `{...layoutSpec}` added in 783ac163 is actually creating an object that looks like this on first tick:

```{layout: {computed: {...}, ...}}```

and this on second tick:

```{layout: {computed: {computed: {...}, ...}, ...}}```

and this on third tick:

```{layout: {computed: {computed: {computed: {...}, ...}, ...}, ...}}```

Which turns out to be a 2GB memory leak in ~20 minutes on a project with a bajillion nodes on loop—just a cascading vortex of `Float32Array`s holding 16-element vectors representing affine transformation matrices from previous ticks.

[timeline slows down over time](https://app.asana.com/0/480796620059175/511844918286659/f) is likely to be a side effect of this too.

### DIFF
--- a/packages/@haiku/player/src/Layout3D.ts
+++ b/packages/@haiku/player/src/Layout3D.ts
@@ -146,6 +146,8 @@ function multiplyArrayOfMatrices(arrayOfMatrices: number[][]): number[] {
 }
 
 function computeLayout(layoutSpec, currentMatrix, parentsizeAbsoluteIn) {
+  // Clean out the existing computed layout from the layout spec, if it exists.
+  delete layoutSpec.computed;
   const parentsizeAbsolute = parentsizeAbsoluteIn || {x: 0, y: 0, z: 0};
 
   if (parentsizeAbsolute.z === undefined || parentsizeAbsolute.z === null) {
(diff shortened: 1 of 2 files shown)


stristr:
https://www.youtube.com/watch?v=gh0xYqjdhRk

tsortwehttam:
LGTM; merging.

Minor thing I once heard about that @stristr might be interested in: https://stackoverflow.com/questions/23455678/pros-and-cons-of-dictionary-mode not sure if this could/would impact us, but I thought I'd mention it because it's been a factoid in my brain for a while and I finally get to share it!

stristr:
That's amazing…definitely first I have heard of dictionary mode in V8! [Looks like fixed in Chrome 60](https://jsperf.com/test-dictionary-mode), and I got a good result on my Pixel 2 as well…although it definitely looks like it would have been worth rewriting a few months ago based on those benchmarks. I think let's ignore this for now, but also never forget it.

PR #186, opened 2018-01-06
fix: grandfather in support for plain "backgroundColor" style setter.

stristr:
[Asana task](https://app.asana.com/0/480796620059175/515737276298249/f)

Very short review but prepared for @matthewtoast to reject the solution in favor of something that doesn't leave technical debt behind.

This is a suitable one-liner to grandfather in old Haikus created with `backgroundColor` instead of `style.backgroundColor`. There are more elegant ways to do this, but this definitely works.


tsortwehttam:
LGTM; this is what I'd do for now; merging.

PR #187, opened 2018-01-06
Security Things

tsortwehttam:
Short review.

What:

- Don't allow eval
- Don't disable security protections on webview

Although we don't technically run any untrusted code, since we do allow users to write and execute their own code, it feels safer to plug these more obvious vulnerabilities.


tsortwehttam:
On second thought, removed "OK to merge" in case anybody has input on whether there could be side effects of e.g. enabling web security in the <webviews>. I did test locally and didn't see any obvious issues.

tsortwehttam:
Merging as I haven't seen any adverse effects to this.

PR #188, opened 2018-01-06
Fix gradual UI slowdown during timeline looping playback

tsortwehttam:
OK to merge. Medium review.

What:

- Modify meta programming hack to avoid memory leak
- Refactor Timeline UI React components to update in a more granular way
- Debounce some timeline-frame updates

Before these changes, a simple project would see the Timeline UI playback slow down to total stall after 10 or 20 seconds.

After these changes, the Timeline UI can play on a loop for over 1 minute with only a very slight apparent slowdown.

### DIFF
--- a/packages/haiku-serialization/src/bll/BaseModel.js
+++ b/packages/haiku-serialization/src/bll/BaseModel.js
@@ -54,9 +54,11 @@ class BaseModel extends EventEmitter {
       }
     }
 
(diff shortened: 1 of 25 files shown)


tsortwehttam:
I'm making a judgment call and merging now since I have tasks that may touch this code before others on the team will be available to review.

PR #191, opened 2018-01-08
Require token when connecting to our websocket server

tsortwehttam:
OK to merge; short review.

What:

- Create a hard-to-guess token when we start up, and pass that token to the subviews; then require that all clients provide that token when connecting; forbid those that do not. This prevents unauthorized connections to our websocket server (e.g. a malicious website trying to connect to our websocket server for nefarious purposes).

To the reviewer:

- In the interest of speed, I intentionally did not do any clean up work around the pathways which are involved in setting up the websocket clients nor passing parameters into them. But in looking at this code, it will eventually need a refactor.


stristr:
LGTM, merging.

PR #192, opened 2018-01-09
Allow user to save time display setting (frames vs seconds) per project

taylorpoe:
It also updates a default time display setting every time the setting is changed on any project, so that new projects will use the user's preference (or their last preference at least).

### DIFF
--- a/packages/haiku-serialization/src/bll/Timeline.js
+++ b/packages/haiku-serialization/src/bll/Timeline.js
@@ -88,6 +88,14 @@ class Timeline extends BaseModel {
     return this._timeDisplayMode
   }
 
+  setTimeDisplayMode (newMode) {
+    this._timeDisplayMode = newMode === 'seconds'
+      ? Timeline.TIME_DISPLAY_MODE.SECONDS
+      : Timeline.TIME_DISPLAY_MODE.FRAMES
+
+    this.emit('update', 'time-display-mode-change')
+  }
+
   toggleTimeDisplayMode () {
     if (this.getTimeDisplayMode() === Timeline.TIME_DISPLAY_MODE.FRAMES) {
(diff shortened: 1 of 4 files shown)


taylorpoe:
@stristr I'm asking you to review this one because it is a replacement PR for #179. This should be, more or less, the suggested behavior you outlined on that PR. If this PR is suitable, that PR should be ignored and closed.

The reason for the new PR is that the old one was going to need a git push force and I was a bit wary of using that type of command.

Also of note, it took me an extremely long time to figure out how to get this behavior working because over the weekend some of these pieces were refactored into the BLL. I tried to get all the new pieces here working in the BLL's new `toggleTimeDisplayMode()` call https://github.com/HaikuTeam/mono/blob/master/packages/haiku-serialization/src/bll/Timeline.js#L91 but I couldn't successfully get the user settings calls to work with Envoy, USER_CHANNEL, etc. Perhaps this should be refactored in the future to do that, but for now I've already certainly over-strapped my time-this-task-should-required allotment. Let me know what you think.

packages/haiku-serialization/src/bll/Timeline.js
```diff
@@ -88,6 +88,14 @@ class Timeline extends BaseModel {
     return this._timeDisplayMode
   }
 
+  setTimeDisplayMode (newMode) {
+    this._timeDisplayMode = newMode === 'seconds'
```
stristr: Nit: this could just be `this._timeDisplayMode = newMode`.

TypeScript would give us compile-time safety with an enum, but since everything else is behind a similar `switch` statement in Timeline UI, seems fine to just set it and forget it.

stristr: (I'll commit this one change since this is otherwise good to go.)

stristr:
LGTM, merging.

PR #194, opened 2018-01-09
Enable timeline scrubbing from gauge without forcing the user to grab the actual scrubber bulb

taylorpoe:
[Ticket](https://app.asana.com/0/480796620059175/505957877019440)

> Click anywhere on timeline to reposition scrubber, instead of having to mousedown on the ticker bulb

Also getting [scrub behavior that moves](https://cl.ly/0W2Y2H022n2M) the horizontal scroll back to zero if need-be. In the future this should be able to add more frames as well — seemed out of scope for now.


tsortwehttam:
LGTM, merging.

PR #195, opened 2018-01-10
fix: delete stading-zeroth keyframe if sibling is dragged to zero

taylorpoe:
[Ticket](https://app.asana.com/0/480796620059175/506363302552363)

bug: when dragging a segment away from starting-point 0 then trying to drag back, we get two zero-keyframes

### DIFF
--- a/packages/haiku-serialization/src/bll/Row.js
+++ b/packages/haiku-serialization/src/bll/Row.js
@@ -302,7 +302,6 @@ class Row extends BaseModel {
   }
 
   ensureZerothKeyframe (metadata) {
-
     if (!this.hasZerothKeyframe()) {
       this.createKeyframe(this.getFallbackValue(), 0, metadata)
     } else {
--- a/packages/haiku-serialization/src/bll/Row.js
+++ b/packages/haiku-serialization/src/bll/Row.js
@@ -302,12 +302,15 @@ class Row extends BaseModel {
   }
 
   ensureZerothKeyframe (metadata) {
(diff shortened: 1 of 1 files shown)


packages/haiku-serialization/src/bll/Row.js
```diff
@@ -302,8 +302,12 @@ class Row extends BaseModel {
   }
 
   ensureZerothKeyframe (metadata) {
+    const keyframes = this.getKeyframes()
     if (!this.hasZerothKeyframe()) {
       this.createKeyframe(this.getFallbackValue(), 0, metadata)
+    } else if (keyframes[1] && keyframes[1].ms === 0) {
+      // if a keyframe was dragged onto zero position we want to delete the zero keyframe
+      this.deleteKeyframe(keyframes[0], metadata)
     }
```
stristr: Not sure exactly when/why this would happen, but it might not be complete (what if `keyframes[2] && keyframes[2].ms === 0`?)…

I think a nice concise way to always let the last keyframe at 0 dominate is:

```
else {
  while (keyframes.length > 1 && keyframes[1].ms === 0) {
    this.deleteKeyframe(keyframes.shift(), metadata)
  }
}
```

(`keyframes.shift()` removes and returns `keyframes[0]`.)

taylorpoe: Tk for the code! Seems to work for multiple when I tried it out. New commit below.

tsortwehttam:
LGTM, merging.

PR #198, opened 2018-01-11
 feat: swap out browserify with rollup for generating standalones.

stristr:
Minimalist take on a task that should also be extended to how we create the hosted versions used on the share page. I threw in `uglify-es` because it was "cheap" and pays dividends in terms of how much smaller we can get the bundle.

The main impetus was to un-stick publish without rolling back or hacking into the (security-positive) work in c74b80205, which failed due to an obscure eval call somewhere deep in a `browserify` dependency.

In the spirit of hack-slashing to success, I removed `packages/@haiku/player/src/vendor/just-curves/internal/cssFunction.ts` and its references because it was both unused and causing rollup issues due to some circular imports.

This PR cuts the size of `index.standalone.js` in half for an average-sized Haiku, which is also cool :smile:


stristr:
Merging to test in build.

PR #200, opened 2018-01-12
Fix: A few UI bugs

taylorpoe:
**Fix broken treeline**
Removes the treeline on the timeline. I wasn't a huge fan of it in the first place — too much visual noise for a simple two-tier hierarchy. Feels cleaner now. I also put our component icon next to the root component, which helps with the tier distinction.

**FIx incorrect location of keyframe creation on right click**
[ticket](https://app.asana.com/0/482906470227367/505957877019443)
fix timeline issue where keyframes sometimes don't show up where created (seems to be at certain widths and/or zoom levels)

**Fix Right click create keyframe within a segment appears to be broken**
[ticket](https://app.asana.com/0/480796620059175/519080338804547)

### DIFF
--- a/packages/haiku-timeline/src/components/ClusterRow.js
+++ b/packages/haiku-timeline/src/components/ClusterRow.js
@@ -52,16 +52,6 @@ export default class ClusterRow extends React.Component {
           cursor: 'pointer'
         }}>
         <div>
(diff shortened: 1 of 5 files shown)


stristr:
LGTM, merging.

PR #202, opened 2018-01-13
noop eval instead of throwing

tsortwehttam:
Instead of throwing when `eval()` is invoked, make it a silent noop.

Fixes https://app.asana.com/0/480796620059175/521910445960082

Even despite the Rollup change, arguably we still don't want to crash the app in the case that some third-party library or untrusted code happens to try to `eval()` something, hence this PR


stristr:
LGTM. Is this still WIP?

PR #206, opened 2018-01-14
Feat: add haiku as loader for timeline skeleton state

taylorpoe:
OK to merge

Adds a skeleton state loader for the timeline (an actual haiku).

Asana task link: https://app.asana.com/0/482906470227367/505957877019454

Changes in this PR:

- haiku-component loading state for timeline
- changes to the script to allow for yarn install of haiku to work

Notes for the reviewer:

-@stristr did we want these changes to `scripts/helpers/isHaikuDep.js` in here?

Completed checkin tasks:


packages/haiku-creator/src/react/components/Timeline.js
```diff
@@ -94,7 +97,8 @@ export default class Timeline extends React.Component {
       if (typeof this.props.onReady === 'function') this.props.onReady()
     })
 
-    while (this.mount.firstChild) this.mount.removeChild(this.mount.firstChild)
+    setTimeout(() => { this.setState({ finishedInjecting: true }) }, 7000)
```
stristr, line 100: Can we do something more robust than waiting for `7s`? With some recent perf improvements, small projects definitely shouldn't take this long….

stristr:
Re: deps/not breaking the build: the quickest way to clean this up would be to clone and push it directly. Something like this:

```
cd /tmp
haiku clone TimelineSkeletonState
cd TimelineSkeletonState
# Edit package.json to remove "engines" and bump the version to 0.0.8. While at it, set `@haiku/player` version to 2.3.71
git add package.json
git commit -m 'manual: upgrade package.json'
git tag -a 0.0.8
git push --tags
```

Depending on how many of these we end up having, I can write a write a script to do this for you.

yarn.lock
```diff
@@ -22,6 +22,22 @@
...
 
+"@haiku/player@2.3.49":
+  version "2.3.49"
+  resolved "https://reservoir.haiku.ai:8910/@haiku%2fplayer/-/player-2.3.49.tgz#eeb83edd96cea8e4d2e488de225941585141c314"
+
+"@haiku/taylor-timelineskeletonstate@^0.0.6":
+  version "0.0.6"
+  resolved "https://reservoir.haiku.ai:8910/@haiku%2ftaylor-timelineskeletonstate/-/taylor-timelineskeletonstate-0.0.6.tgz#ed8d0c07cb035214a6edc45da1854cf59fafcda2"
+  dependencies:
+    "@haiku/player" "2.3.49"
```
stristr: It looks like this shouldn't be here…probably another `yarn install` from mono root will clean it out.

packages/haiku-creator/src/react/components/Timeline.js
taylorpoe: I looked for a while to try to find a nice way to listen for it but couldn't get anything wholesome working. The webview is injected immediately on mount, but isn't fully set up — this is when the loader is actually needed (surprisingly). The `this.mount.appendChild(this.webview)` call happens here but this isn't instant. And so you'd  think the loader would be needed down on haiku-timeline but this isn't actually the time when the longest loading stretch takes place. 

Because I couldn't find a way to track down this specific stretch of time, I've set it so the loading screen loads for "too much time" (7s) and the webview displays over it (z-index-wise); obscuring it. Then the loader is removed after 7s. Again, it's jank but shouldn't cost anything other than the loader playing for a bit too long.

tsortwehttam: Can we put this loader in the Timeline project (as opposed to Creator) so we don't have to do this dance?

tsortwehttam:
> caveat: only some; does ui work mandate this?

99% of the time I'd say yes. Part of the reason we have this ongoing battle with instability is that we wait until _lots of changes_ have been merged in to do any E2E and full QA testing. That makes it harder to trace down the cause of bugs (we can't intuit whether one change or another might have caused the problem).

And often seemingly superficial changes have a significant impact on other distant things. For an (artificial) example here, I notice that a `z-index` and `position: relative` was added to an element. Could that affect parts of the tour? Or styling for preview mode? (Again, fake examples.) If we're rolling a build and doing last-minute QA and _then_ we see some issue related to this, we'll have to drop what we're doing then and scramble to fix it, which results in even more self-inflicted pain 📟 

So... the goal is to improve our product cycle health by "baking in" testing and QA into our regular development schedule, rather than have it be an afterthought. In other words, however long you or I might estimate it will take to make some set of changes, we should always assume an add'l +15 minutes tacked on the end for QA. Eventually it will become a good habit like brushing your teeth before bedtime.

> Expanded the QA checklist

I noticed that this is not checked off... but this is a perfect commit to do this and it should only take a minute. All this means is to add an item to the QA Checklist representing the change. Something like "A Haiku loader appears on the timeline panel when the project is loaded." That way we know to check for it when we do full QA don't accidentally let a feature we all worked on slip through the cracks.

packages/haiku-creator/src/react/components/Timeline.js
taylorpoe: That was my first thought — to put it in haiku-timeline. I had it loading there originally, but on average the load duration was always about ~.5 sec — with the bulk of the actual loading time for the Timeline being in Creator ~3 sec.

taylorpoe:
@matthewtoast Right on. Thanks for the details. I wasn't sure how seriously I should be taking UI tweaks, but this makes sense — especially with the tour example. I'll start setting aside 15 more minutes for the teeth brushing. And will do that for this task too. Seems like it would be okay to sometimes have this flow: dev something and PR it for visibility, labeling it "WIP" then come back and do the QA and remove the "WIP" when it's done?

taylorpoe:
@stristr Okay, I followed your clone steps you listed out. Also updated the version of the haiku in Creator to match this change.

yarn.lock
taylorpoe: Yup looks like that worked.

taylorpoe:
@matthewtoast Would this be something I'd want to write a test for as well?

taylorpoe:
PR still WIP until I get a chance to got through the E2E QA

taylorpoe:
QA done; a couple things seemed broken for state setting and path-of-shape changing, but I don't see how this could be related to this PR. WIll file separate tickets.

tsortwehttam:
LGTM, merging.

PR #207, opened 2018-01-16
feat: add endpoint for self-service user creation.

stristr:
OK to merge, as long as we agree it's very unlikely someone will try to unobfuscate the transpiled code we `npm publish`…. and call Inkstone directly to create new users.

Short review.

[Asana task](https://app.asana.com/0/482906470227387/480796620059192)

Changes in this PR:

- [x] Adds new endpoint (`POST /user`) for self-service signup.

### DIFF
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -6,6 +6,7 @@ const ENDPOINTS = {
   PROJECT_CREATE: 'v0/project',
   LOGIN: 'v0/user/auth',
   CHANGE_PASSWORD: 'v0/user/password',
+  ORGANIZATION_CREATE: 'v0/organization',
   ORGANIZATION_LIST: 'v0/organization',
   PROJECT_LIST: 'v0/project',
   INVITE_PREFINERY_CHECK: 'v0/invite/check',
@@ -306,6 +307,32 @@ export namespace inkstone {
           cb(undefined, projects, httpResponse);
         } else {
           cb(safeError(err), undefined, httpResponse);
+        }
+      });
+    }
+
+    export interface OrganizationCreate {
+      Email: string;
+      Password: string;
+      OrganizationName: string;
+      NewsletterOptIn?: boolean;
+    }
+
+    export function create(organization: OrganizationCreate, cb: inkstone.Callback<boolean>) {
+      const options: requestLib.UrlOptions & requestLib.CoreOptions = {
+        url: inkstoneConfig.baseUrl + ENDPOINTS.ORGANIZATION_CREATE,
+        json: organization,
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      };
+
+      request.post(options, (err, httpResponse, body) => {
+        if (httpResponse && httpResponse.statusCode === 200) {
+          cb(undefined, true, httpResponse);
+        } else {
+          const errMessage = body as string;
+          cb(errMessage, undefined, httpResponse);
         }
       });
     }
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -6,7 +6,6 @@ const ENDPOINTS = {
   PROJECT_CREATE: 'v0/project',
   LOGIN: 'v0/user/auth',
   CHANGE_PASSWORD: 'v0/user/password',
-  ORGANIZATION_CREATE: 'v0/organization',
   ORGANIZATION_LIST: 'v0/organization',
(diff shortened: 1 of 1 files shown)


zackbrown:
As long as we're refactoring away from invite claiming, shouldn't we make this an action on the User resource instead of an Organization?

The primary action here is the creation of a user account — the creation of an organization is a downstream detail of that.  Further, at some point in the future we will support the creation of organizations, e.g. like setting up HaikuTeam in Github, which would collide with this resource/verb.

stristr:
In the current database design there is no join table between users and organizations, so the relations are User belongs to Organization, Organization has many User and Organization has many Project (User has many Project through Organization). So from an AR perspective the primary thing we're doing with this endpoint is creating an organization and its template projects, and creating a user account is a hack because we don't properly support multi-user. Anyway, that was how I was thinking about it…if User <-> Organization were many to many it would feel like more of a dilemma.

zackbrown:
User <-> Organization is expected to grow into a many-to-many relationship, like Github orgs and Slack workspaces.

Just want to make sure you know the product reasoning, since that was considered when originally designing the db schema and when writing [the spec for self-serve](https://haiku.quip.com/QyMSAf0a3avV#DBcACAZW1SV), with the included suggestion not to make this refactor.

Anyways, since you've taken the effort now for the benefit of our future colleagues and selves — we should consider that exposing this endpoint for Organizations sets us up for a future boobytrap when we actually need to expose to the user the ability to create an Organization.

stristr:
I would still say (not just because the database design does not reflect the future state, which is problematic) the implementation here is the best and easiest to refactor into the future state when we add a pivot table and migrate. We explicitly don't support the following actions right now:
 - Creating a user without also creating an organization (`POST /users` for an unauthenticated user, `POST /organizations/:id/users` for an authenticated organization admin inviting a user to their org).
 - Creating an organization without also creating a user (authenticated call to `POST /organizations`).

It's really not a boobytrap, just necessary refactoring when the requirements change. In the future, we'd probably make the two calls above, in order, for a new user signing up a new org.

zackbrown:
We _don't_ want to support `creating a user without also creating an organization` (because Organization acts as a smooshed polymorphic union of `UserAccountEntity` and the future `TeamAccountEntity` — and every User should have exactly one `UserAccountEntity`, which is where their short-username is stored)

Even in the case a new user is being added to an existing team, they need their own username and thus their own `UserAccountEntity`-flavored `organization`.

Also in the meantime, we _don't_ want to support stand-alone organization creation at all, especially not in the case when not also creating a user (that is, until we both spec out and build the organization mgmt features.)

So until we support multi-org: this is definitely an action on the User rather than the Organization. (I mean, [look at the params](https://github.com/HaikuTeam/mono/pull/207/files#diff-908953f7e78195f6e979c99976d220a5R314))

stristr:
> Even in the case a new user is being added to an existing team, they need their own username and thus their own `UserAccountEntity`-flavored `organization`.

Ah, I think this may be a significant point of disagreement. In GitLab, users have their own credentials—if you parse the detailed PR description in [Inkstone #23](https://github.com/HaikuTeam/inkstone/pull/23) one consequence of the more expressive permissions model is that a user-flavored organization is not a hard requirement and feels like an unneeded bit of bloat for the use case you describe. Every user owns their own Git namespace, so all we have to do is uncap their project limit from 0 to allow them to make their own projects. This is how I would design it—it's exactly how GitHub works. Sign up, optionally create organization(s). When GitHub and Slack built enterprise, neither built distinct entities for organizations and teams into the same product—they instead sharded their infrastructure into "instances". This feels like a straightforward and semantically deliberate approach to this problem.

With this in mind, a new user being added to an existing "team" is exactly what I had in mind when I mentioned creating a user without also creating an organization. So we're talking about the same use cases with (mostly) the same expectations.

For the endpoint: we don't support stand-alone creation of organizations *or* users, but when we eventually build a "pure" `POST /organizations`, it will do the same thing with basically the same code (create organization in DB, create organization in GitLab, copy template projects). But the "pure" `POST /users` will be different than what's here now, because we wouldn't know the org ID yet when creating the user. I don't think the number of user-specific parameters really means this is fundamentally a User endpoint—that's really just a side effect of how much easier it is to specify an org vs. a user 😄 !

stristr:
After productive in-person, modified this to `POST /user` here and in [Inkstone #23](https://github.com/HaikuTeam/inkstone/pull/23) (
[dc1df2d](https://github.com/HaikuTeam/inkstone/pull/23/commits/dc1df2d25e234bf67b2c305b2800705d927cd403)).

roperzh (approved):
LGTM

PR #215, opened 2018-01-19
Fix perf on timeline scrubbing

tsortwehttam:
OK to merge. (Didn't do a crazy amount of testing, but this seems fairly limited.)

Short review.

Asana task link: n/a, discovered issue during QA

Changes in this PR:

- [x] Instead of registering a new mouse move listener every time the mouse goes down in the gauge, just have one listener that performs work depending on a flag
- [x] Instead of calling `setState` to toggle `avoidTimelinePointerEvents` on every mouse move (which was causing a re-render from the timeline root on every mouse move), just mutate the state, and only toggle it in mouse down and mouse up


roperzh (approved):
LGTM, I have tested it and works properly for me

PR #216, opened 2018-01-19
Ensure action editor opens for frame 0

tsortwehttam:
OK to merge.

Short review.

Asana task link: https://app.asana.com/0/481696547044295/528663464193309

Changes in this PR:

- [x] Use numeric check instead of falsy which was interpreting `0` as "no frame"


tsortwehttam:
This was branched off of #215 so merge that one first plz.

roperzh (approved):
LGTM, and works fine. Ideally would be nice to have `isNumeric` in one place, even if it's in `haiku-serialization`, but I don't consider this a blocker.

PR #221, opened 2018-01-19
Launch perf #2

stristr:
Changes in this PR:

- [x] Elide unnecessary full flush renders during glass playback in favor of surgical cache-skipping patches, only when necessary. This greatly speeds up our ability to render a WIP Haiku. 
 - [x] Fix `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 501 update listeners added. Use emitter.setMaxListeners() to increase limit` on percy playback.
 - [x] Rehydrate mutable timelines during cache clearing, which we do frequently while mutating bytecode in place. Failing to do this was a longstanding player bug introduced by the perf pass, mainly due to my prior ignorance of how the component is reloaded during hot editing. The unnecessary full flush rendering we were always doing in Player was obscuring the negative side effects of not doing this, since the entire optimization routine is bypassed in full flush. This may indirectly clear the path to @matthewtoast's in-mem undo/redo moonshot!
 - [x] Fix `setState()` sometimes called an unmounted `<Library />` in Glass, which caused an occasional bug where the user is unable to select elements on the stage. I incorrectly described this as a tour bug as this was the first place I noticed it (sorry @roperzh), but it was an actual issue now fixed.
 - [x] Fix for [Gauge doesn't show enough values on longer projects](https://app.asana.com/0/480796620059175/529942638926528/f)
 - [x] Fix for [Parent label in timeline is HaikuC…mponent (HaikuComponent)](https://app.asana.com/0/480796620059175/529933088332841/f)
 - [x] Fix for [Race on forceUpdate() when collapsing timeline property rows](https://app.asana.com/0/480796620059175/530298508710782/f)
 - [x] Fix for [Triggering preview mode during timeline playback should stop timeline playback](https://app.asana.com/0/480796620059175/530308731229517/f)

### DIFF
--- a/packages/@haiku/player/src/ValueBuilder.ts
+++ b/packages/@haiku/player/src/ValueBuilder.ts
@@ -1210,6 +1210,7 @@ ValueBuilder.prototype.build = function _build(
   propertiesGroup,
   isPatchOperation,
   haikuComponent,
+  skipCache = false,
 ) {
   let isAnythingWorthUpdating = false;
 
@@ -1223,6 +1224,7 @@ ValueBuilder.prototype.build = function _build(
       timelineTime,
       haikuComponent,
       isPatchOperation,
+      skipCache,
     );
 
     // We use undefined as a signal that it's not worthwhile to put this value in the list of updates.
@@ -1273,6 +1275,7 @@ ValueBuilder.prototype.build = function _build(
  * @param haikuComponent {Object} Instance of HaikuPlayer
  * @param isPatchOperation {Boolean} Is this a patch?
  * @param skipCache {Boolean} Skip caching?
+ * @param clearSortedKeyframesCache
  */
 ValueBuilder.prototype.grabValue = function _grabValue(
   timelineName,
@@ -1309,7 +1312,7 @@ ValueBuilder.prototype.grabValue = function _grabValue(
 
   let computedValueForTime;
 
-  // Important: The ActiveComponent depends on the ability to be able to get fresh values via the skipCache optino
+  // Important: The ActiveComponent depends on the ability to be able to get fresh values via the skipCache option.
   if (isPatchOperation && !skipCache) {
     computedValueForTime = Transitions.calculateValueAndReturnUndefinedIfNotWorthwhile(
       parsedValueCluster,
--- a/packages/haiku-plumbing/src/Plumbing.js
+++ b/packages/haiku-plumbing/src/Plumbing.js
@@ -880,15 +880,6 @@ export default class Plumbing extends StateObject {
     // Params always arrive with the folder as the first argument, so we strip that off
     params = params.slice(1)
 
-    // This special method gets called frequently (up to 60 times per second) so fast-path it and don't log it
-    if (method === 'setTimelineTime') {
-      if (alias === 'timeline') {
-        return this.sendFolderSpecificClientMethodQuery(folder, Q_GLASS, method, params.concat({ from: alias }), () => {})
-      } else if (alias === 'glass') {
-        return this.sendFolderSpecificClientMethodQuery(folder, Q_TIMELINE, method, params.concat({ from: alias }), () => {})
-      }
-    }
-
     // Start with the glass, since that's most visible, then move through the rest, and end
     // with master at the end, which results in a file system update reflecting the change
     const asyncMethod = experimentIsEnabled(Experiment.AsyncClientActions) ? 'each' : 'eachSeries'
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -325,23 +325,23 @@ class ActiveComponent extends BaseModel {
     return cb(null, pretty(html))
   }
 
-  setTimelineTimeValue (timelineTime, skipTransmit, forceSeek, setNoMatterWhat) {
+  setTimelineTimeValue (timelineTime, skipTransmit = false, forceSeek = false, skipCache = false) {
     timelineTime = Math.round(timelineTime)
-    if (setNoMatterWhat || this.isPreviewModeActive() || timelineTime !== this.getCurrentTimelineTime()) {
+    if (timelineTime !== this.getCurrentTimelineTime()) {
       // Note that this call reaches in and updates our instance's timeline objects
       Timeline.setCurrentTime({ component: this }, timelineTime, skipTransmit, forceSeek)
-      this.forceFlush()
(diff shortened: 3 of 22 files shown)


tsortwehttam:
Going to look at this to see if I can find where the call needs to take place for the new blocker.

stristr:
Thanks, gonna eat 👍 😄 😸 🥅 🥘

stristr:
Ready for review. Details in PR description.

packages/@haiku/player/src/ValueBuilder.ts
```diff
@@ -1309,7 +1312,7 @@ ValueBuilder.prototype.grabValue = function _grabValue(
 
   let computedValueForTime;
 
-  // Important: The ActiveComponent depends on the ability to be able to get fresh values via the skipCache optino
```
tsortwehttam, line 1312: `-ion/-ino`, my favorite typo, turns mundane words into exotic elementary particles :atom:

packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -880,15 +880,6 @@ export default class Plumbing extends StateObject {
...
 
-    // This special method gets called frequently (up to 60 times per second) so fast-path it and don't log it
-    if (method === 'setTimelineTime') {
-      if (alias === 'timeline') {
-        return this.sendFolderSpecificClientMethodQuery(folder, Q_GLASS, method, params.concat({ from: alias }), () => {})
-      } else if (alias === 'glass') {
-        return this.sendFolderSpecificClientMethodQuery(folder, Q_TIMELINE, method, params.concat({ from: alias }), () => {})
-      }
-    }
-
```
tsortwehttam, line 891: I don't see a problem removing this, but I just want to check: why?

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -325,23 +325,19 @@ class ActiveComponent extends BaseModel {
     return cb(null, pretty(html))
   }
 
-  setTimelineTimeValue (timelineTime, skipTransmit, forceSeek, setNoMatterWhat) {
+  setTimelineTimeValue (timelineTime, skipTransmit = false, forceSeek = false) {
     timelineTime = Math.round(timelineTime)
```
tsortwehttam, line 329: 👍

packages/haiku-serialization/src/bll/Keyframe.js
```diff
@@ -28,19 +28,47 @@ class Keyframe extends BaseModel {
...
+    this._updateReceivers[source] = cb
+    return () => {
+      delete this._updateReceivers[source]
+    }
+  }
+
+  notifyUpdateReceivers (what) {
+    Object.keys(this._updateReceivers).forEach((receiver) => {
+      this._updateReceivers[receiver](what)
+    })
```
tsortwehttam, line 58: 👍

packages/haiku-serialization/src/bll/Project.js
```diff
@@ -944,7 +949,7 @@ function getCodeJs (haikuId) {
         elementName: 'div',
         attributes: {
           'haiku-id': '${haikuId}',
-          'haiku-title': 'HaikuComponent'
+          'haiku-title': '${haikuComponentName}'
```
tsortwehttam, line 952: 👍

packages/haiku-serialization/src/bll/Timeline.js
```diff
@@ -955,9 +955,9 @@ Timeline.getMillisecondModulus = function getMillisecondModulus (pxpf) {
   if (pxpf >= 15) return 50
   if (pxpf >= 10) return 100
   if (pxpf >= 5) return 200
-  if (pxpf === 4) return 250
-  if (pxpf === 3) return 500
-  if (pxpf === 2) return 1000
+  if (pxpf >= 4) return 250
+  if (pxpf >= 3) return 500
+  if (pxpf >= 2) return 1000
   return 5000
```
tsortwehttam, line 961: 👍

packages/haiku-timeline/src/components/ClusterInputField.js
```diff
@@ -36,17 +36,18 @@ class ClusterInputFieldValueDisplay extends React.Component {
...
   componentWillUnmount () {
+    this.throttledForceUpdate.cancel()
     this.mounted = false
     this.props.timeline.removeListener('update', this.handleUpdate)
   }
 
   componentDidMount () {
     this.mounted = true
+    this.throttledForceUpdate = lodash.throttle(this.forceUpdate.bind(this), 64)
     this.props.timeline.on('update', this.handleUpdate)
```
tsortwehttam, line 51: Note: `componentDidMount` may get called multiple times for a single React node.

tsortwehttam:
Code looks good; going to pull and review the functionality.

tsortwehttam:
LGTM after reviewing locally.

One note: My feel from editing Percy is that while on-stage performance is obviously _way_ better with these changes (both in editing and preview mode), scrubbing still seems to be slow (even slower, maybe?). I know you were focusing on stage stuff exclusively with this, but I just wanted to mention this so it's in the zeitgeist.

packages/haiku-plumbing/src/Plumbing.js
stristr: Ah, removed this for being confusing. setTimelineTime Is no longer used!

packages/haiku-timeline/src/components/ClusterInputField.js
stristr: …but not multiple times per mount, right? The point of this change was that you can't un-`cancel()` a throttled function, so to ensure `forceUpdate()` doesn't get called while the component is unmounted, cancel pending invocations on unmount and create a new throttled function on mount.

stristr:
I would buy that scrubbing is _slightly_ slower due to my lack of ambition re: getting to the root cause of timeline event emitter leaks—the whole reason the leaks added up to a warning about 500+ event emitters is the built-in React optimizations around component reuse are being trampled by (I think) overuse of `forceUpdate()`.

Bottom line, this bug cropped up because a set of components with the same key are getting constantly mounted, unmounted, and in many cases for a large project _destroyed/reconstructed_ during ordinary playback with nothing changing except the time. I figured it would be better to address the problematic side effects of this underlying sickness in an unambiguously readable way, and exorcise the demons of Timeline in a separate PR.

stristr:
@matthewtoast, would you say this is okay to merge?

tsortwehttam:
Yep. 👍 👍

PR #226, opened 2018-01-22
Assert code.js contents (crash if empty or invalid)

tsortwehttam:
NOT READY TO MERGE. (I have not tested yet.)

Short review.

Asana task link: n/a, responding to report in Slack about blank code.js causing corruption.

Changes in this PR:

- [x] If we did a bytecode update and the bytecode -> AST -> code pipeline gives us an invalid value or a blank string, throw an error (crash the app) to prevent bad content getting written to disk which may have the effect of corrupting projects.

Notes for the reviewer:

- I think this is a pretty good idea. **However, this is PR is very much provisional. We can close w/o merging if anyone has serious red flags.** My argument is that it's better to crash immediately -- and possibly even get the benefit of a full restart -- than it is to let the app continue to silently allow an insidious problem to get worse (possibly resulting in a much less desirable result for the user, working for hours only to realize later that things are busted). That's not to say we shouldn't investigate the root causes of said bugs, but this feels like a reasonable half-measure until we can find the real culprit.

- If we agree on that, then I suggest we start applying this philosophy more widely.

Completed checkin tasks:

- [ ] Ran the automated tests and linted the code

### DIFF
--- a/packages/haiku-serialization/src/bll/File.js
+++ b/packages/haiku-serialization/src/bll/File.js
@@ -69,10 +69,28 @@ class File extends BaseModel {
     return cb()
   }
 
+  assertContents (contents) {
+    if (typeof contents !== 'string') {
+      throw new Error(`Code was invalid ${this.getAbspath()}`)
+    }
+
+    // Returns truthy for "", " ", "   \n ", etc.
+    if (contents.match(/^\s*$/)) {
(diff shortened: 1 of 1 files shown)


tsortwehttam:
Closing in favor of a different one (forthcoming)

PR #227, opened 2018-01-22
Fixes Jan 22

tsortwehttam:
OK to merge.

Medium review.

Asana task links:

- https://app.asana.com/0/481696547044295/530409460017182

Changes in this PR:

- [x] Added basic harness and collateral for e2e test (will circle back and improve/move)
- [x] Fixed unrelated issue where we would make git commits for no changes
- [x] Fixed bad bug causing problems undoing changes to elements with filters
- [x] Added passing e2e test case for the fix
- [x] If we did a bytecode update and the bytecode -> AST -> code pipeline gives us an invalid value or a blank string, throw an error (crash the app) to prevent bad content getting written to disk which may have the effect of corrupting projects.

Notes for the reviewer:

- Any red flags on the "throw error if content string is invalid"?

Completed checkin tasks:


packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -510,6 +509,8 @@ export default class Plumbing extends StateObject {
       })
 
       this._isTornDown = true
+
+      if (cb) cb()
```
roperzh, line 513: _nit:_ when possible, I like to check by type when doing this kind of conditionals:

```js
if(typeof cb === 'function') cb()
```

This allows us to prevent errors if for some reason the value results to be truthy but not a function (which I have found to be pretty common)

package.json
```diff
@@ -27,7 +27,7 @@
     "push": "node ./scripts/push.js",
     "compile-all": "node ./scripts/compile-all.js",
     "distro": "node ./scripts/distro.js",
-    "test": "tape \"test/**/*.test.js\" | tap-spec"
+    "test": "yarn test-all && yarn compile-all && tape \"test/**/*.test.js\" | tap-spec"
```
roperzh, line 30: dumb question: why are we testing and then compiling? from the outside seems like the logic thing to do will be to compile -> test

roperzh (approved):
LGTM.

1. I don't see any problems with displaying an error message in this case, and sounds like it's way better to do that rather than have the code corrupted.
2. Testing works great!
3. I did a brief testing of the app and everything seems to be working fine for me.

stristr:
Code LGTM.

PR #229, opened 2018-01-22
fix: for crash on loading template projects Move and Moto.

stristr:
OK to merge.

Short review.

[Asana task](https://app.asana.com/0/480796620059175/532035242462772)

Changes in this PR:

- [x] Fix crash on loading fresh template projects Move and Moto.

Context: the files formerly known as `<rootDir>/interpreter.js`, `<rootDir>/embed.js`, and `<rootDir>/react-dom.js` aren't necessarily drop-in replacements for their replacements in `<rootDir>/code/main`. In particular, Move and Moto are trying to load `./bytecode.js` instead of `./code.js`, which leads to problems. There's no need to move these files, because they are created by `ProjectFolder` when they don't exist during project bootstrapping.

### DIFF
--- a/packages/haiku-plumbing/src/ProjectFolder.js
+++ b/packages/haiku-plumbing/src/ProjectFolder.js
@@ -238,13 +238,6 @@ export function buildProjectContent (_ignoredLegacyArg, projectPath, projectName
         fse.outputFileSync(dir(projectPath, nextFilePath), contentsToCopy)
         fse.removeSync(dir(projectPath, formerFilePath))
       }
-      // Now fix any legacy content that may be present inside of the updated file, e.g. references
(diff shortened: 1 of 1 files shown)


stristr:
Update: it looks like this was failing because of this code:

```
let fileContents = fse.readFileSync(dir(projectPath, nextFilePath)).toString()
fileContents.split('bytecode.js').join('code.js') // Respective to the code/main dir
fileContents.split('interpreter.js').join('dom.js') // Respective to the code/main dir
fse.outputFileSync(dir(projectPath, nextFilePath), fileContents)
```

Which, if you look closely, doesn't actually mutate fileContents.

Not sure how this crept in or wasn't noticed before, but the fix here should be durable.

tsortwehttam:
LGTM.

PR #231, opened 2018-01-23
fix: properly disable mouse interaction on timeline when scrubbing

roperzh:
Deferring decision to merge to @matthewtoast.

Short review.

Asana task links/Changes in this PR:

- [x] [Dragging on timeline sometimes drags properties element](https://app.asana.com/0/480796620059175/532681822311470)

- [x] [TImeline shouldn't vertical scroll while ticker is being dragged](https://app.asana.com/0/480796620059175/529942638926527)

Notes for the reviewer:

Setting/unsetting `pointer-events` and `user-select` on the timeline
rows is handled by a state conditional, which was being set
directly (`this.state.foo = 'bar'`) instead of triggering a render
via `setState`.

I will do the E2E checklist once I get the 👍 by Matthew, since this is likely to affect his performance changes.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/Timeline.js
+++ b/packages/haiku-timeline/src/components/Timeline.js
@@ -819,8 +819,10 @@ class Timeline extends React.Component {
           id='gauge-box'
           className='gauge-box'
           onMouseDown={(event) => {
-            this.state.doHandleMouseMovesInGauge = true
-            this.state.avoidTimelinePointerEvents = true
+            this.setState({
+              doHandleMouseMovesInGauge: true,
+              avoidTimelinePointerEvents: true
+            })
             this.mouseMoveListener(event)
           }}
           style={{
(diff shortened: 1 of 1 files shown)


packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -819,8 +819,10 @@ class Timeline extends React.Component {
           id='gauge-box'
           className='gauge-box'
           onMouseDown={(event) => {
-            this.state.doHandleMouseMovesInGauge = true
-            this.state.avoidTimelinePointerEvents = true
+            this.setState({
+              doHandleMouseMovesInGauge: true,
+              avoidTimelinePointerEvents: true
+            })
```
roperzh, line 825: @matthewtoast maybe an extra safe measure would be to do a check like this one:

```js
if (!this.state.doHandleMouseMovesInGauge) {
  this.setState({
    doHandleMouseMovesInGauge: true,
    avoidTimelinePointerEvents: true
  })
}
```

I don't feel like this is needed, since this callback shouldn't be called if `doHandleMouseMovesInGauge` is true, but I may be confused

tsortwehttam: I agree; I don't feel like that is needed.

tsortwehttam:
LGTM.

This looks good and should be good for perf. I overcorrected with that `this.state` call. But really we should only worry about that if doing it all the time e.g. on mouse move.

tsortwehttam (approved):

PR #232, opened 2018-01-23
Fix RWOD when moving keyframes

tsortwehttam:
OK to merge.

Short review.

Asana task link: https://app.asana.com/0/481696547044295/532573343524556

Changes in this PR:

- [x] Simplifies message parameters for `moveKeyframes` method
- [x] Fixes the "Red Wall of Death" (in which bytecode content gets replaced by a `{}` and then by the default empty-bytecode content) that occurs when moving keyframes.

Technical notes:

This turns out to be similar to a bug we saw (and fixed) ages ago: Node.js' `fs` read/write aren't mutually exclusive. This means you can read a file while it's being written to, write a file while it's being read, read as others read, write as others write, and so-on. Reading-while-writing may result in you getting an empty buffer back.

I found that `require` has a similar problem. It is possible to `require` from a file on disk while that file is being written to; it won't wait politely. (This is quite likely to happen in circumstances like ours in which we have async watchers who want to reload changes from disk despite changes that may want to be written to disk elsewhere.) It, in turn, `require(...)` will receive an empty buffer, in which case Node.js gives you a `{}`.

The logic I changed makes our `ModuleWrapper` only call `require(...)` when it knows there aren't any pending reads/writes, by hooking into the `File` class' read/write mutex. I also wrote an automated test that sort of fuzzes keyframe moves against a large-ish project, which creates these races. That test fails consistently without my changes, but passes consistently with them; so that tells me this is the fix.

To the reviewer:

- This could use some add'l team QA.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-bytecode/src/actions/moveKeyframes.js
+++ b/packages/haiku-bytecode/src/actions/moveKeyframes.js
@@ -8,7 +8,7 @@ var ensureTimelineProperty = require('./ensureTimelineProperty')
 // Really this method is best used in tandem with a method like 'moveSegmentEndpoints' which will
 // RETURN a keyframeMoves object that could be fed into this method!
 
-module.exports = function moveKeyframes (bytecode, componentId, timelineName, propertyName, keyframeMoves, frameInfo) {
+module.exports = function moveKeyframes (bytecode, componentId, timelineName, propertyName, keyframeMoves) {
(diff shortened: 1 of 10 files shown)


roperzh (approved):
LGTM, it's really nice to see tests!

PR #237, opened 2018-01-23
Fix copy of share link when on-stage element is selected

tsortwehttam:
OK to merge after @roperzh reviews.

Short review.

Asana task link: n/a

Changes in this PR:

- [x] If the clipboard cut/copy event originates from a local (Creator webview) DOM element, assume it is _not_ a candidate to be handled by Envoy. This fixes the issue observed by @nadonomy that copying the clipboard link while an element was selected on stage would result in the element bytecode being copied.

Notes for the reviewer:

- Any red flags on this @roperzh?

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -383,8 +383,16 @@ export default class Creator extends React.Component {
     })
 
     this.envoyClient.get(GLASS_CHANNEL).then((glassChannel) => {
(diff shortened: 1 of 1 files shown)


tsortwehttam:
After digging in a bit more, this seems low-risk, so I'm going to merge it since I'd like this to be in the next build. Easy to revert in any case.

Would still like Roberto's post-mortem LGTM though.

roperzh:
LGTM, sorry for the delay, was experiencing internet problems.

PR #238, opened 2018-01-23
Prevent crash when reopening project

tsortwehttam:
OK to merge.

Short review.

Fixes regression caused by #226 -- forgot to account for File instance being reused, meaning that a counter value I was using to determine whether to assert content on first load wasn't getting reset.

### DIFF
--- a/packages/haiku-serialization/src/bll/File.js
+++ b/packages/haiku-serialization/src/bll/File.js
@@ -61,8 +61,15 @@ class File extends BaseModel {
     })
 
     this._elementsCache = {}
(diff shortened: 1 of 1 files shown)


tsortwehttam:
To repro the issue as I saw it previous to this fix, just: Open Percy, go back to dash. Open Percy, go back to dash. Open Percy, it should crash.

stristr:
Makes sense, crash confirmed and also confirmed no longer present.

stristr:
(LGTM)

PR #239, opened 2018-01-24
Reinstate MasterProcess.js (Option 1)

tsortwehttam:
ONLY FOR ROLLING A PERF TESTING BUILD.

There are some bugs here that would need to be fixed before rolling out to anyone (for QA or otherwise), but this should work well enough to evaluate whether it fixes prod perf.

**Backstory**

Long ago, MasterProcess.js was hacked into existence out of the provisional belief that we needed to provide a separate process to run I/O-intensive operations (like Git) to avoid blocking important work in the UI thread.

But since the beginning, MasterProcess.js has been plagued with problems, and has been a main source of user-facing bugs and developer mobility-killers. Mysterious crashes, duplicate processes, zombie processes, silent failures resulting in hours of on-stage work being lost, slow updates, race conditions in writing to disk, debugging nightmares, etc.

Almost since its inception, I regretted implementing it, but I never refactored it away. Part of this regret was wishing that I had leveraged the obvious: the separation of processes that Electron provides out-of-the-box, Main and Renderer. For about a year I've believed with near-certainty that MasterProcess.js has existed for no reason other than to make development hell.

While working on multi-component, and nearly tearing my hair out debugging strange bugs and races related to MasterProcess.js, I decided to do an experiment with getting rid of it. This ended up way easier than I imagined. It made development much easier. And perf was still where it had been before; it even got better for undo/redo, copy/paste, etc., thanks to lopping off a websocket roundtrip.

**Tragedy**

Until we began rolling builds and seeing perf mysteriously crater, despite making many tangible improvements to perf when run in dev.

After [research](https://github.com/electron/electron/issues/3363#issuecomment-154905555), it seems that Main and Renderer, despite being separate processes, are effectively bound together by synchronous messages (internal to Chromium's architecture), meaning that they can in fact block one another in some cases.

We never saw this in dev due to a facet of our dev setup that has existed since the beginning. In dev, the Electron runner is spawned by a Node.js process, which gives us the isolation.

In dev, the process architecture looks like this:

```
Plumbing (Node process)
   Master
Electron (Node process)
  Main (Chromium process)
  Renderer (Chromium process)
    Creator, Timeline, Glass
```

In the prod .app, it looks like this:

```
Electron (Node process)
  Main (Chromium process)
    Plumbing
      Master
  Renderer (Chromium process)
    Creator, Timeline, Glass
```

**Next steps**

Option 1 to resolve this problem (this PR) is to reinstate MasterProcess.js. Thankfully, the majority of this code didn't take long. When I refactored it initially, I left most of the legacy code intact.

Option 2, fairly appealing, is that there may be an even simpler and faster way. In short, we may be able to have Electron launch Plumbing as a process.

Option 3, would assume that my hypothesis on this is completely wrong and we can get away with some other lower-impact solution.


tsortwehttam:
@stristr and I verified that this solves the primary perf issue in the distro build.

https://s3.amazonaws.com/haiku-electron-releases-production/releases/production/master/mac/8483187485580/2.3.84/Haiku-2.3.84-mac-pending.zip

Going to quickly investigate viability of Option 2, which would leave code mostly untouched. If that doesn't work we'll circle back to getting this ready for an actual rollout.

roperzh:
Thanks for the detailed PR, that was a nice read.

tsortwehttam:
We have a strong lead on Option 2, but I just pushed changes to this branch that fix the known issues.

tsortwehttam:
Closing since we solved this in a different way (Option 2).

PR #241, opened 2018-01-25
Tweaks and fixes

tsortwehttam:
NOT READY TO MERGE.

Short review.

Asana task links:

- https://app.asana.com/0/481696547044295/530409460017176

Changes in this PR:

- [x] Add back logic to force set the time of on-stage instances after a hard reload. This both prevents automatic playback and ensures that the scrubber time is preserved across hard reloads, which are triggered by things like undo/redo.
- [x] Re-enabled `webSecurity` on the Creator electron.js boot-up file. Doesn't seem to cause an issue when doing a quick check, and is probably best to keep on anyway. Docs for that option are here: https://electronjs.org/docs/api/browser-window
- [x] During merge design, record the changes that were made and log them if any were made. I think this is pretty handy to leave on by default since this is an area where we have historically had a bunch of bugs.
- [x] Removed a bit of cruft.

Notes for the reviewer:

- @stristr IIRC this was removed during other perf changes; but I would guess this should be fine. Let me know if you think this could cause any bad perf regressions.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/electron.js
+++ b/packages/haiku-creator/src/electron.js
@@ -119,10 +119,7 @@ function createWindow () {
   browserWindow = new BrowserWindow({
     title: 'Haiku',
     show: false, // Don't show the window until we are ready-to-show (see below)
-    titleBarStyle: 'hidden-inset',
-    webPreferences: {
-      webSecurity: false
-    }
+    titleBarStyle: 'hidden-inset'
   })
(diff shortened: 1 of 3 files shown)


roperzh:
so far LGTM, I have tested it locally and seems to work fine, I haven't done the full QA matrix tho

PR #242, opened 2018-01-25
Release -> Master

stristr:
OK to merge.

Medium review.

Changes in this PR:

- [x] Rejigger `Haiku.app` bootstrapping so that Electron main spawns `HaikuHelper.js` in both build mode and dev mode.
- [x] Adds `yarn watch-all`, separating a concern that used to belong to `yarn start`. (**"Breaking" change:** the developer is expected to start watchers in a separate tab if they want watchers. I couldn't quite rig it so that the watchers always stop when the watchers are spawned from a spawned/forked process. Ultimately I want to revisit this, but I think it should ideally be in the context of adding support for live reload vs. perfecting existing tooling.)
- [x] Add `.env` support for convenient, time-saving dev overrides (e.g. `HAIKU_API=https://inkstonestaging.haiku.ai` 😄 )
- [x] Support debugging Plumbing on port `9221` and renderers on port `9222`.
- [x] Bring in a forked `nodegit` which, when asked to serve a prebuilt binary for `NODE_MODULE_VERSION` 57 on darwin, has been hacked to serve a prebuilt binary for `NODE_MODULE_VERSION` 54 (which [doesn't exist](https://nodejs.org/en/download/releases) outside of pre-built Electron). This helps `yarn install` finish in less than ten minutes, which is nice.
- [x] Don't break the build.
- [x] Don't break existing dev modes (`yarn start`, `yarn go`, etc.)—but see note above about how `yarn watch-all` is to be run in a separate tab.
- [x] Refresh `README.md`.
- [x] (unrelated) Have GitLab also implement multiple clone attempts; sometimes the upstream is not ready for clone, likely due to EFS issues.

Electron is now test runner for Plumbing tests and seems to work the same as before.

Completed checkin tasks:

- [ ] ~~Did the [E2E checklist](https://haiku.quip.com/dGqeAEVSWdmg)~~
- [ ] ~~Expanded the [QA checklist](https://haiku.quip.com/rqwsAPsCGxqT)~~
- [ ] ~~Wrote an automated test covering new functionality~~


.env.example
```diff
@@ -0,0 +1,2 @@
+HAIKU_API=http://localhost:8080/
+SHARE_URL=http://localhost:8182/share/
```
roperzh, line 2: yay to this! super helpful

roperzh:
LGTM, but I'd wait for someone more familiar with this to do another review.

index.js
```diff
@@ -1,19 +1,40 @@
...
+    haikuHelperArgs
+  );
+
+  haikuHelper.on('message', (data) => {
+    if (!data || typeof data !== 'object' || !data.message) {
+      return;
+    }
+
+    const {message, haiku} = data;
+    if (message === 'launchCreator') {
```
tsortwehttam, line 32: Occurred to me, haven't thought about much: Could this `launchCreator` message ever happen more than once?

packages/haiku-plumbing/HaikuHelper.js
```diff
@@ -31,22 +35,23 @@ function go () {
...
   if (flags.repl) {
-    startEmUp(plumbing, haiku, function (err, folder) {
+    startEmUp(plumbing, haiku, (_, folder) => {
       // A quick-and-dirty 'REPL' mainly for testing the plumbing, but open to all
-      var repl = new ReplBase()
-      var prompt = 'haiku'
-      var opts = { me: plumbing, folder: folder }
+      const repl = new ReplBase()
+      const prompt = 'haiku'
+      const opts = { me: plumbing, folder: folder }
```
tsortwehttam, line 45: I removed this vestigial REPL bit in another branch somewhere. We should get rid of it. (Not necessary for this PR tho.)

stristr: Oh, it actually works now, FWIW.

packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -1031,110 +1021,20 @@ Plumbing.prototype.spawnSubgroup = function (existingSpawnedSubprocs, haiku, cb)
...
-    logger.info(`[plumbing] proc ${name} got error`, error)
-  })
-  proc.on('message', (message) => {
-    logger.info(`[plumbing] proc ${name} got message`, message)
-  })
-  proc.on('request', (message) => {
-    logger.info(`[plumbing] proc ${name} got request`, message)
-  })
-
-  return cb(null, proc)
```
tsortwehttam, line 1137: This may be one of my all-time favorite chunks of red diff I've seen in this project.

tsortwehttam:
Code LGTM on the condition you remind all the folks about the breaking dev workflow change. 👍 

(Note: I have not reviewed locally.)

packages/haiku-plumbing/HaikuHelper.js
tsortwehttam: Ah, awesome. Well, in that case, leave it?

.env.example
stristr: 🙌

index.js
stristr: Thought about adding code to check, but I don't think it's necessary. It would have to be a result of `go()` getting called twice by `HaikuHelper`, since that's the only place that sets `haiku.mode = 'creator'`. I don't see any way this could happen under normal usage.

stristr:
Merging, I will update #engineering channel with deetz.

PR #245, opened 2018-01-25
Tentative: Show loader until editor view is actually ready

tsortwehttam:
OK to merge, but only if OK with @taylorpoe and @zackbrown. This is just tentative in case y'all like this idea.

Short review.

Asana task link: n/a, just an experiment

Changes in this PR:

- Show the "reticulating splines" loader until the editor is totally finished loading.
  - Good because: User cannot nav back when things are in a race-able state.
  - Good because: User doesn't see the "sausage being made" while windows load.
  - Bad because: User doesn't see the Haikuified timeline loader that Taylor built.

Feel free to just leave this open for later discussion if you feel so-so or negative about it.

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -21,6 +21,7 @@ import Timeline from './components/Timeline'
 import Toast from './components/notifications/Toast'
 import Tour from './components/Tour/Tour'
 import AutoUpdater from './components/AutoUpdater'
+import ProjectLoader from './components/ProjectLoader'
 import EnvoyClient from 'haiku-sdk-creator/lib/envoy/EnvoyClient'
 import { EXPORTER_CHANNEL, ExporterFormat } from 'haiku-sdk-creator/lib/exporter'
 // Note that `User` is imported below for type discovery
@@ -94,7 +95,11 @@ export default class Creator extends React.Component {
         shouldCheck: true,
         shouldRunOnBackground: true,
         shouldSkipOptIn: true
-      }
+      },
+      doShowBackToDashboardButton: false,
+      doShowProjectLoader: false,
+      launchingProject: false,
+      newProjectLoading: false
     }
 
     this.envoyOptions = {
@@ -556,7 +561,12 @@ export default class Creator extends React.Component {
   }
 
   setDashboardVisibility (dashboardVisible) {
-    this.setState({dashboardVisible})
+    this.setState({
+      dashboardVisible,
+      doShowProjectLoader: false,
+      newProjectLoading: false,
+      launchingProject: false
+    })
   }
 
   switchActiveNav (activeNav) {
@@ -666,6 +676,8 @@ export default class Creator extends React.Component {
               applicationImage,
               projectObject,
               projectName,
+              doShowProjectLoader: true,
+              doShowBackToDashboardButton: false,
               dashboardVisible: false
             }, () => {
               // Once the Timeline/Stage are being rendered, we await the point that their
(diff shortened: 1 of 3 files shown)


taylorpoe:
Fine by me 👍

zackbrown:
yes please!! if we can hide _all_ of the 'progressive loading' I would be really happy.

packages/haiku-creator/src/react/components/SideBar.js
```diff
@@ -106,13 +106,16 @@ class SideBar extends React.Component {
...
-          </button>
+          {(this.props.doShowBackToDashboardButton)
+             ? <button id='go-to-dashboard' key='dashboard' onClick={() => { this.goToDashboard() }}
+               style={[
+                 BTN_STYLES.btnIcon, BTN_STYLES.btnIconHover, BTN_STYLES.btnText,
+                 {width: 'auto', position: 'absolute', right: 6}
+               ]}>
+               <ChevronLeftMenuIconSVG />
+             </button>
+            : ''
```
roperzh, line 117: _nit - personal preference_: if we only need to render something based on a conditional, I'd rather use `&&` for clarity:

```
{(this.props.doShowBackToDashboardButton) &&
  <button ... />
}
```

packages/haiku-creator/src/react/Creator.js
```diff
@@ -1045,6 +1088,9 @@ export default class Creator extends React.Component {
             </SplitPane>
           </div>
         </div>
+        {(this.state.doShowProjectLoader)
+          ? <ProjectLoader />
+          : ''}
```
roperzh, line 1093: _nit - personal preference_: if we only need to render something based on a conditional, I'd rather use `&&` for clarity:

```
{(this.state.doShowProjectLoader) && <ProjectLoader /> }
```

roperzh (approved):
LGTM, tested locally and works fine, personally I think it adds a good feeling of robustness (weird how perception affect unrelated things)

packages/haiku-creator/src/react/components/SideBar.js
stristr: `<rant>`Argument against this React convention, just to be extra annoying: `a && b` evaluating to `b.valueOf()` for truthy `a` is a weird interpreted language quirk (JS, Ruby, Python, …) that always struck me as indecent. `&&` is a boolean operation and therefore should implicitly cast its operands to booleans.`</rant>`

tsortwehttam: 🙉  You'll only take away `foo = bar || baz` over my 💀

roperzh: well, that's a very interesting point, just for curiosity, do you feel annoyed by this pattern too?

```js
const foo = bar || 'baz'
```

roperzh: lol, we had the same thought

stristr: https://cl.ly/26401i0k0k3c

Conclusion: PHP >> (JS + Ruby + Python)?

roperzh: lol at the semicolon comment 😂 

iirc, the `||` pattern is idiomatic Ruby, so ¯\_(ツ)_/¯

tsortwehttam: Another Ruby fave of mine (kind of neither here nor there) that I miss is `foo ||= bar`

zackbrown: Perhaps it's the rough-and-tumble nature of falsiness/truthiness that bothers Sasha?  There's room for the `const foo = bar || 'baz'` pattern even in languages that are strict about their booleans.  Like the ol' [null coalesce operator](https://en.wikipedia.org/wiki/Null_coalescing_operator)

stristr: `<rant>`Well, it's the muddying of the semantics of "or/and", which are not just binary operators, but _boolean_ operators that I expect to implicitly evaluate their operands as boolean-cast. I can get behind fuzzy `true`-ness when I'm not in a language where it would cause a compiler error…though I definitely prefer a language where these constructs are all compiler errors without explicit casting, a.k.a. languages where you have to shoot yourself in the foot with pointer arithmetic to mess up bools.`</rant>`

PR #247, opened 2018-01-25
tweak: we don't actually need forceSet or skipTransmit

stristr:
OK to merge.

Micro review.

Asana task link: n/a

Changes in this PR:

- [x] Minor cleanup of the method signature for `ActiveComponent#setTimelineTimeValue`.

Notes for the reviewer:

- I don't actually know what the Node 7.9.0 interpreter does with parameter defaults, but I'm sure it's slower than not having these functionally redundant parameters.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -397,7 +397,7 @@ export class Glass extends React.Component {
     seekMs = Math.round(seekMs)
 
     if (this.getActiveComponent()) {
-      this.getActiveComponent().setTimelineTimeValue(seekMs, true)
+      this.getActiveComponent().setTimelineTimeValue(seekMs)
     }
   }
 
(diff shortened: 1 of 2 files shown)


tsortwehttam:
Double double checking because, hmm, it makes me a bit nervous. Especially removing skipTransmit.

1.) Are you 1000% sure this doesn't cause a websocket-time-update-spiral-of-death during playback, scrubbing, etc.?

2.) Did you review to check that the playback issue doesn't still occur?

stristr:
So, as I understand it, this entire PR is a noop other than changing the method signature of `setTimelineTimeValue`.

The method was only called twice in the entire codebase, once during `hardReload` with parameters `(ms, true, true, true)` and once in Glass with parameters `(ms, true)`.

1. Since `skipTransmit` was always true, I omitted it as a parameter (default `false`) and just set it as `true` in the call to `Timeline.setCurrentTime`, which seems to be the only place where it was used.
2. Since `forceSeek` and `forceSet` were either always both `false` or both `true`, I just made it one parameter, default `false`.

Definitely merits another look since this is such core code, but I can't see how this actually changes any functionality we've already been testing.

tsortwehttam:
Got it, makes sense. Going to do a review just to be double sure once I get done with another thing.

tsortwehttam:
LGTM, merging.

PR #251, opened 2018-01-26
fix: for the mysterious disappearing polygon

stristr:
OK to merge.

Short review.

Asana task link: n/a

Changes in this PR:

- [x] Fix for late-breaking "after instantiating, deleting, and reinstating a polygon with gradient fill, it shows up as disappeared".

Notes for the reviewer:

This comes down to switching the order of `forceFlush()` and `setTimelineTimeValue()`. The super-arcane reason this matters is actually a bug in player that I don't think we need to fix right now, stemming from the behavior of a special cache [here](https://github.com/HaikuTeam/mono/blob/master/packages/%40haiku/player/src/renderers/dom/assignAttributes.ts#L49-L59) for `d` and `points` attributes. With the previous invocation order, if we tick (which is what happens inside `setTimelineTimeValue(ms, true)`) before performing a full flush render, we populate the cache on the first tick, then skip setting `points`/`d` on full flush and detect no changes on subsequent patches. This is fundamentally broken but not critical to fix right now.

Note: gradient fill was 100% a red herring. This bug exists for all `<polygon points="..." />`- and `<path d="..." />`-type constructs.

Completed checkin tasks:

- [x] Expanded/modified the [QA checklist](https://haiku.quip.com/rqwsAPsCGxqT)
- [ ] ~~Wrote an automated test covering new functionality~~ (literally don't know where to start 🙃 )

### DIFF
--- a/packages/@haiku/player/src/renderers/dom/assignStyle.ts
+++ b/packages/@haiku/player/src/renderers/dom/assignStyle.ts
@@ -7,10 +7,6 @@ export default function assignStyle(domElement, style, component, isPatchOperati
     domElement.__haikuExplicitStyles = {};
   }
 
-  // if (domElement.haiku.element.attributes['haiku-id'] === '1fe03cab1105') {
-  //   console.log(style);
-  // }
-
   if (!isPatchOperation) {
     // If we have an element from a previous run, remove any old styles that aren't part of the new one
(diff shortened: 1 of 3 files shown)


stristr:
Merging to get out le build.

PR #252, opened 2018-01-26
feat: Provide comprehensive group composition support in Lottie.

stristr:
OK to merge.

Short review?

[Asana task](https://app.asana.com/0/480796620059175/515376246146331/f) [continued]

Changes in this PR:

- [x] Replaces the Lottie group composition hack (which only supported translation) from 17408d7b with a proper, matrix-based approach.
- [x] Fixes vestigal rectangle layout issues encountered when group parents with translation matrices meet `<rect />` elements with `x` and `y` attributes.

Notes for the reviewer:

This issue started with our inability to support weird Sketch artifacts like these:

```
<svg>
  <g [translation.x]=60 [translation.y]=90>
    <g [translation.x]=-60 [translation.y]=-90>
      <rect ... />
    </g>
  </g>
</svg>
```

due to the fact that Lottie doesn't have groups (only precomps, which are different), and really needs you to give it this:

```
<svg>
  <rect [translation.x]=0 [translation.y]=0 ... />
</svg>
```

As I dug into some user-submitted Lottie files that _weren't_ fixed by 17408d7b, I realized there was no choice but to actually hack into Player's layout logic and compose group layout transformations correctly. To achieve this, I ripped out the part of Player's `parseCssTransformString` that is responsible for taking a thing that looks like our layout spec and turning it into things that look like our timeline properties, and exported `LAYOUT_3D_VANITIES` from `vanities` to avoid duplication of effort. I intentionally did all this without refactoring (other than moving the last part of `parseCssTransformString` into its own method) to avoid side effects.

This clears up longstanding Lottie layout issues that we've had several user reports about. Will be interesting to see if @nadonomy's Facebook emojis (minus masks, still not supported) lay out correctly when this change is in.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/player/src/helpers/composedTransformsToTimelineProperties.ts
+++ b/packages/@haiku/player/src/helpers/composedTransformsToTimelineProperties.ts
@@ -1,0 +1,58 @@
+import Layout3D from '../Layout3D';
+import mat4Decompose, {DecomposedMat4} from '../vendor/mat4-decompose';
+import math3d from '../vendor/math3d';
+import MathUtils from './MathUtils';
+
+export default function composedTransformsToTimelineProperties(out, matrices) {
+  // Note the array reversal - to combine matrices we go in the opposite of the transform sequence
+  // I.e. if we transform A->B->C, the multiplication order should be CxBxA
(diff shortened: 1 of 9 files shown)


tsortwehttam:
Code LGTM.

PR #266, opened 2018-01-30
Meow

tsortwehttam:


packages/@haiku/sdk-inkstone/src/index.ts
```diff
@@ -59,8 +59,8 @@ export namespace inkstone {
   }
 
   const inkstoneConfig: InkstoneConfig = {
-    baseUrl: 'https://inkstone.haiku.ai/',
-    baseShareUrl: 'https://share.haiku.ai/',
+    baseUrl: process.env.HAIKU_API || 'https://inkstone.haiku.ai/',
```
zackbrown, line 62: since sdk-inkstone can run in non-node settings (e.g. it's a dep of both the share page and www.haiku.ai) we can't assume `process` will exist here.  it _might_ work to gate behind a `typeof process !== 'undefined'` check first, though?

tsortwehttam:
Going to close in favor of the other PR, but IMHO allowing process.env to set these configs is a nice practice (presuming one is organized about the naming).

PR #268, opened 2018-01-30
infra: add script for syndication with CDN invalidation

stristr:
OK to merge.

Short review.

[Asana task](https://app.asana.com/0/480796620059175/515878783187361/f)

Changes in this PR:

- [x] Add a cache-busting syndication script (no more logging into S3!)
- [x] Simplify distro upload (we don't need to upload the same file stream to S3 twice—and then we don't need to check if their checksums match when we do auto-distro). It suffices to copy the version-specific `-pending.zip` file to the places it needs to go.

Notes for the reviewer:

- Worth reading through carefully to look out for any gotchas.

Completed checkin tasks:

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -30,6 +30,7 @@ module.exports = {
       region: 'us-east-1',
       bucket: 'haiku-electron-releases-production',
       user: 'haiku-electron-releases-writer-2',
+      distribution: 'E29RYBWU7AU9',
       key: process.env.HAIKU_RELEASE_WRITER_KEY,
       secret: process.env.HAIKU_RELEASE_WRITER_SECRET
     },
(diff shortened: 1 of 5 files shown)


tsortwehttam:
LGTM - but, one small comment:

What would "rolling back a release" look like here? 
For example, when I _manually_ syndicate a release, I usually take `*-latest.zip` and rename it to `*-{previousReleaseVersion}.zip` or`*-previous.zip`, so that I have something to roll back to in a hurry, just in case there's a critical last-minute issue. Should we account for that workflow here? Otherwise, this looks great to me.

stristr:
Hmmm, would you be happy to simply do `*-previous.zip`? The one piece of that we can't readily accommodate is knowing which version it was…since we don't know without squirrel logic.

tsortwehttam:
That'd be a great little addition to this if not too hard to add.

tsortwehttam:
LGTM.

PR #273, opened 2018-01-31
fix: smarter compound path handling for Lottie.

stristr:
OK to merge.

Short review.

[Asana task](https://app.asana.com/0/506410768732346/506410768732337/f)

Changes in this PR:

- [x] Instead of using `fill-rule` as a random guess about how we should decompose compound paths, actually use some #math to make a better (but still slightly random) guess. At its core we use the [ray casting algorithm](https://en.wikipedia.org/wiki/Point_in_polygon#Ray_casting_algorithm) to determine if we think the next composed path we encounter is supposed to be inside or outside the "enclosure" formed by the last path we processed, treating it as a polygon instead of a path with curves. As usual, this is incomplete, but much better than what we were doing before, and seems to allow us to cover a good amount of text ligatures cast to shapes we've seen in the wild.

Notes for the reviewer:

- Still much more to do here but "unit tests pass".

Completed checkin tasks:

### DIFF
--- a/packages/haiku-common/src/types/index.ts
+++ b/packages/haiku-common/src/types/index.ts
@@ -9,6 +9,23 @@ export type ContextualSize = {
   z: number;
 };
 
+export type CubicCurve = {
+  type: string;
+  x1: number;
+  y1: number;
+  x2: number;
+  y2: number;
+  x: number;
+  y: number;
+};
+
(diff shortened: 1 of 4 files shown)


stristr:
Merging so we can get this in next release.

PR #275, opened 2018-01-31
Improve Async Issues

tsortwehttam:
Probably ready to merge.

Needs careful review and testing.

- Instead of trying to catch "not opened" and "EPIPE" errors and assuming the stream closed because the websocket clients on the webviews were innocuously closed, properly destroy those client websockets when the user navigates back to the dashboard. And allow websocket critical errors (including timeouts) to crash the app, since the above change means that these only happen if something went really wrong.

- Decouple heavy disk writes and, importantly, AST updates, from the logic that updates the in-memory bytecode. Put these updates into a separate queue. This should make it at least significantly less likely that the user will navigate back to the dashboard while changes are still pending, which leads to confusing delays and crashes. It is also generally a good idea since, with projects that may have large paths and bitmaps, even small changes can lead to a huge amount of traffic waiting to go to disk.

Some tests are failing after these changes (as expected) so there is a bit more work to be done. But I could use any help reviewing and improving this until it's mergeable, as I think this is going to be important for the viability of the app.


packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -260,12 +260,6 @@ export default class Plumbing extends StateObject {
             logger.info(`[plumbing] websocket closed (${type} ${alias})`)
             this.removeWebsocketClient(websocket)
           })
-
-          // Allegedly this prevents uncatchable EPIPE crashes
-          websocket.on('error', (err) => {
-            logger.info(`[plumbing] websocket error (${type} ${alias})`, err)
-            this.removeWebsocketClient(websocket)
-          })
```
roperzh: are we sure we don't want to handle _any_ websocket errors?

roperzh (commented):
So far LGTM, I can take it for a spin if you need extra help testing.

packages/haiku-plumbing/src/Plumbing.js
tsortwehttam: For a bit of context, I've never actually seen _this_ handler fire, even when the websocket connection closes. Instead we see a `not opened` or an `EPIPE` error thrown on `.send()`. So, despite what the internet tells me about this handler, I don't know how useful it really is. But, to your point:

What I'm asking myself about this is: Which errors _can happen_, and which of those _can we handle i.e. recover from_? Since we don't have an answer to the first question (yet), we can't confidently answer the latter question. Crashing at least has the benefit of: Telling us what errors can happen, telling us what circumstances lead to those errors, and preventing a worse situation when the user believes things are working when they aren't.

So, I think as we develop an understanding of the errors we _actually_ get, we can start to handle them in a more surgical way rather than what we had before, which was effectively using errors generically for program flow control.

tsortwehttam:
Tests are passing now and I've done a fair amount of local testing. I've gone through the E2E checklist and not seen any weird issues. The main types of issues to look for here are:

1. Changes from Sketch seen in the diff log getting "done and then undone"
2. Changes made on stage not reflecting on share page after published
3. Elements on stage looking different after reopening project
4. Undo/redo working even worse than before

More or less, any consequence that would arise out of "not blocking receipt of incoming changes on flushing previous changes to disk." But I believe I have this set up correctly, since the in-mem bytecode receives in order, and AST/disk only occur when needed.

packages/haiku-plumbing/src/Plumbing.js
stristr: I think a significant improvement would be to log the error, then throw a new one to force crash.

That way we'll get a specific marker for this situation and maybe even a little more context.

packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -462,24 +462,8 @@ export default class Plumbing extends StateObject {
...
-            }
-          }
-        })
-      } else {
-        logger.info(`[plumbing] websocket readyState was not open so we did not send message ${message.method || message.id}`)
-        callback() // Should this return an error or remain silent?
-      }
-    })
+    const data = JSON.stringify(message)
+    return websocket.send(data)
```
stristr: Another case where I suspect it would be more helpful to fail with context.

I would prefer:

```
if (websocket.readyState === WebSocket.OPEN) {
  …
} else {
  callback(new Error('[plumbing] websocket readyState was not OPEN'))
}
```

packages/haiku-serialization/src/bll/File.js
```diff
@@ -26,6 +26,8 @@ const DEFAULT_TIMELINE_NAME = 'Default'
 const DEFAULT_TIMELINE_TIME = 0
 const HAIKU_ID_ATTRIBUTE = 'haiku-id'
 const DEFAULT_CONTEXT_SIZE = { width: 550, height: 400 }
+const DISK_FLUSH_TIMEOUT = 32
+const DISK_FLUSH_INTERVAL = 500
```
stristr: This makes me overall nervous. I don't think it would make things _worse_, but as a variant of "hope `setTimeout()` prevents races" it might actually fix problems if we add some `clearTimeout()`s to the mix. Here's what I was thinking in broad strokes:

 - Keep a stack of our `setTimeout()`s. e.g. `this._flushes.push(setTimeout(…))`
 - Every time we're about to push a new `setTimeout()`, first clear the timeouts for everything on our queue:

```
while (this._flushes.length > 0) {
  clearTimeout(this._flushes.shift());
}
this._flushes.push(setTimeout(…))
```

packages/haiku-serialization/src/bll/Project.js
```diff
@@ -967,14 +967,14 @@ function getCodeJs (haikuId, haikuComponentName) {
...
-          'haiku-id': '${haikuId}',
-          'haiku-title': '${haikuComponentName}'
+          "haiku-id": "${haikuId}",
+          "haiku-title": "${haikuComponentName}"
         },
         children: []
       }
-    }
+    };
   `.trim()
```
stristr, line 959: Guessing this was unintentional reformatting?

tsortwehttam: No, that's intentional - it's now inline with what our bytecode formatter produces automatically.

tsortwehttam:
Closing in favor of #283

PR #288, opened 2018-02-12
Hj2

tsortwehttam:
OK to merge.

Short review.

- [x] Improve logging; have ActiveComponent and Project use the logger instance so we see their output in the on-disk logs
- [x] Make some small player tweaks that fix/improve behavior when we have injected content; this will become important with control-flow and multi-component
- [x] Fix last "known" crash caused by instantiating something on stage then moving it; hashes being generated were not stable

Completed checkin tasks:


packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -1447,11 +1447,18 @@ function computeAndApplyTreeLayouts(tree, container, options, context) {
...
 function computeAndApplyNodeLayout(element, parent) {
-  if (parent) {
+  // No point proceeding if our parent element doesn't have a computed layout
+  if (parent && parent.layout && parent.layout.computed) {
     const parentSize = parent.layout.computed.size;
 
-    element.layout.computed =
-      Layout3D.computeLayout(element.layout, element.layout.matrix, parentSize);
+    // Don't assume the element has/needs a layout, for example, control-flow injectees
+    if (element.layout && element.layout.matrix) {
```
roperzh, line 1455: _just in case_ : should we check for `element` first? ```if (element && element.layout && element.layout.matrix) {```

packages/@haiku/core/src/properties/dom/vanities.ts
```diff
@@ -632,6 +632,15 @@ const TEXT_CONTENT_VANITIES = {
   },
 };
 
+const CONTENT_VANITIES = {
+  children: (_, element, value) => {
+    element.children = value;
+  },
+  insert: (_, element, value) => {
+    CONTENT_VANITIES.children(_, element, [value]);
```
roperzh, line 640: _nit_: feels off to use `_` for this variable here, as it's actually being used, and could be beneficial to know what it is.

packages/haiku-creator/src/entry.js
```diff
@@ -33,7 +33,7 @@ import {remote} from 'electron'
     // Give Raven some time to send a crash report before we close
     setTimeout(() => {
       remote.getCurrentWindow().close()
-    }, 500)
+    }, (process.env.NODE_ENV === 'production') ? 500 : 50000)
```
roperzh, line 36: _just in case_: what is the purpose of this? AFAIK, `Raven` won't send reports if the env is not production. Maybe the order of the second and third operands is wrong? (same for the other entry reporters)

roperzh (approved):
LGTM

packages/haiku-creator/src/entry.js
tsortwehttam: The timeout's purpose (in production) is to give Raven a chance to send the report. If we close/crash immediately, it doesn't send because the process exits.

The extra delay in dev is just some extra time padding for us, so that we can e.g. open up dev tools or take a look around before the app closes. There's probably a smarter way to do this.

roperzh: Sounds good to me! thanks for the explanation

tsortwehttam:
I dunno why the health check is failing (lint passes ok), but I'm going to merge this. Need to get this in for QA testing asap.

PR #293, opened 2018-02-12
fix: for various races affecting publish UI

stristr:
OK to merge.

Short review.

Purpose of changes:

- Fix various races encountered while testing publish UI leading to unstable Git sha and incorrect bytecode being published.

- [x] In `Master.js`, move `bumpSemverAppropriately` out of `saveProject` and into its own routine that gets called first. This ensures that that bytecode isn't updated with the new version _after_ a tag is created, which was happening with intermittent regularity on projects where bytecode flushes were slow and Lottie export was fast (e.g. bytecode with large images, painstakingly processed by Haiku.app but just skipped by Lottie export).
- [x] Ensure there are no pending bytecode flushes before continuing on to Lottie export. This ensures that bytecode metadata is actually written to disk before we try to start committing stuff, in case Lottie is very fast as in bullet above. _[STILL BROKEN: making a change on stage and quickly hitting publish (before at least 500ms have passed) results in false "no publish necessary" behavior. I figured this is okay to leave broken and ticket for later.]_
- [x] Commit changes *after* populating project content during save, vs. before.

Taken together, these changes should make it so the commit SHA we produce during `this._git.commitProjectIfChanged('Updated metadata', cb)` stays the same from after we click the **Publish** button to after the Publish modal is closed and bytecode mutations are possible within the app.

Regressions to look for:

- Anything related to publishing never actually finishing. Tested carefully on several projects but worth another look.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-plumbing/src/Master.js
+++ b/packages/haiku-plumbing/src/Master.js
@@ -791,6 +791,11 @@ export default class Master extends EventEmitter {
         })
       },
 
+      // Ensure we bump the semver before proceeding.
+      (cb) => {
+        return this._git.bumpSemverAppropriately(cb)
+      },
+
       // Populate the bytecode's metadata. This may be a no-op if the file has already been saved
       (cb) => {
         // Just in case this ran somehow before we initialized the project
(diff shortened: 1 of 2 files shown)


tsortwehttam:
LGTM, merging.

PR #300, opened 2018-02-15
feat: Add ability to duplicate a project.

stristr:
OK to merge.

Medium review.

Purpose of changes:

- Adds ability to duplicate projects inside the app ([Asana task](https://app.asana.com/0/539750334128224/558122386005711/f)).

- [x] Fork the existing "new project" modal to support creating an automatically-named new project. Try chain: `OriginalNameCopy` -> `OriginalName1` -> `OriginalName2` -> …
- [x] Change a common reason project creation fails (duplicate name) to notify the user before they start to create the project, not after (https://cl.ly/1k0w3P3c323n). @taylorpoe, please check design!
- [x] Implement naive project duplication in Plumbing with these rules: copy bytecode and designs, but change the filename of `<originalProjectName>.sketch`, as well as any pointers to it in bytecode, to resolve to `<newProjectName>.sketch`. This should preempt likely issues of confusion with users who have that Sketch file open at the time of duplication, although isn't going to do anything for other Sketch files with a non-default name. I thought this was a good compromise between doing nothing and a holistic solution, which might be overkill. Thoughts/concerns about the resultant UX are welcome!
- [x] Get one of the Plumbing tests to pass (but not its leaked handles 😢)….

Regressions to look for:

- Some light refactoring of project creation was performed, so in addition to testing the duplication functionality, you should also spend some time confirming everything is normal with project creation.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/components/ProjectBrowser.js
+++ b/packages/haiku-creator/src/react/components/ProjectBrowser.js
@@ -30,18 +30,20 @@ class ProjectBrowser extends React.Component {
     this.state = {
       username: null,
       error: null,
-      showNeedsSaveDialogue: false,
       projectsList: [],
       areProjectsLoading: true,
-      recordedNewProjectName: '',
       isPopoverOpen: false,
       showNewProjectModal: false,
+      showDuplicateProjectModal: false,
       showDeleteModal: false,
       recordedDelete: '',
+      recordedNewProjectName: '',
       projToDelete: '',
       projToDeleteIndex: null,
+      projToDuplicateIndex: null,
       confirmDeleteMatches: false,
-      atProjectMax: false
+      atProjectMax: false,
+      newProjectError: null
     }
   }
 
@@ -98,7 +100,8 @@ class ProjectBrowser extends React.Component {
 
   hideDeleteModal () {
     this.deleteInput.value = ''
-    this.setState({showDeleteModal: false, projToDelete: ''})
+    this.closeModals()
+    this.setState({projToDelete: ''})
   }
 
   handleDeleteInputKeyDown (e) {
@@ -178,7 +181,22 @@ class ProjectBrowser extends React.Component {
   }
 
   showNewProjectModal () {
-    this.setState({ showNewProjectModal: true })
+    this.setState({ showNewProjectModal: true, newProjectError: null })
+  }
+
+  showDuplicateProjectModal (projToDuplicateIndex) {
+    const duplicateNameBase = `${this.state.projectsList[projToDuplicateIndex].projectName}Copy`
+    let recordedNewProjectName = duplicateNameBase
+    let iteration = 1
+    while (this.state.projectsList.find((project) => project.projectName === recordedNewProjectName) !== undefined) {
+      recordedNewProjectName = `${duplicateNameBase}${iteration}`
+    }
+    this.setState({
+      projToDuplicateIndex,
+      recordedNewProjectName,
+      showDuplicateProjectModal: true,
+      newProjectError: null
+    })
   }
 
   projectsListElement () {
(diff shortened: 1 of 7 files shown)


packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -178,7 +181,22 @@ class ProjectBrowser extends React.Component {
...
+    this.setState({ showNewProjectModal: true, newProjectError: null })
+  }
+
+  showDuplicateProjectModal (projToDuplicateIndex) {
+    const duplicateNameBase = `${this.state.projectsList[projToDuplicateIndex].projectName}Copy`
+    let recordedNewProjectName = duplicateNameBase
+    let iteration = 1
+    while (this.state.projectsList.find((project) => project.projectName === recordedNewProjectName) !== undefined) {
+      recordedNewProjectName = `${duplicateNameBase}${iteration}`
+    }
```
roperzh: Shouldn't we increase `iteration` on every loop?

edit: yes, this is causing the app to hang if I duplicate a project three times.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -251,36 +271,71 @@ class ProjectBrowser extends React.Component {
...
-    const fixedValue = rawValue
+    const recordedNewProjectName = rawValue
       .replace(/\W+/g, '') // Non-alphanumeric characters not allowed
       .replace(/_/g, '') // Underscores are considered alphanumeric (?); strip them
       .slice(0, 32) // Keep the overall name length short
 
-    this.setState({recordedNewProjectName: fixedValue})
+    // Check for any projects with the exact same name.
+    const equivalentNameMatcher = new RegExp(`^${recordedNewProjectName}$`, 'i')
+    const newProjectError = this.state.projectsList.find((project) => equivalentNameMatcher.test(project.projectName))
```
roperzh: nit: Feels like we are doing roughly the same here and on line 191, is there an elegant way to abstract this logic?

packages/haiku-creator/src/react/components/ProjectLoader.js
```diff
@@ -120,7 +120,7 @@ class ProjectLoader extends React.Component {
       <div style={STYLES.fullScreenCenterWrap}>
         <div style={STYLES.contentHolster}>
           <div style={STYLES.reticulator}>
-            {reticulations[this.state.retic]}
+            {reticulations[this.state.retic]}…
```
roperzh, line 123: :D

roperzh (changes requested):
Overall LGTM, only requested change is a bug where the app hangs when you duplicate a project three times.

edit: another thing that would be nice to have if _not too complicated_ (we should confirm with @taylorpoe) is to optimistically display the project browser with the duplicated project in there (maybe with the loader spinner only on that project) instead of freezing the entire screen.

packages/haiku-creator/src/react/components/ProjectBrowser.js
stristr: Yep, that was the unfulfilled plan! (Side note: when we switch to Airbnb, linter will complain about using `let` when `const` will do here, which would also catch this.) Fixing.

packages/haiku-creator/src/react/components/ProjectBrowser.js
stristr: Yeah, good point—this check is actually stronger and more comprehensive (the one in L191 wasn't doing case-insensitive). Fixed.

stristr:
Changes made, merging.

stristr:
(actually, please approve when ready)

roperzh (approved):
LGTM!

PR #302, opened 2018-02-15
V0 proxy support

roperzh:
OK to merge.

Medium review.

- [x] Detect if the user is using proxies, and if so display a message instead of crashing the app: https://app.asana.com/0/539750334128224/548822693280685
- [x] Use `loopback.haiku.ai` as our new host for websocket communications

Regressions to look for:

- Should we expect any kind of regression due to our host change?

Notes for reviewers:

If you want to emulate the crash / test this, you will need to configure your network to use a proxy, you can either fetch a free proxy from the internet or set up your local proxy. I have been using [squidman](http://squidman.net/squidman/) for the latter with great success.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-common/src/proxies/index.ts
+++ b/packages/haiku-common/src/proxies/index.ts
@@ -11,5 +11,5 @@ export const enum ProxyType {
  * Canonical method to detect if a proxy string provided by electron
  */
 export const isProxied = (proxyString: string) => {
-  return proxyString !== ProxyType.Direct
+  return proxyString !== ProxyType.Direct;
 };
--- a/packages/haiku-common/src/proxies/index.ts
+++ b/packages/haiku-common/src/proxies/index.ts
@@ -11,5 +11,5 @@ export const enum ProxyType {
  * Canonical method to detect if a proxy string provided by electron
  */
 export const isProxied = (proxyString: string) => {
-  return proxyString === ProxyType.Direct
+  return proxyString !== ProxyType.Direct
 };
--- a/packages/haiku-common/src/proxies/index.ts
+++ b/packages/haiku-common/src/proxies/index.ts
@@ -1,0 +1,15 @@
+/**
+ * Proxy Type.
+ * @enum {string}
+ */
+export const enum ProxyType {
+  Proxyed = 'PROXY',
+  Direct = 'DIRECT',
+}
+
+/**
+ * Canonical method to detect if a proxy string provided by electron
+ */
+export const isProxied = (proxyString: string) => {
+  return proxyString === ProxyType.Direct
+};
--- a/packages/haiku-common/src/proxies/index.ts
+++ b/packages/haiku-common/src/proxies/index.ts
@@ -3,7 +3,7 @@
  * @enum {string}
  */
 export const enum ProxyType {
-  Proxyed = 'PROXY',
+  Proxied = 'PROXY',
   Direct = 'DIRECT',
 }
 
--- a/packages/haiku-creator/src/electron.js
+++ b/packages/haiku-creator/src/electron.js
@@ -1,9 +1,10 @@
-import { BrowserWindow, app, ipcMain, systemPreferences } from 'electron'
+import { BrowserWindow, app, ipcMain, systemPreferences, session } from 'electron'
 import EventEmitter from 'events'
 import path from 'path'
 import { inherits } from 'util'
 import mixpanel from 'haiku-serialization/src/utils/Mixpanel'
 import TopMenu from './TopMenu'
+import { isProxied } from 'haiku-common/lib/proxies'
 
 if (!app) {
   throw new Error('You can only run electron.js from an electron process')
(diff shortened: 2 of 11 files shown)


packages/haiku-common/src/proxies/index.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Proxy Type.
+ * @enum {string}
+ */
+export const enum ProxyType {
+  Proxyed = 'PROXY',
```
stristr: Nit (since it's the name of a enum value): `Proxied`

packages/haiku-creator/src/electron.js
```diff
@@ -47,7 +48,7 @@ if (!haiku.plumbing.url) {
     throw new Error(`Oops! You must define a HAIKU_PLUMBING_PORT env var!`)
   }
 
-  haiku.plumbing.url = `http://${global.process.env.HAIKU_PLUMBING_HOST || '0.0.0.0'}:${global.process.env.HAIKU_PLUMBING_PORT}/?token=${process.env.HAIKU_WS_SECURITY_TOKEN}`
+  haiku.plumbing.url = `http://${global.process.env.HAIKU_PLUMBING_HOST || 'loopback.haiku.ai'}:${global.process.env.HAIKU_PLUMBING_PORT}/?token=${process.env.HAIKU_WS_SECURITY_TOKEN}`
```
stristr: I would actually prefer to keep this default and instead set `HAIKU_PLUMBING_HOST` in the `_config.js.handlebars` in mono root so that we use loopback in releases.

Alternatively, set `HAIKU_PLUMBING_HOST` to `0.0.0.0` in `start.js`.

Doing neither of these things means you can't do local dev on mono without DNS, which requires interwebs.

packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -110,7 +110,7 @@ const WAIT_DELAY = 10 * 1000
 const HAIKU_DEFAULTS = {
   socket: {
     port: process.env.HAIKU_CONTROL_PORT,
-    host: process.env.HAIKU_CONTROL_HOST || '0.0.0.0'
+    host: process.env.HAIKU_CONTROL_HOST || 'loopback.haiku.ai'
```
stristr: (See notes above about not binding local dev to DNS here and below.)

scripts/start.js
```diff
@@ -178,7 +178,7 @@ function setup () {
   global.process.env.HAIKU_AUTOUPDATE_SERVER = 'http://localhost:3002'
 
   if (inputs.devChoice === 'everything') {
-    global.process.env.HAIKU_PLUMBING_URL = 'http://0.0.0.0:1024'
+    global.process.env.HAIKU_PLUMBING_URL = 'http://loopback.haiku.ai:1024'
```
stristr: Per comments above about how local dev shouldn't depend on DNS, let's nix this completely.

stristr (changes requested):
Overall LGTM and nice skillz to get this all to work together!

I am requesting changes because as far as I can tell the current implementation breaks local dev if you don't have an Internet connection to resolve DNS. If I'm incorrect about this, please let me know!

We could hack this with `/etc/hosts` and call it a day, but I would prefer not to depend on a hosts hack for offline dev to be possible.

packages/haiku-common/src/proxies/index.ts
roperzh: Damn, sorry about this one. Fixed.

packages/haiku-creator/src/electron.js
roperzh: > I would actually prefer to keep this default and instead set HAIKU_PLUMBING_HOST in the _config.js.handlebars in mono root so that we use loopback in releases.

👍 Fixed, thanks so much for noticing this.

roperzh:
Thanks so much for the review, all comments are fixed, so this is ready for another review.

Once we merge, I'd like to trigger a build to test, specially considering that `loopback.haiku.ai` is now only set for production.

_edit_: heads up, I rebased against master and thus force pushed this branch.

packages/haiku-plumbing/src/Plumbing.js
```diff
@@ -110,7 +110,7 @@ const WAIT_DELAY = 10 * 1000
 const HAIKU_DEFAULTS = {
   socket: {
     port: process.env.HAIKU_CONTROL_PORT,
-    host: process.env.HAIKU_CONTROL_HOST || 'loopback.haiku.ai'
+    host: process.env.HAIKU_CONTROL_HOST || '0.0.0.0'
```
stristr: It looks like this got renamed to `HAIKU_PLUMBING_HOST` in `_config.js.handlebars`, but not here…I don't see it anywhere else so we should probably just change it to var used elsewhere.

packages/haiku-sdk-creator/src/envoy/index.ts
```diff
@@ -32,7 +32,7 @@ export interface EnvoyOptions {
 }
 
 export const DEFAULT_ENVOY_OPTIONS: EnvoyOptions = {
-  host: 'loopback.haiku.ai',
+  host: '0.0.0.0',
```
stristr: Here and below (`DEFAULT_HOST` in `findOpenPort.ts`): I was reading through and it looks like this doesn't actually get passed by Plumbing in all the right places.

Easy fix should be to change this to `global.process.env.HAIKU_PLUMBING_HOST || '0.0.0.0'` like other things.

stristr (commented):
Sorry, caught 2-3 more minor things. Else LGTM!

packages/haiku-plumbing/src/Plumbing.js
roperzh: Sorry about this, fixed `HAIKU_CONTROL_HOST` and since we are here I also changed `HAIKU_CONTROL_PORT` for the same reasons.

packages/haiku-sdk-creator/src/envoy/index.ts
roperzh: Fixed.

roperzh:
New batch of requested changes in 4347c2d5f072835c7830d1bf3948db771e627eba, thanks so much for the feedback.

stristr (approved):
LGTM, let's fire it up!

PR #309, opened 2018-02-19
chore: upgrade electron.

stristr:
OK to merge to `development`.

Short review.

Purpose of changes:

- Upgrade to latest electron, allowing us to un-muck our electron-based test bootstrapping (future PR)!

### DIFF
--- a/packages/haiku-plumbing/package.json
+++ b/packages/haiku-plumbing/package.json
@@ -49,7 +49,7 @@
     "lodash.merge": "4.6.0",
     "lodash.throttle": "4.1.1",
     "moment": "2.18.1",
-    "nodegit": "HaikuTeam/nodegit.git#7a5c502b03d6fc92b6b450a9b1f9a4e0b7689c0f",
+    "nodegit": "0.20.3",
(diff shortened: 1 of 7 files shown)


packages/haiku-plumbing/package.json
```diff
@@ -49,7 +49,7 @@
     "lodash.merge": "4.6.0",
     "lodash.throttle": "4.1.1",
     "moment": "2.18.1",
-    "nodegit": "HaikuTeam/nodegit.git#7a5c502b03d6fc92b6b450a9b1f9a4e0b7689c0f",
```
tsortwehttam, line 52: Just want to confirm that this seems OK. Does this impact our tools at all?

tsortwehttam:
Had one question, but LGTM.

packages/haiku-plumbing/package.json
stristr: Yes, but in a good way: Electron 1.8.2 actually uses the same Node API version as we do, so we no longer need to use this ugly hack!

PR #310, opened 2018-02-20
Fix tests

stristr:
OK to merge.

Short review.

Purpose of changes:

- Restore sanctity to the Jenkins health check.

- [x] Fix tests that were flaking due to order-sensitivity. Note: I ignored the reasons order mattered and just made them run in an order that passes the same way the tests pass when run standalone.
- [x] Fix plumbing test suite to fully pass, and add TAP test report for plumbing tests.
- [x] Fix lint errors that crept in while nobody was reading health checks.
- [x] Revert to having node run tests instead of electron, now that we've upgraded electron.

Notes to the reviewer:
- Some minor actual code changes:
  - export `HAIKU_WS_SECURITY_TOKEN` so we can use it in tests. Apparently plumbing tests have been broken since we added ws security.
   - Modify `plumbing.teardown` to also call `this.teardownMaster` before calling `master.teardown`. These are similar names for two notionally different things, but both should be happening when we tear down plumbing, just in case.
   - Modify `chokidar` implementation to emit a `ready` event, which is a more stable way of knowing `chokidar` is ready than waiting a canned amount of time. Not using this yet except in tests, but slated for later.

Completed checkin tasks:


stristr:
:boom: Finally green health check! 💥

roperzh (approved):
🎉 yes! LGTM

PR #313, opened 2018-02-20
fix: for "keyframes should still listen to expressions while paused()"

stristr:
OK to merge.

Medium review—should check demos.

Purpose of changes:

- Expressions should still work when an animation is paused.

- [x] Alter the duplicative and slightly incorrect interplay of `HaikuTimeline#isActive` and `HaikuTimeline#isPlaying` by simply getting rid of the concept of `isActive()`.
- [x] Use `isPlaying` instead of `isActive` to decide if we should update the the timeline with the global clock time, but not to decide if we should consider the timeline "running".
- [x] Like with looping, never consider the timeline "finished" if it's being time controlled.

Regressions to look for:

- Anything playback weirdness in Glass after going through several loops in and out of preview mode, between various changes.

Notes for reviewers:

- Technically this is an API breaking change, but the API isn't documented yet, so I didn't think we needed to worry about this.

Completed checkin tasks:

- [x] ~~Expanded the [QA checklist](https://haiku.quip.com/rqwsAPsCGxqT)~~ (somebody already did this)

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -514,21 +514,6 @@ HaikuComponent.prototype.getDefaultTimeline = function getDefaultTimeline() {
   return timelines[DEFAULT_TIMELINE_NAME];
 };
 
-HaikuComponent.prototype.getActiveTimelines = function getActiveTimelines() {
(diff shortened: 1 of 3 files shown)


stristr:
@matthewtoast gave verbal LGTM, merging.

PR #316, opened 2018-02-22
feat: make all adapter bundles available on the CDN.

stristr:
OK to merge.

Short review.

Purpose of changes:

- Have some moar CDN bundles on the CDN.

- [x] Death to `browserify`! Use `rollup` for all CDN bundles (was already doing this or project files, so we know it's safe/works).
- [x] Move all `@haiku/core` bundling work to `@haiku/core` (`yarn bundle`). Add `react-dom` and `vue-dom` standalones as well.
- [x] Add utilities for copying files on S3, which is faster than uploading the same content twice to two different keys, and syndicate all the new standalones when we push a new version.

Completed checkin tasks:


tsortwehttam:
LGTM.

PR #323, opened 2018-02-23
Proposal: format expressions code to avoid weirdness

roperzh:
OK to merge after discussion & rebasing (I'm lazy enough to wait for approval before rebasing).

Short review.

Purpose of changes:

- In multiline-expression land, every time you open the expression editor the code is messed up due to how our pipeline treats bytecode. In actions land, we _temporally_ solved this by formatting the code with prettier everytime the editor is open. This change reuses that logic in expressions.

- [x] Move `HandlerManager` into `haiku-serialization`, we discussed this in the past, and maybe `utils` is still not the right place but at least is closer.
- [x] Abstract logic to prettify code into a class method of `HandlerManager` and use it to format expressions.

Regressions to look for:

- Expression multiline editing
- Actions editor formatting

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/utils/HandlerManager.js
+++ b/packages/haiku-serialization/src/utils/HandlerManager.js
@@ -1,5 +1,5 @@
-import prettier from 'prettier'
-import functionToRFO from '@haiku/core/lib/reflection/functionToRFO'
+const prettier = require('prettier')
+const functionToRFO = require('@haiku/core/lib/reflection/functionToRFO').default
 
 const ALPHABET = 'abcdefghijklmnopqrstuvwxyz'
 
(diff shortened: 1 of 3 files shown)


packages/haiku-serialization/src/utils/HandlerManager.js
```diff
@@ -15,6 +15,41 @@ class HandlerManager {
...
+        // Outdent lines
+        prettierHandlerBody = `${prettierHandlerBodyLines.map((s) => s.slice(outdentBy)).join('\n')}\n`
+        // If we somehow got nothing back, just restore the original body (e.g. for only comments).
+        if (prettierHandlerBody.length === 0) {
+          prettierHandlerBody = body
+        }
+      }
+    } catch (e) {
+      // noop. User likely was permitted to save invalid JS.
+      console.warn(`[glass] caught exception prettying handler body: ${e.toString()}`)
```
stristr, line 47: Should not still be a `[glass]` console warning. (I think `haiku-serialization` generally uses `logger` and doesn't have any `console.log` statements in it.)

packages/haiku-serialization/src/utils/HandlerManager.js
```diff
@@ -15,6 +15,41 @@ class HandlerManager {
...
 
+  /**
+   * Prettifies a piece of JS code
+   *
+   * @param {string} body
+   * @param {object} options
+   * @returns {string}
+   */
+  static prettifyHandlerBody (body, {outdentBy = 2}) {
+    let prettierHandlerBody = null
```
stristr, line 26: I think we should bail here if `body` is falsey, thoughts?

stristr (commented):
Overall LGTM/seems to work. I am not expert in the original weirdness that required this hack to exist in the first place, and it would be great to shed the webview-land dependency on `prettier`…especially because we have `prettier` running next to AST work when we flush to disk. But guessing that's harder to fix than this.

roperzh:
> I am not expert in the original weirdness that required this hack to exist in the first place

Yes, and the source of the bug is still a bit obscure for me too. I'm totally okay if we decide to not include this hack and find a proper solution, iirc @matthewtoast mentioned in the past that is not _really simple_

stristr:
I'd say let's leave it for now, and backlog an Asana ticket to actually fix. Re: waiting for approval before rebasing—technically we should only approve a rebase-able PR, since there is manual work involved in fixing conflicts.

roperzh:
> technically we should only approve a rebase-able PR

👍 understood, sorry for not being clear, with approval I'm referring to the green light to introduce this hack in the expressions editor too, since it's just a random idea of mine and it wasn't discussed in the past!

PR #326, opened 2018-02-24
Add ability to mark projects as "public to community" from publish UI

taylorpoe:
Do not merge. So far only the UI is created for this. We still need the following functionality (to be created by someone other than myself [that task here](https://app.asana.com/0/548868462189817/571204757867885)):
* Read if project has been set to 'public' previously. If not, then — when they open the publish UI, set the project to public. This way we can default to 'public' going forward (when this ui is in place) without backfilling older projects with a 'public' setting (seems invasive/undesired).
* When the setting in the UI is toggled, it needs to store the setting to the project, so that it can be remembered for next time. E.g. if I've set my project to private, next time I open the publish modal, it should still be private.

Short review.

[The UI Task](https://app.asana.com/0/548868462189817/567843326852816)
Give the user the ability to change whether or they'd like their project to be visible on their public profile (user profiles coming soon). This will also allow us, as admin, to scope out which public, user projects are worth curating onto our public gallery.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -9,6 +9,7 @@ const ENDPOINTS = {
   CHANGE_PASSWORD: 'v0/user/password',
   ORGANIZATION_LIST: 'v0/organization',
   PROJECT_LIST: 'v0/project',
+  PROJECT_UPDATE: 'v0/project',
   INVITE_PREFINERY_CHECK: 'v0/invite/check',
   INVITE_CHECK: 'v0/invite/:CODE',
   INVITE_CLAIM: 'v0/invite/claim',
@@ -16,6 +17,7 @@ const ENDPOINTS = {
   SNAPSHOT_SYNDICATED_BY_ID: 'v0/snapshot/:ID/syndicated',
   PROJECT_SNAPSHOT_BY_NAME_AND_SHA: 'v0/project/:NAME/snapshot/:SHA',
   PROJECT_GET_BY_NAME: 'v0/project/:NAME',
+  PROJECT_GET_BY_UNIQUE_ID: 'v0/project/:UNIQUE_ID',
   PROJECT_DELETE_BY_NAME: 'v0/project/:NAME',
   SUPPORT_UPLOAD_GET_PRESIGNED_URL: 'v0/support/upload/:UUID',
   UPDATES: 'v0/updates',
@@ -468,6 +470,7 @@ export namespace inkstone {
       GitRemoteUrl: string;
       GitRemoteName: string;
       GitRemoteArn: string;
+      IsPublic: boolean;
 
       // Current: GitLab-specific fields.
       RepositoryUrl: string;
@@ -488,6 +491,15 @@ export namespace inkstone {
       Name: string;
     }
 
+    export interface ProjectUpdateParams {
+      ID?: number;
+      UniqueId?: string;
+      // Name: string; // for renaming
+      // GitRemoteUrl: // for user-specified git remotes
+      MakePublic?: boolean;
+      MakePrivate?: boolean;
+    }
+
     export function create(authToken: string, params: ProjectCreateParams, cb: inkstone.Callback<Project>) {
       const options: requestLib.UrlOptions & requestLib.CoreOptions = {
         url: inkstoneConfig.baseUrl + ENDPOINTS.PROJECT_CREATE,
@@ -498,6 +510,26 @@ export namespace inkstone {
       };
 
       request.post(options, (err, httpResponse, body) => {
+        if (httpResponse && httpResponse.statusCode === 200) {
+          const project = body as Project;
+          cb(undefined, project, httpResponse);
+        } else {
+          cb(safeError(err), undefined, httpResponse);
+        }
+      });
+    }
+
+
+    export function update(authToken: string, params: ProjectUpdateParams, cb: inkstone.Callback<Project>) {
+      const options: requestLib.UrlOptions & requestLib.CoreOptions = {
+        url: inkstoneConfig.baseUrl + ENDPOINTS.PROJECT_CREATE,
+        headers: _.extend(baseHeaders, {
+          Authorization: `INKSTONE auth_token="${authToken}"`,
+        }),
+        json: params,
+      };
+
+      request.put(options, (err, httpResponse, body) => {
         if (httpResponse && httpResponse.statusCode === 200) {
           const project = body as Project;
           cb(undefined, project, httpResponse);
@@ -528,6 +560,25 @@ export namespace inkstone {
     export function getByName(authToken: string, name: string, cb: inkstone.Callback<ProjectAndCredentials>) {
       const options: requestLib.UrlOptions & requestLib.CoreOptions = {
         url: inkstoneConfig.baseUrl + ENDPOINTS.PROJECT_GET_BY_NAME.replace(':NAME', encodeURIComponent(name)),
+        headers: _.extend(baseHeaders, {
+          Authorization: `INKSTONE auth_token="${authToken}"`,
+        }),
+      };
+
+      request.get(options, (err, httpResponse, body) => {
+        if (httpResponse && httpResponse.statusCode === 200) {
+          const project = JSON.parse(body) as ProjectAndCredentials;
+          cb(undefined, project, httpResponse);
+        } else {
+          cb(safeError(err), undefined, httpResponse);
+        }
+      });
+    }
+
+    export function getByUniqueId(authToken: string, uniqueId: string, cb: inkstone.Callback<ProjectAndCredentials>) {
+      const options: requestLib.UrlOptions & requestLib.CoreOptions = {
+        url: inkstoneConfig.baseUrl +
+          ENDPOINTS.PROJECT_GET_BY_UNIQUE_ID.replace(':UNIQUE_ID', encodeURIComponent(uniqueId)),
         headers: _.extend(baseHeaders, {
           Authorization: `INKSTONE auth_token="${authToken}"`,
         }),
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -522,7 +522,7 @@ export namespace inkstone {
 
     export function update(authToken: string, params: ProjectUpdateParams, cb: inkstone.Callback<Project>) {
       const options: requestLib.UrlOptions & requestLib.CoreOptions = {
-        url: inkstoneConfig.baseUrl + ENDPOINTS.PROJECT_CREATE,
+        url: inkstoneConfig.baseUrl + ENDPOINTS.PROJECT_UPDATE,
         headers: _.extend(baseHeaders, {
           Authorization: `INKSTONE auth_token="${authToken}"`,
         }),
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -29,6 +29,7 @@ import { EXPORTER_CHANNEL, ExporterFormat } from 'haiku-sdk-creator/lib/exporter
 // Note that `User` is imported below for type discovery
 // (which works even inside JS with supported editors, using jsdoc type annotations)
 import { USER_CHANNEL, User } from 'haiku-sdk-creator/lib/bll/User' // eslint-disable-line no-unused-vars
+import { PROJECT_CHANNEL } from 'haiku-sdk-creator/lib/bll/Project' // eslint-disable-line no-unused-vars
 import { GLASS_CHANNEL } from 'haiku-sdk-creator/lib/glass'
 import { InteractionMode, isPreviewMode } from '@haiku/core/lib/helpers/interactionModes'
 import Palette from 'haiku-ui-common/lib/Palette'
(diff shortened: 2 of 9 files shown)


taylorpoe:
@zackbrown Please review text "Display on public profile." This felt better to me than what we'd quoted in task: "Submit to Haiku Community."

Please feel free to now assign [the part II of this task](https://app.asana.com/0/548868462189817/571204757867885) to someone else now that the UI is done.

taylorpoe:
@zackbrown Lemme know when you push yer commits here. I'm going to style the toggle just a twiddle bit more.

zackbrown:
@taylorpoe just pushed.  In case you dive into the disabled/enabled/known states of the toggle, just a heads-up:

there are some inherent subtleties to the logic now that it's hooked up (do we know yet if the public status has been toggled by the user before?  do we have data from across the network yet? ) esp. with the consideration that we should never show the user their project is "private" if it will in fact end up being public (such as during the load time while waiting for the initial publish.)  I added a "disabled" state to help not break expectations here.

taylorpoe:
@zackbrown Coo' so this is good to merge after I tweak the styles a bit? Just want to know so I'm not blocking its merging.

zackbrown:
Yes, should be good to merge.  I'll kick off an inkstone prod deploy now so you can use the new hook-ups without needing to run dev inkstone

taylorpoe:
Hmm... @zackbrown  after pulling your changes, starting app shows me as logged out, and upon logging in I hit this error ```2018-02-28T18:38:18.185Z - error:  TypeError: Cannot read property 'statusCode' of undefined
    at /Users/taylorpoe/code/haiku/mono/packages/haiku-plumbing/lib/Plumbing.js:986:25
    at Request._callback (/Users/taylorpoe/code/haiku/mono/packages/@haiku/sdk-inkstone/lib/index.js:99:21)```
Think this might be due to some of the inkstone stuff you've done here?

stristr:
The health check failure is a WTF…starting CI building for 16 hours, stuck on getting a test report for `haiku-websockets`. Probably transient, and will be fixed with next commit(s).

taylorpoe:
I believe this is now no longer WIP and ready for reviewing/merging.

zackbrown:
+1, this is r2g and would be great to get into the next build.  Since both @taylorpoe and I contributed code here, it'd be nice to get a review from @stristr or @matthewtoast

packages/@haiku/sdk-inkstone/src/index.ts
```diff
@@ -488,6 +491,15 @@ export namespace inkstone {
       Name: string;
     }
 
+    export interface ProjectUpdateParams {
+      ID?: number;
+      UniqueId?: string;
+      // Name: string; // for renaming
+      // GitRemoteUrl: // for user-specified git remotes
```
roperzh, line 498: Should we leave this here? I can hear @stristr's heart breaking apart 💔

packages/haiku-creator/src/react/Creator.js
```diff
@@ -427,6 +428,13 @@ export default class Creator extends React.Component {
       }
     )
 
+    this.envoyClient.get(PROJECT_CHANNEL).then(
+      (project) => {
+        this.setState({envoyProject: project})
+        // this.handleEnvoyProjectReady()
```
roperzh, line 434: Same as before (💔)

packages/haiku-ui-common/src/react/ShareModal/index.tsx
```diff
@@ -17,58 +19,106 @@ const STYLES = {
...
-  };
-
   static defaultProps = {
     projectUid: '',
     sha: '',
-  };
+  } as PropTypes;
 
-  constructor () {
+  constructor (props:PropTypes) {
```
stristr, line 57: Lint.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -1206,6 +1214,7 @@ export default class Creator extends React.Component {
                   <Stage
                     ref='stage'
                     folder={this.state.projectFolder}
+                    envoyProject={this.state.envoyProject}
```
roperzh, line 1217: extreme nit: would be great to include this in the `Stage` `propTypes`, I have been trying to always do it, and we should revisit components in general to include `propTypes` on all of them. 🥇.

This same comment applies for `StageTitleBar` and `ShareModal`

packages/haiku-ui-common/src/react/ShareModal/index.tsx
```diff
@@ -17,58 +19,106 @@ const STYLES = {
...
+      isPublicKnown: false,
     };
+    this.togglePublic = this.togglePublic.bind(this);
+    this.toggleTooltip = this.toggleTooltip.bind(this);
   }
 
-  componentWillReceiveProps({error, isSnapshotSaveInProgress}) {
-    if (error) {
-      this.error = error;
+  componentWillReceiveProps(nextProps:PropTypes) {
```
stristr, line 68: Lint.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
zackbrown: `yarn lint-all --fix` was a no-op

packages/@haiku/sdk-inkstone/src/index.ts
```diff
@@ -543,6 +575,25 @@ export namespace inkstone {
       });
     }
 
+    export function getByUniqueId(authToken: string, uniqueId: string, cb: inkstone.Callback<ProjectAndCredentials>) {
+      const options: requestLib.UrlOptions & requestLib.CoreOptions = {
```
stristr, line 579: I think this would be easier to maintain as the method body for this call:

```
// Note: name and unique ID can be used interchangeably to get a project from Inkstone.
return getByName(authToken, uniqueId, cb);
```

And we could of course change it, if that changes.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
zackbrown: `yarn lint-all --fix` was a no-op

stristr:
Overall LGTM, left some minor comments.

packages/haiku-ui-common/src/react/ShareModal/ProjectShareDetails.tsx
```diff
@@ -44,6 +104,8 @@ export class ProjectShareDetails extends React.PureComponent {
       linkAddress,
       isSnapshotSaveInProgress,
       onHide,
+      isPublic,
+      togglePublic,
```
roperzh, line 110: nit: Please include proptypes for this two.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
stristr: Oh, weird—it *should* want a space after the colon. Will check into why it's not asking for it, but would you mind adding it here and in the other spot?

packages/haiku-ui-common/src/react/ShareModal/index.tsx
```diff
@@ -17,58 +19,106 @@ const STYLES = {
...
     super();
 
     this.state = {
       showDetail: false,
+      isPublic: true, // default true so we don't ever accidentally tell the user their projects are more private than they are
+      showTooltip: false,
+      isPublicKnown: false,
     };
+    this.togglePublic = this.togglePublic.bind(this);
+    this.toggleTooltip = this.toggleTooltip.bind(this);
```
roperzh: nit: prefer arrow functions in JSX over `bind` (performance benefits noted by @stristr )

packages/@haiku/sdk-inkstone/src/index.ts
stristr: These are notes to future self, exempted from the rules about commenting out source code because they're intended to be used, someday.

roperzh:
LGTM!

packages/haiku-creator/src/react/Creator.js
zackbrown: same, intended as a breadcrumb for a future maintainer that this is 'the consistent way' to handle envoy lifecycle hooks

packages/@haiku/sdk-inkstone/src/index.ts
zackbrown: I disagree — this sort-of hack would increase the surface area for hard-to-track-down bugs in the future (mostly because it's inconsistent w/ anything else in sdk-inkstone.) 

The original goal of sdk-inkstone was to be auto-generated based on static analysis of inkstone.  I still see that as the platonic ideal of this particular module — think of this near-duplication of this code as a bit of "Go"ism that's infused by proximity.

packages/haiku-creator/src/react/Creator.js
zackbrown: I disagree — if we're going to be revisiting components to add types, they should be actual types, not propTypes. Similar refactor effort, much greater benefit.  Notice the refactor of ShareModal.

packages/@haiku/sdk-inkstone/src/index.ts
stristr: This is a single endpoint in inkstone, though: https://github.com/HaikuTeam/inkstone/blob/master/src/api/server.go#L94

It calls a private method `getProjectsByNameOrUniqueId`.

packages/haiku-creator/src/react/Creator.js
roperzh: I see, and makes sense. My only objection would be that if we are not including types right now, at least include propTypes that so at least we can use them as a reference when migrating to types instead of guessing. IMO it's a very low effort thing that helps when viewing components for first time.

packages/@haiku/sdk-inkstone/src/index.ts
stristr: If you'd like to improve the semantics, `getById` would be appropriate. I was trying to avoid suggesting that because it would require refactoring of existing working code. However, it looks like this `getByName` method is only used inside of `mono`, so you could fix that in this PR if you want.

If it were actually two separate endpoints in inkstone, I would agree.

packages/haiku-creator/src/react/Creator.js
roperzh: one last note: this is just an extreme nit and this discussion shouldn't block merge!

tsortwehttam: Since adding a propType is 1 line of code, I'm in favor of using them for any module for which migrating to TypeScript is probably months away. But like R said, it's not a blocker.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
zackbrown: Yeah I mind.  This should not be handled by a human, but by tooling (if at all.)  I'm already frustrated with how much (25%?) the lint rules slow down my coding — "is that red squiggle a syntax or type error?  or just another fucking spacing lint 'error'?  alright let's arrow key up there and add a space, arrow-key over there and add another space," compounded by the fact that [my IDE's auto-formatter](https://github.com/vvakame/typescript-formatter) — which uses TS's built-in formatter — apparently doesn't align with our linting rules.  Here's what happens after I auto-format AND run `yarn lint-all --fix` — which would defeat the whole point of linting if I did this on every file I touch, meaning I have to lint every line of code I write by hand.

image

While I believe *style* linting is a net detriment to our company — a separate conversation TBH — I will meet you far enough in the middle and run `yarn lint-all --fix`.  But if you care about that space (and happy to hear your reasoning) the linter needs to fix it, not a human.

packages/haiku-ui-common/src/react/ShareModal/ProjectShareDetails.tsx
taylorpoe: fixed.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
taylorpoe: +1 to having an is-linting-beneficial-in-light-of-friction-added conversation.

roperzh: Throwing my 2 cents: I like linting as a tool to find problematic patterns, not so to find _style_ issues.

IMO the best solution to styling is autoformatting of the code a la `gofmt` or `mix format` (Elixir ftw!), and I think that the equivalent "standard" in JS is `prettier`

packages/@haiku/sdk-inkstone/src/index.ts
zackbrown: I'll put that another way:  the expected amount of maintenance effort currently expected on this method is ~0 (if maintenance is required, it would most likely be done in aggregate on the whole module, since this method is just another cookie-cutter imprint of every other method in the file.)  So I see "easier to maintain" as a bit prematurely optimized.  

That said, you're right (and my previous comment about this reflecting a 1:1 mapping was wrong) that this single endpoint should have a single method call.  IIRC during implementation, there was some bug that I fixed by shelling out `getProjectsByNameOrUniqueId` from `LoadAuthorizedProject` – and then instead of having to make sure I wouldn't break anything by mutating `getByName` in sdk-inkstone, I simply added a new method.  I'd probably make the same call again.  

I would quantify this as $0.01 of technical debt that would require $0.10 of effort to pay down now.  And the opportunity cost of fixing it now dwarfs the expected interest on the tech debt (perhaps $0.05 of lifetime technical debt vs $2.00 of lifetime opportunity cost.)  This all make sense?  Maybe it's counter-intuitive or maybe my prowess as a perfectionist engineer has been forever marred by my experience as a product manager.  Probably sucks sometimes to have to deal with my code — but again, I'd make the same call again.  This passes my gut check as "OK to ship."  If you still strongly disagree let's chat offline.

PR #333, opened 2018-03-01
Timeline perf 1

stristr:
OK to merge.

Short review?

Purpose of changes:

- I discovered realized I was already in a totally unmergeable state relative to `ms` and getting worse the further I went, so made the hard call to bring in timeline perf work to `ms` and finish it there.
- I commented out a broken test (`'ActiveComponent.prototype.pasteThing[1]'`) so we can see obvious breakage in health check.


stristr:
The only really crucial stuff to read/understand is in [BaseModel.js](https://github.com/HaikuTeam/mono/pull/333/files#diff-072b591d5a64a45ee6c0828f7c1a3442). Rest, I think, can go through normal QA.

packages/haiku-serialization/src/bll/BaseModel.js
```diff
@@ -79,10 +72,58 @@ class BaseModel extends EventEmitter {
...
+    return () => {
+      delete this._updateReceivers[source]
+    }
+  }
+
+  notifyUpdateReceivers (what) {
+    Object.keys(this._updateReceivers).forEach((receiver) => {
+      this._updateReceivers[receiver](what)
+    })
+  }
```
stristr, line 122: Note—just moved these two methods above out of `Keyframe.js` and into `BaseModel.js`.

packages/haiku-serialization/src/bll/BaseModel.js
```diff
@@ -196,47 +227,36 @@ class BaseModel extends EventEmitter {
   }
 
   destroy () {
-    if (this.beforeDestroy) this.beforeDestroy()
     this.constructor.remove(this)
     this.constructor.clearCaches()
-    this.setPrimaryKey(undefined)
     this.__destroyed = Date.now()
-    if (this.afterDestroy) this.afterDestroy()
     return this
```
stristr, line 233: `beforeDestroy`/`afterDestroy` were unused. We can put them back when we need them.

packages/haiku-serialization/src/bll/BaseModel.js
```diff
@@ -302,68 +322,131 @@ BaseModel.extend = function extend (klass, opts) {
...
   klass.idx = (instance) => {
-    for (let i = 0; i < collection.length; i++) {
-      if (collection[i].sameAs(instance)) return i
+    for (let i = 0; i < arrayCollection.length; i++) {
+      if (arrayCollection[i] === instance) return i
     }
     return -1
   }
 
+  klass.setInstancePrimaryKey = (instance, primaryKey) => {
```
stristr, line 391: Important: this replaces the naive way we were changing keyframe uids before (`this.uid = 'blah'`). I changed for the benefit of the `hashmapCollection`, which is a major boost for `findById` and hence a huge timesaver for things like `Keyframe#next` etc.

tsortwehttam:
Code looks great (LGTM). Agree with the call to just bring into `ms`. Merging.

PR #336, opened 2018-03-02
Adds Intercom Support Widget to in-editor experience

taylorpoe:
OK to merge.

Adds a [help button in the app, that triggers Intercom Messenger](https://app.asana.com/0/548868462189817/553837299804739)

Completed checkin tasks:


packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -403,6 +406,11 @@ class Timeline extends React.Component {
               }
             }
           )
+          user.getUserDetails().then(
```
taylorpoe, line 408: In testing, this returned the json for me: {"Username":"taylor+taylor@haiku.ai","IsAdmin":true}, so when I use it on intercom I send over the `email` of this object with JSON.parse(user)['Username'].

My concern is that this misnomered 'Username' which is really an email address, is not _always_ an email address — would that ever happen?. I.e. if we've changed flip/flopped at all during the username/email/orgname ruckus'ing throughout the course of our company life.

taylorpoe:
A byproduct of this feature is the need to actually check Intercom. Currently, I check it about once a week. (I should probably be more active there.) With this feature in place, we'll all certainly need to be more active, and probably need _all the devs_ to check in. Otherwise, we hit the scenario: "Hey I need help w Lottie." Then me — getting to this message 17 hours later — *tags Sasha (on intercom)*. Sasha, sees intercom assigned message, never?

So maybe hand-in-hand with this is a subtask of pinging our slack whenever someone is assigned a new message in intercom? It'd be great it could actually ping the right person.

Edit: Made a task on asana for this discussion.

packages/haiku-timeline/src/components/IntercomWidget.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import Palette from 'haiku-ui-common/lib/Palette'
+import Radium from 'radium'
+import Intercom from 'react-intercom'
+import { IntercomAPI } from 'react-intercom'
```
roperzh: nit: this can be written as:

```js
import Intercom, { IntercomAPI } from 'react-intercom'
```

packages/haiku-timeline/src/components/Timeline.js
stristr: tl;dr: don't worry…yet!

I would recommend putting a TODO in a comment here, to remind us that if/when we allow users to choose usernames. Right now, all 9,330 users have email usernames.

packages/haiku-timeline/src/components/IntercomWidget.js
```diff
@@ -0,0 +1,68 @@
...
+
+class IntercomWidget extends React.Component {
+  render () {
+    if (!this.props.user) {
+      return <span />
+    }
+
+    const user = {
+      email: JSON.parse(this.props.user)['Username']
+    }
```
roperzh: nit: I'm not sure how frequently this component is going to be rendered, but could be a good performance enhancement to move this JSON parsing into the constructor:

```js
constructor (props) {
  super()
  this.user = {
    email: JSON.parse(props.user)['Username']
  }
}
```

stristr:
Good idea, and looks like this is definitely a thing: https://www.intercom.com/integrations/slack

packages/haiku-timeline/src/components/IntercomWidget.js
```diff
@@ -0,0 +1,68 @@
...
+          </g>
+        </svg>
```
roperzh, line 59: This could use an SVGO pass, [this tool](https://jakearchibald.github.io/svgomg/) can come in handy.

Note: once merged, let's move this into `ui-common`, I'm not requesting this as part of this PR because there will be conflicts and pains.

roperzh (approved):
LGTM

packages/haiku-timeline/src/components/IntercomWidget.js
taylorpoe: I tried this out. For some reason, Proptypes failed because of it. https://cl.ly/0q2g2I1r213R Because the user prop will only be set once by its parent component, it should only cause rendering the one time that it is needed afaik.

packages/haiku-timeline/src/components/IntercomWidget.js
taylorpoe: Kk down to move after merging. Pasting the markup into https://jakearchibald.github.io/svgomg/ didn't work as far as I could tell. Going to leave as is since is only 6 LOC. In general, my personal philosphy with svgs is if they're small and the component is also small AND the svg is not expected to be reused, it's okay to leave in the markup. That too archaic??

packages/haiku-timeline/src/components/IntercomWidget.js
taylorpoe: Changed.

packages/haiku-timeline/src/components/Timeline.js
taylorpoe: Kk whew. Added a note.

packages/haiku-timeline/src/components/IntercomWidget.js
stristr: `super(props)`?

stristr:
:shipit:

packages/haiku-timeline/src/components/Timeline.js
zackbrown: Agreed with @stristr that every account has a unique email address, and that will continue to be a solid choice for unique ID.  To be clear, the `[user].[username]` field in the database is where we store the email address, which is why you're seeing that field named as such, @taylorpoe [renaming database columns is ugly business].  But you're doing it right.

Re:
> if/when we allow users to choose usernames

we already do.  ~9300 of those 9330 users already have a unique short username too [they selected this when they signed up.]  

There are a small number (~30) legacy accounts that were created when we were "multiple-users-per-org, no-short-username." We can either a.) just delete the extraneous accounts [many of which were support accounts I created so I could access folks' projects,] or b.) manually give them short unique usernames, then this issue is solved for good.  For now, it's 99.7% solved, with a 99.999% chance of never being an issue (old/dead accounts.)

taylorpoe: Kk I was going to say — I remember the username option now being on the signup form. Makes sense how it's the short_username. Thanks for the detailed explanation. Down for whatever finagling needs to happen in order for me to **never** need to touch the db.

taylorpoe:
One more cool thing about this is that if we respond while the user has the app closed, next time they open it up, they see our reply.

PR #340, opened 2018-03-02
Missed a distro script hook to get changelog into build

tsortwehttam:
OK to merge.

Short review.


stristr:
So short I almost forgot to say LGTM

PR #341, opened 2018-03-02
fix: update the read paths for changelogs

roperzh:
OK to merge.

Short review.

### DIFF
--- a/packages/haiku-serialization/src/bll/Changelog.js
+++ b/packages/haiku-serialization/src/bll/Changelog.js
@@ -1,10 +1,12 @@
 const fs = require('fs')
 const path = require('path')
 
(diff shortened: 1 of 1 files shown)


tsortwehttam:
LGTM.

stristr:
This is probably not going to work with ASAR.

tsortwehttam:
Is it worth trying or should I just flag it to `asarUnpack` it?

roperzh:
@stristr do you know what we can do? current prod builds can't find the `changelogs/public/` folder

stristr:
Might be worth trying…`packages/haiku-serialization` and `node_modules/haiku-serialization` are at the same level.

Really the best thing to do would have been to make `haiku-changelog` a package, which we should backlog anyway. Then you could just `require.resolve()` and everything would work.

Let's give it a go.

PR #342, opened 2018-03-03
feat: add animated splash screen

zackbrown:
OK to merge.

Short review.

Purpose of changes:

- replaces white-on-white splash screen with a haiku, see: https://cl.ly/1k041H03292j


stristr:
Animation looks good, and bouncy!

Technical issue: since we haven't fixed fuzzy core version yet, it's best to avoid creating a `@haiku/core` version binding which will get out of sync with the one we ship from mono, creating multiple raf loops in Creator the next time we bump the semver. Here are two ways to work around this available now:

 - Manually push a new tag in your project with a fuzzy version of `@haiku/core` (`^3.1.10`, not `3.1.10`). Steps are: `haiku clone` project, bump the version in `package.json` with text editor, commit, add tag, and `git push --tags`. You'd have to follow these steps each time you make changes to the animation until we fix fuzzy core version.
 - Dump the bytecode only in `haiku-creator/src/react/bytecode-fixtures` and mount it manually, e.g. using the approach in the `<ProjectPreview />` React component in creator.

I think the "hack a tag" approach is best for now, especially because fuzzy versions needs to happen really soon. The second option requires writing much more code.

zackbrown:
Technical issue fixed — good to ship?  Let's keep an eye out for start-up judders & flickers on distro.

stristr:
LGTM!

PR #351, opened 2018-03-07
tweak: rename republish script and add support for project filter.

stristr:
OK to merge.

Short review.

Purpose of changes:

- Tweak republish script to allow targeting a comma-separated list of project names instead of blindly republishing all projects.

### DIFF
--- a/scripts/republish-projects.js
+++ b/scripts/republish-projects.js
@@ -1,4 +1,6 @@
 const { each } = require('async')
+const { argv } = require('yargs')
+
 const Plumbing = require('haiku-plumbing/lib/Plumbing').default
 const haikuInfo = require('haiku-plumbing/lib/haikuInfo').default
 const { ExporterFormat } = require('haiku-sdk-creator/lib/exporter')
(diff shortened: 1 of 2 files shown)


stristr:
Used this yesterday, just gonna merge it.

scripts/republish-projects.js
```diff
@@ -71,11 +73,15 @@ plumbing.launch({ ...haikuInfo(), mode: 'headless' }, (err) => {
...
     log.hat(`currently signed in as ${organizationName}`)
 
-    plumbing.listProjects((err, projects) => {
+    plumbing.listProjects((err, allProjects) => {
       if (err) {
         throw err
       }
 
+      const projects = argv.onlyProjects
+        ? allProjects.filter(({ projectName }) => argv.onlyProjects.split(',').includes(projectName))
```
roperzh, line 82: nit: there's also [String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes), so no need to split!

roperzh:
redundant but: LGTM!

scripts/republish-projects.js
stristr: Problematic though.

Just apply that logic to project list `['FooBar', 'Foo', 'Bar']` with filter `--onlyProjects=FooBar` and you'll see why :smile:

roperzh: w00t! you are right :D

PR #354, opened 2018-03-08
hf-mar7

tsortwehttam:
`08f0bb8` is an important fix we should merge in any event (unrelated to today's issue)

`983de02` is theory 1

`2cc5816` is theory 2


stristr:
Reverted e99d8eb, mainly because I think it's very unlikely to be the culprit and it's helpful with core perf.

I see how theory 2 is strong having looked into it, but I had another solid idea I'm tossing in here before merging: unaccountable weirdness due to publishing multiple projects async at the same time. Maybe missing a filter, somewhere, which guarantees we're receiving updates from the correct project folder….

Anyway, LGTM, merging, and LMK if you feel strongly about e99d8eb.

PR #356, opened 2018-03-08
fix: Don't crash if the function signature is something we cannot parse

tsortwehttam:
OK to merge.

Short review.

Purpose of changes:

- Fix crash that occurs when opening projects that include ES6 destructuring parameters.

Completed checkin tasks:


roperzh (commented):
LGTM

Are there any other places where a similar thing could happen? for example expressions/actions body?

tsortwehttam:
Not in this pathway I think. The function params are special since we use them to build the list of summonables to inject into the function. But the body is just treated as a text blob.

roperzh:
Gotcha, thank you :)

PR #359, opened 2018-03-12
[mono] Community changes

stristr:
OK to merge.

Medium-length review.

Purpose of changes:

- Everything that needs to change in mono to support community Phase 2.

- [x] Modify "replace summonables with default `0`" functionality (currently used with Lottie) to instead use default `1`, which matches edit mode and is overall safer (e.g. you can divide by 1).
- [x] Extend Core right-click menu to show a "Fork this component" option. We use the Fetch API (browser support is good enough, especially on Mac) wrapped in a try/catch to check with Inkstone if the project should be "forkable".
- [x] `@haiku/sdk-inkstone` tweaks to community API interfaces to match final functionality/needs.
- [x] New "HaikuStatic" bundle generated during publish. (I made the call not to experiment flag this because the exporter integration is known to be extremely fault-tolerant and has good test coverage for its fault tolerance.)
- [x] Adds support for `haiku://fork/:organizationName/:projectName`-type URLs in Creator, including the ability to pop a new/unauthed user straight into fork-land as soon as they sign in.

Regressions to look for:

- None expected, but post-merge this should all go through QA.

Notes for reviewers:

- This PR has significant Inkstone and community site dependencies that are being PR'ed asynchronously—specifically around forking functionality (including the check to see if the right-click menu should show the forking option at all and the actual forking functionality). This will make end-to-end testing difficult if you don't have a local inkstone setup. However, for safe merge, you only need to confirm that the app works correctly even without  any of the inkstone work deployed. In Core, this means: the request to "see if this project can be forked" should fail until that endpoint is actually deployed, so the "Fork this project" option shouldn't show up in the wild. In the app, this means: as long as you don't go to a `haiku://fork/…` type of URL, all the usual project init stuff is working fine.

Completed checkin tasks:


packages/@haiku/core/src/renderers/dom/createRightClickMenu.ts
```diff
@@ -93,6 +93,20 @@ export default function createRightClickMenu(domElement, component) {
   const menu = findOrCreateMenuElement(doc);
 
   const escaper = doc.createElement('textarea');
+  const metadata = component._bytecode && component._bytecode.metadata;
+  let forkAvailable = false;
+
+  if (metadata && metadata.project && metadata.organization && window && window.fetch) {
+    try {
+      window.fetch(`https://inkstone.haiku.ai/v0/community/OrgGuy/CheckTutorial`).then(({ok}) => {
```
roperzh: Heads up, _just in case_ (I'm not saying that we should change it!), here's the [browser support](https://caniuse.com/#feat=fetch) for `fetch`

screen shot 2018-03-12 at 09 51 29

packages/@haiku/core/src/renderers/dom/createRightClickMenu.ts
```diff
@@ -128,14 +140,25 @@ export default function createRightClickMenu(domElement, component) {
...
     );
+    if (forkAvailable) {
+      lines.push(
+        '<a onMouseOver="this.style.backgroundColor=\'rgba(140,140,140,.07)\'"' +
+        ' onMouseOut="this.style.backgroundColor=\'transparent\'"' +
+        ' style="display:block;color:black;text-decoration:none;padding: 5px 13px;line-height:12px;"' +
+        ` href="https://share.haiku.ai/u/${metadata.organization}/${metadata.project}/fork` +
+        '" target="_blank">' +
+        sharePageIcon +
+        ' Fork this component</a>',
```
roperzh: nit: why don't just use a template literal for the whole string?

roperzh (approved):
LGTM! very nice!

packages/haiku-formats/src/exporters/haikuStatic/haikuStaticExporter.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Parses class-local bytecode using internal methods.
+   *
+   * Essentially replaces all non-scalar timeline values with the equivalent scalar value of edit mode.
+   */
+  private parseBytecode() {
+    this.visitAllTimelineProperties((timeline, property) => {
+      const timelineProperty = timeline[property];
+      for (const keyframe in timelineProperty) {
+        const {value} = timelineProperty[keyframe];
+        if (typeof value === 'string' || typeof value === 'number') {
```
tsortwehttam: Properties may also be objects, arrays, booleans, and `null`.

packages/haiku-formats/src/exporters/haikuStatic/haikuStaticExporter.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Interface method to provide raw output.
+   * @returns {{}}
+   */
+  rawOutput() {
+    if (!this.bytecodeParsed) {
+      this.parseBytecode();
+    }
+
+    // Put bytecode through the JSON wringer to ensure our output is pure.
+    return JSON.parse(JSON.stringify(this.bytecode));
```
tsortwehttam, line 43: Just noting since I haven't traced this to code upstream, but it's possible that haiku/core (and possibly, by accident, some of our model utilities) attach properties to the bytecode that result in the object being circular, which would cause a crash at this line. (If need be, `ActiveComponent#getReifiedDecycledBytecode` can be called to get an object without cycles.)

packages/haiku-formats/src/exporters/injectables.ts
```diff
@@ -23,7 +23,7 @@ class DefaultStub {
    */
   get (_, property) {
     if (property === Symbol.toPrimitive) {
-      return () => 0;
+      return () => 1;
```
tsortwehttam, line 26: This is a good idea. But (and keep in mind this may also be an issue in haiku/core), if somebody sets `rotation.[x|y|z]` to an expression,  a default value of 1 radians is quite different from 0 radians.

packages/haiku-plumbing/src/publish-hooks/saveExport.js
```diff
@@ -2,7 +2,7 @@ import fse from 'haiku-fs-extra'
 import { handleExporterSaveRequest } from 'haiku-formats/lib/exporters'
 
 export default (request, activeComponent, cb) => {
-  const bytecodeSnapshot = activeComponent.fetchActiveBytecodeFile().getReifiedBytecode()
+  const bytecodeSnapshot = activeComponent.fetchActiveBytecodeFile().getReifiedDecycledBytecode()
```
tsortwehttam, line 5: Heh, I guess nm my comment above

packages/haiku-creator/src/react/Creator.js
```diff
@@ -953,7 +974,86 @@ export default class Creator extends React.Component {
...
+      setTimeout(() => {
+        this.openNewlyForkedProject(forkedProjectName, numAttempts + 1)
+      }, FORK_OPERATION_TIMEOUT)
+    }
+
+    this.props.websocket.request(
+      { method: 'getProjectByName', params: [forkedProjectName] },
+      (err, forkedProject) => {
+        if (err) {
+          this.showForkingError()
```
tsortwehttam, line 1038: Did you mean to early return here? (Also see line 1047 below where we may eventually want to add an early return.)

tsortwehttam (commented):
Only one comment I think is necessary to address before merging, but look closely at my other comments which may be important (your call).

packages/@haiku/core/src/renderers/dom/createRightClickMenu.ts
stristr: Yeah, I mentioned this without supporting data in PR description: "(browser support is good enough, especially on Mac)".

My approximate thinking is, if you're on mobile, you can't right-click. If you're on desktop and you're not using a supported browser on desktop, you're already not going to be impressed by Haiku enough to want to fork.

packages/@haiku/core/src/renderers/dom/createRightClickMenu.ts
stristr: Ah yeah, I also thought about this way too much. I think mainly for readability, the surrounding code also broke lines, and I was going to for a light touch and didn't refactor. My IDE added the line breaks before `style` and `onMouseOut`, though, so I should take those out.

packages/haiku-formats/src/exporters/haikuStatic/haikuStaticExporter.ts
stristr: Hmmm…which? (Was going off `schema.ts`).

And for these, if we're short-circuiting an expression with the default return value of `1`, is that problematic?

packages/haiku-plumbing/src/publish-hooks/saveExport.js
stristr: :smile:

packages/@haiku/core/src/renderers/dom/createRightClickMenu.ts
roperzh: Sounds good! sorry for missing that bit in the PR description.

packages/haiku-formats/src/exporters/injectables.ts
stristr: I believe this is identical to edit mode, but was having trouble finding the code…. @roperzh, can you confirm?

packages/haiku-creator/src/react/Creator.js
stristr: Yes :smile:

packages/haiku-formats/src/exporters/haikuStatic/haikuStaticExporter.ts
tsortwehttam: JIT properties targeting (for example) <path> `d` and <polygon> `points` can take arrays describing the points data; we can accept a color object for colors; and `null` can signal no content for components that accept injected content. (This may not be relevant right now, but it will become a thing soon.) Maybe it's sufficient to just skip these.

stristr: Fixed [here](https://github.com/HaikuTeam/mono/commit/51ac015b). Also shored up Lottie export's ability to handle this.

stristr:
Merging!

PR #362, opened 2018-03-13
Fix transform issues with Figma-produced SVG content

tsortwehttam:
OK to merge.

Short review.

Purpose of changes:

- Handle SVG content as produced by Figma, i.e., parse transform component strings whose values are delimited by spaces.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/helpers/parseCssTransformString.ts
+++ b/packages/@haiku/core/src/helpers/parseCssTransformString.ts
@@ -13,7 +13,14 @@ function separate(str) {
   const type = bits[0];
   return {
     type,
-    values: bits[1].replace(')', '').split(/\s*,\s*/gi).map((str2) => parseCssValueString(str2, type)),
+    // The transform component string may look like any of these:
+    //    translate(0.12,3)
+    //    translate(0.12 3.555)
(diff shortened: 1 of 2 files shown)


tsortwehttam:
For future reference:

We have to convert the attributes of all "raw" SVG content to align with our own layout engine, such that all elements can be transformed using one unified API.

To do that, a parsing step that occurs whenever a component (raw SVG content) is instantiated onto the stage. It converts layout instructions such as `style="transform: translate(1, 2)"` (and lots of other things) into one of our objects, like `{"translation.x": 1, "translation.y": 2}`.

The transform component string for all SVGs we've seen to date always delimits individual values with a comma, but Figma-generated SVGs use spaces: `translate(1 2)` vs. `translate(1, 2)`. The result was we were dropping the y-value, leading to Figma objects looking wrong when placed on stage.

Also for posterity, the chain of events is:

```
ActiveComponent#instantiateComponent
-> ActiveComponent#instantiateMana
-> Template.prepareManaAndBuildTimelinesObject
-> convertManaLayout
-> parseCssTransformString
```

tsortwehttam:
Got LGTM from @roperzh

PR #367, opened 2018-03-14
proposal: do not close the app on crashes in dev

roperzh:
OK to merge.

Short review.

Purpose of changes:

On development, when the app crashes, it is useful to have the stack trace visible at all times, and the comfort to navigate the code with devtools without time constraints.

My current workflow is to read the error as fast as possible and hit cmd + R to reload the app before it crashes, because reloading is way faster than having to `yarn go`.

Adding @matthewtoast as a reviewer, because iirc, he implemented this and I could be missing something.

### DIFF
--- a/packages/haiku-creator/src/entry.js
+++ b/packages/haiku-creator/src/entry.js
@@ -32,8 +32,10 @@ import {remote} from 'electron'
 
     // Give Raven some time to send a crash report before we close
     setTimeout(() => {
-      remote.getCurrentWindow().close()
-    }, (process.env.NODE_ENV === 'production') ? 500 : 50000)
+      if (process.env.NODE_ENV === 'production') {
+        remote.getCurrentWindow().close()
+      }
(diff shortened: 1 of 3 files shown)


stristr:
I think this is a great idea, although looks like we could move the `process.env.NODE_ENV === 'production'` up higher in all cases.

tsortwehttam:
I'm OK with this, but just want to say that the main reason I retained crashing is: Having it crash in dev has a small safety advantage in confirming that the app doesn't just stay open when it should, in fact, crash. (Long ago, we had some very insidious bugs where a part of the app crashed, but the app remained open, making it seem like it was working, leading to lost work.) If that seems like it won't be a risk, then LGTM.

roperzh:
I see, so there are constraints! I moved up the production check clauses as Sasha suggested, but I will leave the final choice up to you @matthewtoast close/merge as you think is better! :)

PR #370, opened 2018-03-14
poc: Pre-interpolate keyframes

tsortwehttam:
For @stristr's consideration.

Tested only as far as confirming that playback still _works_ with `morph-loop`.

Have not run tests, tested demos, nor even profiled this.

This is just a POC of a _theoretical_ perf optimization.

Feel free to just close this.


tsortwehttam:
When I profile this, I see a ~20% improvement in animation frame performance. However, it's a risky change. Maybe we can explore this more in the future. Sasha has cherry-picked this into his perf branch for local review, so closing this.

PR #371, opened 2018-03-14
feat: support DMG upload and syndication.

stristr:
OK to merge.

Short review.

Purpose of changes:

- Use `.dmg` instead of `.zip` for new users.

- [x] Upload `.dmg` during distro upload.
- [x] Back up and release `.dmg` (instead of `.zip`) as `-latest` for new users.

Notes to reviewer:
 - `squirrel-express` [always](https://github.com/HaikuTeam/squirrel-express/blob/master/server.js#L38) filters down to files ending with `.zip` for its syndication list, so parking a `.dmg` in a subdirectory without a `-pending` flag should be fine.
 - This will require manual intervention the first time we use it; specifically, we'll have to park the `…-latest.dmg` file in S3 or the "backup" step will fail. This will give us a great opportunity to test it before we actually release.

Completed checkin tasks:

### DIFF
--- a/scripts/distro-syndicate.js
+++ b/scripts/distro-syndicate.js
@@ -21,16 +21,16 @@ const distribution = deploy.deployer[environment].distribution
 const s3 = initializeAWSService('S3', region, key, secret)
 const cloudFront = initializeAWSService('CloudFront', region, key, secret)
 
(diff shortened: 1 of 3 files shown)


roperzh (approved):
LGTM!

PR #374, opened 2018-03-15
fix: check if abspath exist before matching

roperzh:
OK to merge.

Short review.

Purpose of changes:

Fix a crash on race conditions: https://sentry.io/haiku-systems/javascript-glass/issues/495111461/?referrer=slack


stristr:
Do we know why this race condition exists/what its underlying cause is?

roperzh:
No, but we are already doing similar checks [next to where this code is called](https://github.com/HaikuTeam/mono/blob/9ada86acf2ac39443bf57ae8fc86c3b8ed59ee43/packages/haiku-glass/src/react/Glass.js#L1760) so I figured it was fine for now.

Sorry for not clarifying.

stristr:
Merging.

PR #377, opened 2018-03-15
Fix fork on open

stristr:
OK to merge.

Short review.

Purpose of changes:

- Fix fork on open in release mode!

- [x] Attaches an earlier event listener than the one that wasn't firing to ensure we can handle a URL that was used to open the app.


roperzh:
LGTM!

PR #386, opened 2018-03-16
Fix changelog

stristr:
OK to merge.

Short review.

Purpose of changes:

- Fix changelog to loosen the requirement that we always have a sliced-off changelog version exactly equal to the last version seen.
- Use correct semver ordering so that e.g. `1.3.15` is higher than `1.3.2`.

Completed checkin tasks:


tsortwehttam:
LGTM.

PR #389, opened 2018-03-19
Publish ui storybook

roperzh:
OK to merge.

Short review.

- [x] Port the publish UI to react storybook
- [x] Center the Publish UI modal against the whole app screen

Regressions to look for:

- The path of some `PNG` images in packages/haiku-creator/public/stylesheets/global.css was wrong, after changing it they seem to still work fine in dev, but they will in prod? do we change that path during build?

Completed checkin tasks:

### DIFF
--- a/packages/haiku-ui-common/src/react/ShareModal/index.tsx
+++ b/packages/haiku-ui-common/src/react/ShareModal/index.tsx
@@ -11,7 +11,7 @@ const STYLES = {
   wrapper: {
     width: 500,
     overflow: 'hidden',
-    left: '50%',
+    left: 'calc(50% + 150px)',
     transform: 'translateX(-50%)',
     top: 110,
     margin: 0,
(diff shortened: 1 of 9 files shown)


packages/haiku-ui-common/src/react/ShareModal/index.tsx
```diff
@@ -11,7 +11,7 @@ const STYLES = {
   wrapper: {
     width: 500,
     overflow: 'hidden',
-    left: 'calc(50% + 150px)',
```
taylorpoe: It *was* originally centered against the whole app screen. I added this piece to center it over the stage like the Actions modals are for consistency. Had a couple members of the team say the way it was originally felt off-center because it wasn't centered with the stage. I think we should leave it as was here before this commit unless you think I'm misunderstanding this.

roperzh: LOL! my bad, sorry! I remembered some members saying something about the position, and I just went ahead and changed to whatever the opposite of the current state was.

I will revert, thanks!

tsortwehttam:
Awesome, LGTM, merging.

PR #394, opened 2018-03-23
fix: improve error Figma messaging when there are no assets to import

roperzh:
OK to merge.

Short review.

Summary of changes:

Current error message is ambiguous and doesn't provide info about the problem at all, I have found a user very confused on Intercom about this. 

Maybe we can revisit the UX later and add an empty file to the library or something on those lines, but for now we are going to display this toast:

screen shot 2018-03-22 at 21 55 26

Notes for reviewers:

**Important note**: this also includes a tweak in the logic to display `error.err` as a message if there's any, which will also display messages coming from the Figma API that we don't handle yet. IMO this is fine and even useful, because when the user reported this error I was in the dark about what could be the cause since the default message is too generic.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Figma.js
+++ b/packages/haiku-serialization/src/bll/Figma.js
@@ -138,6 +138,7 @@ class Figma {
       const uri = API_BASE + 'images/' + id + '?' + params.toString()
 
       if (ids.length === 0) {
+        // eslint-disable-next-line
         reject({err: 'We couldn\'t find any groups or slices in your project.'})
       }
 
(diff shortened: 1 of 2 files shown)


packages/haiku-serialization/src/bll/Figma.js
```diff
@@ -137,6 +137,11 @@ class Figma {
       const params = new URLSearchParams([['format', 'svg'], ['ids', ids]])
       const uri = API_BASE + 'images/' + id + '?' + params.toString()
 
+      if (ids.length === 0) {
+        // eslint-disable-next-line
+        reject({err: 'We couldn\'t find any groups or slices in your project.'})
```
tsortwehttam: Should you early return here?

roperzh: fixed, thanks :)

PR #397, opened 2018-03-24
Fix crashy crashy

stristr:
OK to merge. Potentially more coming to those who wait.

Short review.

Summary of changes:

- Don't crash if unable to resolve loopback.haiku.ai. Instead, try to fall back to `0.0.0.0` first.
- Don't crash if the authentication request does not return an HTTP response, which is a signal we got intercepted by a proxy. Send the auth error (which is likely to have the only info we can use to debug to Sentry) and show a "slightly less unhelpful" error message instead.

Notes for reviewers:

- Branched from `3.1.24`, hence weird extra commits + changelog. Please ignore.


tsortwehttam:
LGTM so far. (lmk if you push any more commits and I'll review those too.)

stristr:
I'm trialing a build that actually tries to fix this using the `electron-proxy-agent` package, and at least adds more logging, with a user on Community Slack—sent him the build, but want to hold off on submitting until we have either confirmation the fix actually works (unlikely) or have more data. Just gonna merge this one to get it into the build.

PR #399, opened 2018-03-24
tweak: bump everyone's project limit up

taylorpoe:
DO NOT MERGE until backend work is done as well.

Changes project limit from 15 to 50.


stristr:
Per comment in Asana, back end work is "done"—insofar as there's a bug that doesn't stop you from creating > MAX projects already.

PR #401, opened 2018-03-24
feat: add link in user menu to user's Community profile

taylorpoe:
OK to merge.

[Adds an external link to a user's profile from the user menu.](https://app.asana.com/0/591362404931103/606823903724181)

image

Completed checkin tasks:


taylorpoe:
I've yet to do any mixpanel hookups on _anything_ I have made in the app. Is it as easy as changing out the tracking string like below? Is that all I need to do? Or is there an adjacent piece in Mixpanel that I'd need to go over to and set up?

```
onClick={() => {

            mixpanel.haikuTrack('creator:project-browser:user-menu-option-selected', {
              option: 'view user profile'
            })
          }}>
```

tsortwehttam:
@taylorpoe  Yes, that's all you should have to do. I recommend signing into Mixpanel, look at `Haiku App (Dev)`, then do the action in dev mode to verify that the instrumentation works. (You should see the data reflect the action almost immediately.) You may also want to change the `option` field too, but that's just optional metadata for the ping in any case.

PR #404, opened 2018-03-28
Fix: use strict mode in Actions/Expressions editors

roperzh:
OK to merge.

Short review.

Summary of changes:

This will prevent plumbing (and therefore the app) from crashing 
when parsing Actions/Expressions in certain scenarios where things 
that are valid in non-strict mode cause syntax errors in strict mode, 
for example `000` is interpreted as an octal in non-strict mode,
but causes a SyntaxError in non-strict mode.

This was biting us because all of the parsing infrastructure is based
on babylon, which uses strict mode by default.

Regressions to look for:

- Actions/Expressions bytecode saving/editing

Completed checkin tasks:


tsortwehttam:
LGTM, merging.

PR #416, opened 2018-04-01
Add React Router

taylorpoe:
I've noticed that most of the popular React projects use this and we have yet to install it. Figured better late than never. It's declarative — so that's good. May very well fix many imperative things that are wrong that we don't even know about. #unknownunkonwns

From their docs:
> We call it “Dynamic Routing”, which is quite different from the “Static Routing” you’re probably more familiar with.

"**you’re probably more familiar with**" spoken like a true recalcitrant teen. At the end of the day I do wish their team and ours a blessed Easter and April 1st. Please merge asap.

### DIFF
--- a/packages/haiku-creator/package.json
+++ b/packages/haiku-creator/package.json
@@ -60,6 +60,7 @@
     "react-motion": "0.4.7",
     "react-onclickoutside": "^6.4.0",
     "react-popover": "^0.4.18",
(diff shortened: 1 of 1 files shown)


stristr:
LGTM, merging.

tsortwehttam:
LGTM 🎆 🍾 🎉 merging 💯 🥇 👍

PR #417, opened 2018-04-05
Public/private tweaks

roperzh:
OK to merge.

Short review.

Summary of changes:

- [Display privacy toggle state instantly](https://app.asana.com/0/607625220671718/612367254519327)
- [Make projects "Private" by default](https://app.asana.com/0/607625220671718/615645754685511)

Notes for reviewers:

Instead of using any kind of storage as initially suggested on Asana, I took another approach:

- "Whitelist" in plumbing the `IsPublic` attribute that is coming when listing projects, allowing `projectObject`s to have the attribute
- Update the state that contains `projectObject` with the proper value of `isPublic` when the toggle is switched (we need some form of state management)

Do you agree that this is a better approach? If no I can change it.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -854,6 +854,10 @@ export default class Creator extends React.Component {
     )
   }
 
+  onProjectPublicChange (isPublic) {
+    this.setState({projectObject: {...this.state.projectObject, isPublic}})
+  }
+
   launchProject (projectName, projectObject, cb) {
     projectObject = {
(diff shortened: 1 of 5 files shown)


stristr:
Since we're doing "optimistic" here, this seems like a good, and less heavy-handed, approach. You should check project instantiation and forking if you haven't already—they should be fine, but there is some hackery for formatting the project objects in Plumbing right after they're instantiated and worth double-checking.

tsortwehttam:
LGTM, merging.

PR #420, opened 2018-04-09
Perf experiment: debounced hard reloads

stristr:
DO NOT MERGE.

Candidate for future merge after next release goes out.

As design-sketched in earlier call:
 - Store locks as a queue so we can introspect them and see if there are any pending requests.
 - Short-circuit hard reloads in progress if a new hard reload is already queued, while ensuring any custom rehydrations are still run.

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -66,6 +66,23 @@ const keyframeUpdatesToHotComponentDescriptors = (keyframeUpdates) => {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tsortwehttam:
Although this solves the perf issue with dragging, the same issue would occur when shifting between dragging and resizing and rotating. Given that is the case, I decided to implement a custom rehydration for `updateKeyframes` which seems to work well, and will use that instead.

stristr:
Reopening for potential future merge.

stristr:
Goodbye for now, debounced hard reloads.

PR #430, opened 2018-04-12
fix: use selectSoftly during selectAll to fix selectAll.

stristr:
OK to merge.

Short review.

Summary of changes:

- Fix weird `selectAll` regression. If not using `selectSoftly`, the multiselect becomes self-canceling for "reasons".


stristr:
Invoking emergency protocol.

PR #431, opened 2018-04-12
fix: un-break origin upgrade in headless mode.

stristr:
OK to merge.

Short review.

Summary of changes:

- Don't try to mount event listeners if we don't have a mount. Fixes `upgradeBytecodeInPlace` silent crash in headless mode which leads to the necessary mount offsets not being persisted on disk.
- Log `upgradeBytecodeInPlace` crashes. They are too important for us to not know what they are.


stristr:
Invoking emergency protocol.

PR #433, opened 2018-04-12
Percy fix

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes [Putting coin in Percy is broken](https://app.asana.com/0/607625220671718/630526049204162/f). Somewhere in `eventing-refactor` we landed an optimization that only allowed one event handler per event type. This is now fixed to: "one event handler per event type per selector".


stristr:
Invoking emergency protocol.

PR #436, opened 2018-04-13
tweak: update icon file

taylorpoe:
The `/build` folder was the only place I could find this file to update. Is this the right place to update it though?

Looking way better at smaller sizes

image


stristr:
That's the one! Failing health check is unrelated, flaky test in `haiku-serialization`. LGTM, merging.

PR #437, opened 2018-04-14
UI tweaks

taylorpoe:
OK to merge.

[Stage should not show grab-cursor unless spacebar is pressed](https://app.asana.com/0/607625220671718/603221132353436)
The piece we're trying to avoid is showing the grab cursor on the stage which can't be grabbed. This task was made before we had marquee selection so it's somewhat mitigated, but I still think it's better (and more common in design tools) to always show a grab cursor. I've traded it out for a default cursor — though we do still see a grabbing cursor when the mouse is down. I think this is okay, as the user is either drawing a marquee or grabbing an element.

[Should be able to use cmd as well as shift to multi-select segments](https://app.asana.com/0/607625220671718/614190753418984) Self-explanatory. Zack should finally sleep well now.

Regressions to look for:

Odd timeline selection states

Completed checkin tasks:

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -1827,7 +1827,7 @@ export class Glass extends React.Component {
   getCursorCssRule () {
     if (this.isPreviewMode()) return 'default'
     if (this.state.isAnythingRotating || this.state.isAnythingScaling) return 'none'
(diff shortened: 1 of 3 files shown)


tsortwehttam:
Health check requests to be fixed in master, so merging. LGTM.

PR #438, opened 2018-04-14
Utility to help solve unsolved mysteries

tsortwehttam:
DO NOT MERGE JUST YET.

Short review. 

Summary of changes:

- As discussed with @stristr -- this is a mostly-transparent utility intended to help us diagnose problems that are difficult to reproduce that tend to arise during free-form QA and slams. The intention of this change is that, before beginning to QA/slam, users should set the environment variable `HAIKU_RECORDER_WORKSPACE=~/Desktop/rec open /Applications/Haiku` when they launch Haiku via command line. When set, this will instruct Haiku to stream verbose logs of whitelisted method calls and UI events into the text file `~/Desktop/rec/logfile`. The user is then expected to QA/slam normally. If any problem occurs, that log file can be sent along with the bug report. This is intended to be more useful than plumbing logs since we log at the granularity of method calls, interspersed with info about UI events, instead of just high-level actions which often give little insight into race conditions and other tricky issues. Modules to listen to can be configured in `monkey.config.js`.

- Also did a tiny refactor to a bit of plumbing that had some legacy args. Will need to test that websocket broadcast works before merging.

Other notes:

- This pathway is a no-op if the env var is not configured. There should be no risk to having the hooks in the production codebase.

- New sessions will erase the pre-existing log file, so QA users need to remember to move the log file out of the recording dir before starting another session.

- Enabling this does incur a performance hit since logging a lot of method calls plus their arguments and piping that data as text into a file write stream is not 0 cost.

- There is no reason this is called "monkey" other than it works by monkey patching require'd modules to intercept all method calls.


tsortwehttam:
Closing this PR because I'm going to roll this in along with the mc stuff instead, where it has been extremely helpful.

PR #446, opened 2018-04-20
feature: add an overlay to elements hovered

roperzh:
OK to merge.

Short review.

Summary of changes:

5ada3c495c6d4568757367

- Add an outline to elements when hovered on stage.
- Add an outline of stage elements when their row is hovered on the timeline.

Notes for reviewers:

- This featured is flagged under `OutliningElementsOnStage` and currently disabled in production
- I tried really hard to draw an outline of the svg shape instead of the bounding box, but it will require a huge amount of work by the approaches that I have found. I have settled to outline for now.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -67,6 +67,8 @@ const DIMENSIONS_RESET_DEBOUNCE_TIME = 100
 
 const BIG_NUMBER = 99999
 
+const OUTLINE_CLONE_SUFFIX = 'outline-clone-helper'
+
 function isNumeric (n) {
   return !isNaN(parseFloat(n)) && isFinite(n)
 }
@@ -252,6 +254,15 @@ export class Glass extends React.Component {
         case 'dimensions-reset':
           this.handleDimensionsReset()
           break
+        case 'hoverElement':
+          this.hoverHighlight(args[1])
+          break
+        case 'unhoverElement':
+          this.unhoverHighlight(args[1])
+          break
+        case 'selectElement':
+          this.unhoverHighlight(args[1])
+          break
       }
     })
 
@@ -262,6 +273,15 @@ export class Glass extends React.Component {
           break
         case 'setInteractionMode':
           this.handleInteractionModeChange()
+          break
+        case 'hoverElement':
+          this.hoverHighlight(args[1])
+          break
+        case 'unhoverElement':
+          this.unhoverHighlight(args[1])
+          break
+        case 'selectElement':
+          this.unhoverHighlight(args[1])
           break
       }
     })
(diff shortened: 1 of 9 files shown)


tsortwehttam:
Great job tracking all the logic down that needed to be in place to set this up. 👍👍 

There are two implementation details that I think we need to look very carefully at.

**Issue 1:** The way it attaches the DOM nodes _inside the root element of the user's component_, which is actively managed by haiku/core: https://github.com/HaikuTeam/mono/pull/446/files#diff-faac8754effb1942994db60572619863R845 Since haiku/core doesn't know about these additional nodes, there could be unexpected behavior in rendering.

(To understand the code that could be affected by this, take a look at this code for when the DOM renderer needs to flush the retained scene graph to the DOM:  https://github.com/HaikuTeam/mono/blob/development/packages/%40haiku/core/src/renderers/dom/renderTree.ts#L20 Make note of how it iterates through the children to determine whether a node needs to be added, updated, removed, or replaced.)

Imagine that the stage has 5 elements instantiated, [A,B,C,D,E]. Are there any issues with the hover node when you:

- Hover E while deleting E
- Hover E while deleting D
- Hover D while deleting E
- Hover E and cut an element (also try: cut, copy, paste)
- Delete E, hover D, then undo the deletion
- Instantiate F, delete F, hover E, then 
- Z-move E to front, then Z-move E to back, hover E, and undo/redo
- _Future: Hover E, then group [C, D, E]_
- _Future: Hover E, then create component from [C, D, E]_
- _Future: Hover E, when E is a node with a `$repeat` defined_
- Any of the above while also quickly hovering and un-hovering (race conditions?)

There might not be any problems, but it's worth testing.

I think before merging we should mount these overlay DOM nodes at higher level than the user's own component -- ideally in the same way we draw the transform controls. See here in Glass.js:  https://github.com/HaikuTeam/mono/blob/development/packages/haiku-glass/src/react/Glass.js#L1576 (What we have is a `this.refs.overlay` DOM element which sits in front of the component node. This allows the transform controls to be displayed in front of the element in question without potentially conflicting with the component's own rendering.) My gut says this revision should be small, so let me know if you aren't sure how to do it or see some complexity in it.

**Issue 2:** The way it uses `.cloneNode()`, can you double check that cloning is _shallow_ (only clones the root node, not its descendants)? We will also get rendering conflicts on stage if we have mounted elements with duplicate `id`s and `haiku-id`s. I believe `.cloneNode()` clones all descendants too by default, but that may have changed.

packages/haiku-glass/src/react/Glass.js
```diff
@@ -263,6 +274,15 @@ export class Glass extends React.Component {
...
           break
+        case 'hoverElement':
+          this.hoverHighlight(args[1])
+          break
+        case 'unhoverElement':
+          this.unhoverHighlight(args[1])
+          break
+        case 'selectElement':
+          this.unhoverHighlight(args[1])
+          break
```
stristr: Nit: please consolidate cases in a `switch` statement that have the same outcome (here and below):

```
case 'unhoverElement':
case 'selectElement':
  this.unhoverHighlight(args[1])
  break
```

roperzh:
Although we discussed this on Slack, I'm writing this down for posterity:

**Issue 1**: we decided to use another DOM node to attach the outlines instead of inserting them in the stage mount controlled by Glass, [here's the commit](https://github.com/HaikuTeam/mono/pull/446/commits/a61428296e2036c4cd873aee8a7989d9c6e905ec) with the fix.

**Issue 2**: `node.cloneElement` does a shallow copy unless a truthy param is provided, which is exactly what we want for now.

This is ready for another review 🔍

tsortwehttam:
Cool, LGTM, merging.

PR #452, opened 2018-04-25
fix: properly print help info when an unknown command is used in nib

roperzh:
OK to merge.

Short review.

Summary of changes:

Random issue that I have found investigating possible issues with `haiku init`. Currently, if you provide an invalid command to the CLI, it prints misleading info, for example:

```bash
§ ./bin/haiku asdf

NAME:
    undefined

USAGE:
    haiku asdf

VERSION:
    3.2.16
```

This modifies it to print the default help screen instead.


stristr:
LGTM, merging.

PR #454, opened 2018-04-26
@haiku/sdk-inkstone additions.

stristr:
OK to merge.

Short review.

Summary of changes:

- Add endpoint to log out.
- Fix issue when subbing `xhr` module for `request` to reduce web bundle size—it spells `delete` like `del`.
- Hook up `ContinueUrl` in the places they're add in [inkstone #42](https://github.com/HaikuTeam/inkstone/pull/42).

Regressions to look for:

- Create account, request password reset, or resend confirmation email endpoints not working in their production applications (should be backward-compatible).

Completed checkin tasks:


packages/@haiku/sdk-inkstone/src/index.ts
```diff
@@ -238,14 +253,28 @@ export namespace inkstone {
...
+      } else {
+        body.email = params.Email;
+        body.continue_url = params.ContinueUrl;
+      }
+
       const options: requestLib.UrlOptions & requestLib.CoreOptions = {
         url: inkstoneConfig.baseUrl + ENDPOINTS.RESET_PASSWORD,
         headers: baseHeaders,
-        body: JSON.stringify({email}),
+        body: JSON.stringify(body),
```
tsortwehttam, line 274: Should we try/catch this and return a 422 if parse fails?

tsortwehttam (dismissed):

packages/@haiku/sdk-inkstone/src/index.ts
tsortwehttam: ~parse~ -> serialize

stristr: Don't think so—with the TypeScript code base that will soon be using this, compile-time typechecking ensures an invalid type won't ever get passed here. The current JavaScript codebase that uses it passes params as a string (email only), at which point the generated JavaScript boils down to:

```
const body = {};
body.email = 'foo@bar.com';
...JSON.stringify(body);
```

So if it fails with a runtime error, that should be the correct/expected behavior.

stristr:
Merging.

PR #456, opened 2018-05-01
Windows/Linux port

jonaias:
OK to merge.

Short review.

Summary of changes:

- It implements everything to develop and build for Windows (nsis installer) and Linux platform (deb installer).

Regressions to look for:

- Validate `scripts/compile-all.js` on a mac os 

Notes for reviewers:

- Tests and lint are not working on windows yet (TODO list)
- Windows installer is set to 32bits only for now
- Platform specific bugs will be sent in a different PR
- Not signing windows installer (TODO list) 

Completed checkin tasks:


scripts/distro-build.js
```diff
@@ -1,15 +1,31 @@
 const fse = require('fs-extra')
 const cp = require('child_process')
 const path = require('path')
-const deploy = require('./deploy')
+//const deploy = require('./deploy')
```
jonaias: I should uncomment this

index.js
```diff
@@ -16,7 +17,7 @@ if (process.env.HAIKU_APP_LAUNCH_CLI === '1') {
 } else {
   const {app, dialog} = require('electron');
 
-  if (process.env.NODE_ENV === 'production' && !app.isInApplicationsFolder()) {
+  if (process.env.NODE_ENV === 'production' && os.platform() === 'darwin' && !app.isInApplicationsFolder()) {
```
stristr, line 20: Just putting this on your radar: we have this check because we try to do an in-place update of the code using our update server, which downloads a zipfile and expands it to overwrite the existing copy in `/Applications/Haiku.app`. Although it doesn't have to work the same way on Windows, we do need a way to update the running app, close it down and reopen the latest version.

package.json
```diff
@@ -107,6 +109,26 @@
         "dmg"
       ]
     },
+    "linux": {
+    "maintainer" : "Haiku Team <contact@haiku.ai>",
+    "category": "Graphics;Development;VectorGraphics",
+    "icon" : "build/icon.icns",
```
stristr: Nit: please fix indentation starting at `"maintainer"` through `"icon"`.

scripts/compile-all.js
```diff
@@ -12,29 +13,41 @@ if (!process.env.NODE_ENV) {
   process.env.NODE_ENV = 'development'
 }
 
+
+function getModificationTime(file){
+  return new Date(fs.statSync(file).mtime)
+}
```
stristr: Prefer:

```
const getModificationTime = (file) => new Date(...)
```

scripts/compile-all.js
```diff
@@ -12,29 +13,41 @@ if (!process.env.NODE_ENV) {
...
-    const filesChanged = Number(
-      cp.execSync(`find ${pack.abspath}/src -type f ${timeFilter} | wc -l | tr -d ' '`).toString().trim()
-    )
 
-    if (filesChanged > 0) {
-      log.warn(`Detected ${filesChanged} changed file(s) in ${pack.shortname}. Compiling....`)
-      cp.execSync('yarn run compile', { cwd: pack.abspath, stdio: 'inherit' })
+    /* Get modified file since last compilation */
+    let files = glob.sync(`${pack.abspath}/src/**`, {})
+    let modifiedFiles = files.filter(file => getModificationTime(file) > lastCompileTime)
```
stristr: Nit: add "optional" parentheses: `(file) =>`.

scripts/distro-build.js
stristr: This shouldn't be commented out (it's needed).

scripts/distro-build.js
```diff
@@ -1,15 +1,31 @@
...
+    
+    //cp.execSync(".\\node_modules\\.bin\\build --windows --publish=never", { cwd: ROOT, stdio: 'inherit' })
+
+    cp.execSync(".\\node_modules\\.bin\\build --windows --publish=never --config.forceCodeSigning=false", { cwd: ROOT, stdio: 'inherit' })
+
+
+}
+else if (process.env.HAIKU_RELEASE_PLATFORM==='linux') {
+    cp.execSync(`./node_modules/.bin/build --linux --publish=never`, { cwd: ROOT, stdio: 'inherit' })
+}
```
stristr: Couple of items of feedback on this file:
 - Please remove commented-out code. 
 - The file needs a lint pass.
 - Generally prefer:

```
switch (x) {
  case a:
    // ...
    break
  case b:
    // ...
    break
  case c:
    // ...
    break
  default:
    throw new Error(`Unexpected value: ${x}`)
}
```

over

```
if (x === a) {
  // ...
} else if (x === b) {
  // ...
} else if (x === c) {
  // ...
}
```

I think 3 is the magic number of items at which point a `switch` statement becomes preferable to `if`/`else if`.

scripts/distro-configure.js
```diff
@@ -12,13 +13,27 @@ var ENVS = { development: true, production: true }
...
+function getReleasePlatform(){  
+  switch (os.platform()) {
+    case 'darwin':
+      return 'mac'
+    case 'win32':
+      return 'windows'
+    case 'linux':
+      return 'linux'
+    default:
+      return 'unknown'
```
stristr: Ideally, we should throw an error here to prevent any callers from proceeding.

scripts/distro-prepare.js
```diff
@@ -43,21 +59,56 @@ fse.writeJsonSync(path.join(DISTRO_DIR, 'package.json'), {
...
+  rimraf.sync('packages/*/*/node_modules/haiku-*')
+  log.log('success!')
+} catch (err) {
+  log.err(err)
+}
+
 // Place node modules in their canonical location.
-logExec(ROOT, `rsync ${RSYNC_FLAGS} node_modules ${DISTRO_DIR}`)
+try {
+  log.log(`Copy node modules to their canonical location. It may take a while..`)
```
stristr: Nit: don't use template literals if there are no expressions inside.

scripts/distro-prepare.js
```diff
@@ -31,10 +32,25 @@ function logExec (cwd, cmd) {
   return childProcess.execSync(cmd, { cwd, stdio: 'inherit' })
 }
 
+
 // Clear previous contents.
-logExec(ROOT, `rm -rf ${DISTRO_DIR}`)
+try {
```
stristr: Here and throughout this file: the code we replaced (`logExec`, which used `childProcess.execSync`) crashed on errors (process returns with a non-zero success code). We should preserve that and not do `try/catch`. If any of the commands that set up the distro code fail, the whole build should crash at the first point of failure.

scripts/distro-prepare.js
```diff
@@ -43,21 +59,56 @@ fse.writeJsonSync(path.join(DISTRO_DIR, 'package.json'), {
...
 logExec(ROOT, `yarn install ${YARN_INSTALL_FLAGS} --production=false`)
 // Uglify sources in release.
 logExec(ROOT, 'node ./scripts/distro-uglify-sources.js')
-// Rebuild.
-logExec(ROOT, `yarn electron-rebuild --module-dir ${path.join(DISTRO_DIR, 'node_modules', 'nodegit')}`)
+
+
+// Rebuild native modules
+// Use 32bits on windows to ensure maximum compatibility 
+if (process.env.HAIKU_RELEASE_PLATFORM==='windows'){
```
stristr: As discussed this morning: can we be sure to build both a 64-bit and a 32-bit version here?

stristr (commented):
Nice work in progress!

I left some general comments, some more important than others. (Comments prefixed with "Nit:" are the less important ones.)

Note that `yarn lint-all` does not cover the mono `./scripts` directory; for that, you should run `yarn lint` and clean up any errors. I didn't flag all the places where I noticed lint errors, but they include:
 - Using double quotes (`"`) instead of single quotes (`'`).
 - Not leaving a newline at the bottom of a file.
 - Inconsistent spaces around conditional statements (`if (x){` instead of `if (x) {`)

scripts/distro-build.js
stristr: Ah, sorry, had already started reviewing when you wrote this!

index.js
jonaias: That would be fun in windows. Probably we can use windows registry to check install directory or something like that (since windows installers allow to change install dir)

scripts/distro-build.js
jonaias: np

package.json
jonaias: :+1

scripts/compile-all.js
jonaias: :+1:

scripts/compile-all.js
jonaias: :+1:

scripts/distro-build.js
jonaias: I was used to employ switch only with return statements because forgetting a break statement could be really dangerous (old c habbits, no lint there). Btw looks much better with a switch statement.

I'll let commented code there because its about windows installer signing (TODO). When I address it, I'll remove.

scripts/distro-configure.js
jonaias: :+1:

scripts/distro-prepare.js
jonaias: :+1:

scripts/distro-prepare.js
jonaias: You are right. Fixed it. :+1:

scripts/distro-prepare.js
jonaias: Its in my todo list

jonaias:
Thanks about the tip about `yarn lint`

Todo list:
- [ ] Move getReleasePlatform (duplicated on distro-build.js and start.js) to somewhere saner and avoid duplication
- [ ] Add option to change at runtime archtecture (32bits or 64bits) to build windows installer
- [ ] Check what to do with windows installer signing
- [ ] Tests aren't working on windows

jonaias:
TODO:
- [ ] Test Open in Terminal across several OS
- [ ] Check windows shortcut keybindings with Taylor Poe ( Especially zoom in and zoom out)

jonaias:
I implemented OS detection according to discussion on Slack. I want some feedback if typescript code organization (import/export) were done as expected. If so, I'll refactor every os detection occurrence (famous grep).

OS detection implementation ba8eb8df166d0b8e3ea3dee6ddd386469f6ca610  and usage example https://github.com/HaikuTeam/mono/commit/95a59c1da162566e78c6d89806946f23d128c47f  

I plan to tag this PR as OK to merge on Sunday

jonaias:
Things to do before windows Release:
- [ ] [Beta] Add environment variable to select windows installer architecture (32 or 64)
- [ ] [Beta] Fix Figma integration
- [ ] [Beta] Fix all Windows tests
- [ ] [Release] Keyframes dots and text (vertically) are not aligned in timeline
- [ ] [Release] On Windows, drag from asset library to stage shows a not allowed cursor 
- [ ] [Release] Syndication (Including auto-update/windows installer signing)

Looked at the history, I didn't find a merge workflow. Do we use rebase workflow? I'm not that used to rebase workflow, that's why I left with merge.

tsortwehttam:
I made those outstanding todo items into tasks in the Asana "Venus" backlog.

stristr:
LGTM, merging.

PR #457, opened 2018-05-01
fix: allow the Vue adapter to receive options

roperzh:
OK to merge.

Short review.

Summary of changes:

With the changes introduced in 79225a, `haikuComponentFactory`
doesn't accept options wrapped in an `options` object anymore,
instead it expects the values at the root of the config element.

Regressions to look for:

- Some code in the React adapter has been modified, all demos are working fine for me, but may be worth a gander.

Completed checkin tasks:


stristr (approved):
LGTM. Note/nit: you should feel free to use `const` instead of `var` in unit tests written in JavaScript. We don't need a transpiler to understand those keywords anymore 😺!

roperzh:
Ah! thanks about that 😄 

I pushed a commit to use `let`/`const` instead. I will merge as soon as the CI is green.

PR #464, opened 2018-05-09
Add Haiku bytecode definition

jonaias:
OK to merge.

Short review.

Summary of changes:

Regressions to look for:

Notes for reviewers:

- Need to check the best place to concentrate Haiku types. Note that `haiku-common/src/types/enums.ts` contains types for curve, a type used by Haiku bytecode. My proposal is either to move all Haiku related types to `@haiku/core/api/` or `haiku-common/src/types/`. Curve definition is temporally duplicated until we decide.

- I'll build test cases from demo projects (ideally using them in place). To test locally I used this script
``` bash
mkdir t
cp packages/@haiku/core/api/HaikuBytecode.ts t
cat t/HaikuBytecode.ts >> t/t.ts
echo "function func(tim: HaikuBytecode): boolean {  return false;}" >> t/t.ts

for OUTPUT in $(ls packages/@haiku/core/demo/projects/*/code/main/code.js )
do
    a=`echo $OUTPUT | cut -d/ -f 6`
    cp $OUTPUT t/$a.js
    
    ecapaded=`echo $a | sed -r "s/-/_/g"`
    echo "import * as $ecapaded from './$a'" >> t/t.ts
    echo "func($ecapaded)">> t/t.ts
done 
```
- We might have some problem with nested index signatures on Haiku timeline typedef (https://basarat.gitbooks.io/typescript/docs/types/index-signatures.html). During the tests I'll have a better example.

Completed checkin tasks:

- [X] Wrote an automated test covering new functionality


packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+  Linear = 'linear',
+}
+  
+/**
+ * Haiku element tree. eg. <div><svg>...</div></svg>. 
+ * `source` and `identifier` are rarely used.
+ */
+export type HaikuTemplate = {
+  elementName: string;
+  attributes: {[attribute: string] : string;
```
tsortwehttam: Note: `source` and `identifier` are legacy and have changed to `haiku-source` and `haiku-var`.

We probably also want to add all of these: https://github.com/HaikuTeam/mono/blob/bfbdce8835c12dfb9f6c3165c58ed046fa781555/packages/haiku-serialization/src/bll/Template.js#L31

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+    source?: string;
+    identifier?: string; };
+  children : HaikuTemplate[] | string[];
+};
+
+/**
+ * Haiku states.
+ */
+export type HaikuStates = {
+  [stateName: string] : {value: string | {text: string}[] };
```
tsortwehttam: A state value can be any JSON type (`string`, `boolean`, `object`, `array`, `null`).

tsortwehttam: A state may also have the keys:
- `setter` (fn)
- `getter` (fn)
- `access` (string)
- `type` (string)

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+/**
+ * Haiku function `handler` for an specific `eventName`. 
+ */
+export type HaikuEventHandler = {
+  [eventName: string]: {handler: (any) => any};
```
tsortwehttam: Not sure if this would belong in the typing or not, but _at runtime_ the render system may add a `.original` function to keep track of the function signature and body prior to being wrapped. See here: https://github.com/HaikuTeam/mono/blob/bfbdce8835c12dfb9f6c3165c58ed046fa781555/packages/%40haiku/core/src/HaikuComponent.ts#L1035

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+ */
+export type HaikuEventHandlers = {
+  [elementName: string]: HaikuEventHandler;
+};
+
+/**
+ * Value of an element property in a given frame. 
+ */
+export type HaikuTimelineValue = {
+  value: (boolean | string | ((any) => string));
```
tsortwehttam: A timeline value can also be any valid JSON type, including `number`, `object`, `array`, `null`.

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+  player?: string;
+  type: string;
+  name?: string;
+  relpath: string;
+  core?: string;
+  version?: string;
+  username?: string;
+  organization?: string;
+  project?: string;
+  branch?: string;
```
tsortwehttam: `player` is legacy.

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Bytecode definition. Properties are *rarely* used.
+ */
+type HaikuBytecode = {
+  template: HaikuTemplate | string;
+  states?: HaikuStates;
+  eventHandlers: HaikuEventHandlers;
+  timelines: HaikuTimelines;
+  metadata?: HaikuMetadata;
+  properties?: HaikuProperties[]
```
tsortwehttam: `properties` is legacy.

packages/@haiku/core/api/HaikuBytecode.ts
```diff
@@ -1,38 +1,38 @@
...
+  easeInOutQuad,
+  easeInOutQuart,
+  easeInOutBounce,
+  easeInOutElastic,
+  easeInOutQuint,
+  easeInOutSine,
+  linear,
 }
-  
+
```
stristr: Hmmm, why do this?

These curves are used extensively in `haiku-formats` and I didn't see a commit fixing that code.

Additionally, although the linter doesn't enforce this (yet), I had been adopting the convention that enum keys should begin with Capital Letter For Emphasis.

stristr:
Hey, heads up (@jonaias)—I changed the base branch to `rc-mc-group-highlight`. Would you mind rebasing to it and force pushing to get clean history?

```
git fetch
git rebase origin/rc-mc-group-highlight
```

packages/@haiku/core/api/HaikuBytecode.ts
stristr: Suggestion for deprecated properties: don't type them but do keep them in for formal correctness, with JSDoc `@deprecated` symbol.

e.g.

```
  /**
   * @deprecated
   */
  properties?: any[]
```

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: As both haiku-source and haiku-var are string indexes, they are included by string index signature type ([attribute: string] : string). Added deprecated comment

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: I defined a type `HaikuStateTypes` to be used as state value type.  Added missing state keys.

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: You are right. As we discussed, bytecode typing is only considered on compile time, so it won't be necessary until we start adding types across all code base. For example, to add a .original function, if bytecode is HaikuBytecode, you should add this function as optional to HaikuBytecode.

I'm really excited to start using types everywhere!! I think i will save a large amount of time for us in the future!! On any new feature or refactoring, the compiler will help us =)

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: Added deprecated comment

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: @stristr thanks for suggestion, i'm adopting it.

packages/@haiku/core/api/HaikuBytecode.ts
jonaias: > Hmmm, why do this?

Don't worry its a very ugly temporary solution. The deal is because ["Keep in mind that string enum members do not get a reverse mapping generated at all."](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime), we cannot use string enum as type of curve `HaikuTimelineValue` attribute:
``` Typescript
export type HaikuTimelineValue = {
  value: (boolean | string | number | ((any) => HaikuStateTypes));
  edited?: boolean;
  curve?: keyof typeof Curve;  // <<<<<<<<<<HERE we cannot use old Curve definition eg. `curve?: Curve`
};
```
A way better explanation is given by [There's some oddities (perhaps bugs?) with how the enum is type checked, its not just a string literal type shorthand, which is what would be truly correct. Some issues I've bumped into](https://stackoverflow.com/q/35760096/1524655). And in our case, we use curve names from bytecode as string.

For now, I see two possible solutions
- Rename bytecode curve names to Capital Letter. eg. on bytecode file `curve: 'EaseOutSine'` instead of `curve: 'easeOutSine'`   (con: it will break compatibility)
- Use [String Literal Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types) only for bytecode type definition. eg  `type CurveType = "easeInBack" | "easeInCirc" | "easeInCubic"..` and keep them in sync with Curve definition (con: need to keep in sync Curve and CurveType )

I'm not happy with these solutions, so i'll come back latter to this. Maybe I can find better solution.

stristr: Just to keep "eye on the prize" here: I think for now, a perfectly suitable type for `curve?:` is simply `string`. We want to support arbitrary bezier curves, spring transitions, and more in the near future—and in order to do that, it should be clear that the "string" version soon to be extended to include `{x1: number, y1: number, x2: number, y2: number}` (one way to represent arbitrary beziers), `{tension: number, friction: number, velocity: number}` (one way to represent arbitrary springs)…and along the way, we might want users to be able to make a "curve library", give a custom curve a name like `"myCustomCurve"`, and refer to it by name (`"curve": "myCustomCurve"`). It might be a bezier, or a spring, or an arbitrary function `f(t): [0, 0] -> [1, 1]`…better IMO to not constrain this now, especially since it's a single field and the relevant type is a curve.

One other bit of related advice: we need to keep `@haiku/core` dependency free and open source, so nothing in `packages` outside of `packages/@haiku/core` is really fair game for including as a dependency unless we plan to open source it. Even though you technically _can_ import files from `haiku-common` into `@haiku/core` and use its types, the typing file `(.d.ts`) that comes out will confuse compilers in other TypeScript projects that use `@haiku/core` as a direct dependency. So for now, all typings that exist in other packages that you might want to use here should be moved into `@haiku/core`. `Curve` probably isn't one that you want or need, but in case there are others 😉 ….

jonaias: I was thinking about moving Curve to @haiku/core.. but with your explanation cleared things.  Thanks.

Sure, when other curve types are implemented, we can change it from string to exactly possible curve types.

stristr:
BUILDME!

packages/@haiku/core/test/TestHelpers.js
```diff
@@ -5,6 +5,9 @@ const HaikuDOMAdapter = require('../lib/adapters/dom').default;
 const HaikuDOMRenderer = require('../lib/renderers/dom').default;
```
stristr: Some inconsistent style here—note use of `"` instead of `'` and commented out code, etc.

Request, since I was going to do this anyway: as part of this branch, would you please convert our test suite to TypeScript, and add a script for linting it with the linter (e.g. `yarn test-lint`)? There are 33 test files, so shouldn't be a huge amount of work, and it would really help move the TypeScript project forward. There's a good reference for how to write `tape` tests using TypeScript in `packages/haiku-formats` you can imitate.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
+/**
```
stristr, line 1: Please note changes made in 4db9003fa7c967a53264624fca0b67ec864971ac here. There are some things missing in `tslint-config-haiku` I'll be adding later next week.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export type BytecodeStates = {
+  [stateName: string]: BytecodeState;
+};
+
+/**
+ * Haiku bytecode function `handler` for an specific `eventName`. 
+ */
+export type BytecodeEventHandler = {
+  [eventName: string]: {handler: (any) => any};
```
stristr: Let's definitely improve this `(any) => any` 😃.

These are all `(component: HaikuComponent, event: Event, ...args: any[]) => void`…I think.

Similar to the `PrimitiveType` above, definitely want to export a `BytecodeEventHandler` type, which will come up elsewhere….

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+/**
+ * Tuples of `elementName` and `BytecodeEventHandler`. 
+ */
+export type BytecodeEventHandlers = {
+  [elementName: string]: BytecodeEventHandler;
```
stristr, line 79: These aren't really `elementName`s, but `eventSelector`s.

Note/nit: since it's better to have no name than a wrong name, feel free to stick with the simpler `[key in string]: BytecodeEventHandler ` when it makes sense to do so.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+  edited?: boolean;
+  curve?: string;
+};
+
+/**
+ * Map `propertyName` to a list of `frameNum` and its given `BytecodeTimelineValue`.
+ * *Most* of tests use frameNum as string.
+ */
+export type BytecodeElementTimeline = {
+  [propertyName: string]: {[key in string]: BytecodeTimelineValue};
```
stristr: Note edit I made in 4db9003fa7c967a53264624fca0b67ec864971ac here: when a "`number`" is a key of a JS object, it's really a string masquerading as a number. Although TypeScript lets you type it the way you had it (and the ES6 `Map<T, U>` type actually lets you type it the way we _want_ it 😸), "number keys of plain objects" are not really a thing.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+type HaikuBytecode = {
+  template: BytecodeTemplate | string;
+  states?: BytecodeStates;
+  eventHandlers: BytecodeEventHandlers;
+  timelines: BytecodeTimelines;
+  metadata?: BytecodeMetadata;
+  /**
+   * @deprecated as of 3.2.20
+   */
+  properties?: BytecodeProperties[];
```
stristr: Given that this is fully deprecated, `BytecodeProperties` is also fully deprecated, and (FWIW) `any[]` would be perfectly suitable here.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+  options?: any
```
stristr: This is not `any`, but (@matthewtoast please correct me if I'm wrong) an instance of `typeof DEFAULTS` as found in `Config.ts`. This should also get a type.

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export type BytecodeEventHandlers = {
+  [elementName: string]: BytecodeEventHandler;
+};
+
+/**
+ * Value of an element property in a given frame. 
+ */
+export type BytecodeTimelineValue = {
+  value: (boolean|string|number|((any) => BytecodeStateType));
```
stristr: Similar to `BytecodeEventHandler`, these are very special functions, always the result of calling `Haiku.inject` (`HaikuInjectedFunction`)? And the way we call `Haiku.inject` is very special too: `Haiku.inject(expressionHandler: (...args: HaikuInjectable[]) => PrimitiveElement, ...argNames: string[])`, so as part of typing bytecode we should trace the peripheral behaviors of `Haiku.inject`.

`HaikuInjectable` is a union type `BytecodeStateType|HaikuSummonable`, where `HaikuSummonable` is yet another type 💃!!!

Look for `INJECTABLES['$user']`, `INJECTABLES['$element']`, etc. to see how these work.

stristr (changes requested):
Nice progress! :+1:

Someday, I would like to turn on the `"noImplicitAny"` config option in `tsconfig.json`. If you do that now, you'll notice that there are errors all over—ideally, this PR will reduce those errors wherever bytecode and things "beneath" bytecode are found, so I think there's still quite a bit of work to be done here to trace the types where they go. I mentioned `haiku-formats` as a good testing ground for how complete this work is because it's a much smaller code base where you _could_ turn on `"noImplicitAny"`, delete all the lazy `any` types in the `haiku-formats` package, apply the `HaikuBytecode` type, and see what happens. It's still important to carefully trace the path of bytecode through the `HaikuComponent` constructor and down to all the other things that happen with `this.bytecode`. As you go, strike out bytecode-descended things that are explicitly or implicitly typed as `any` and replace them with their actual types.

Some examples from the constructor `bytecode.template = xmlToMana(bytecode.template);`—so `xmlToMana` is a function that needs to accept `string` and return `BytecodeTemplate`. So go type its arguments, and the methods it calls, and come back and make sure everything still compiles.

A more complex example is:

```
  eachEventHandler (iteratee: Function) {
    if (this.bytecode.eventHandlers) {
      for (const eventSelector in this.bytecode.eventHandlers) {
        for (const eventName in this.bytecode.eventHandlers[eventSelector]) {
          iteratee(
            eventSelector,
            eventName,
            this.bytecode.eventHandlers[eventSelector][eventName],
          );
        }
      }
    }
  }
```

Okay, so `iteratee` isn't really just any function…it's really an `(eventSelector: string, eventName: string, handler: BytecodeEventHandler) => void`! To be thorough, we should make sure everything that calls `eachEventHandler` works with this type.

In general, trace what happens to `this.bytecode` inside `HaikuComponent.ts`, type the related methods, and come back and make sure everything still compiles. `HaikuComponent.ts` constructor was meant to be a starting point for finding all the places we should be applying types to flesh out and validate the API, but not the only place where we apply the related types.

packages/@haiku/core/test/api/05_HaikuTypes.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+
+test('Test Haiku bytecode typing on compile time', function (t) {
+  /* Load bytecode type definition from disk */ 
+  const haykuBytecodeTypeString = fs.readFileSync('src/api/HaikuBytecode.ts',{encoding:'utf8'})
+
+  /* Get all bytecode file names from demo projects */
+  const bytecodeFiles = glob.sync(`demo/projects/*/code/main/code.js`, {})
+
+  const nonConformantBytecodeFiles = [
```
stristr: Note: these non-conformant bytecode files are actually supported by our "in-place migration" procedure that happens inside `Migration.ts`. You should be able to trigger and fix the related compiler errors (through additional typing) by explicitly using the `HaikuBytecode` type on `component.bytecode` in `Migration.ts:runMigrations`.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: That will be very useful.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: As they involve typing from elsewhere (HaikuComponent, Event, etc), i think it might be better to add these types later (even create other files under api/ to move some types there) , when starting adding types to HaikuComponent. 

I know from some tests I did, HaikuBytecode isn't complete to be used everywhere (I set the type of bytecode HaikuComponent to HaikuBytecode, and there are some fields that are add to be used in runtime only).

One idea is to have a property under HaikuBytecode called for example `runtime`, so every added runtime objects would stay under this property.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: Thanks, changed it to eventSelector.

One thing I noticed when adding typing to haiku-formats is that BytecodeTemplate would be better as BytecodeNode.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: Nice info! I will update it to be a string.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: :+1:

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: I'll put in TODO list for this PR

packages/@haiku/core/test/TestHelpers.js
jonaias: Sounds a nice improve using only typescript for testing, I can do this tomorrow.

packages/@haiku/core/test/api/05_HaikuTypes.test.js
jonaias: I'll check it tomorrow. (notice that I suppose the migration occurs in runtime and this test is compile time only)

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: I saw this type being used in `packages/haiku-formats/src/exporters/injectables.ts` function `evaluateInjectedFunctionInExportContext`.

jonaias:
Back to WIP. My TODO list:
 [] - Add options correct type
 [] - Transform tests in TS only
 [] - Check about typings of  HaikuSummonable/EventHandler
 [] - Check about using migration.ts to test typing on non conformant bytecodes

packages/@haiku/core/src/api/HaikuBytecode.ts
stristr: Yeah, I agree with that being a better name for the type.

stristr:
Sounds good! Please layer into TODO list:
 - [ ] Enable `"noImplicitAny"` in `packages/haiku-formats/tsconfig.json`, remove all `any` types, and fix lint errors.
 - [ ] Per above, carefully trace the path of bytecode through the `HaikuComponent` constructor and down to all the other things that happen with `this.bytecode`. As you go, strike out bytecode-descended things that are explicitly or implicitly typed as any and replace them with their actual types.

jonaias:
- [x] Add options correct type
- [x] Transform `@haiku/core test` into TS only
  -  An interesting fact, the test `core/test/unit/reflection.functionToRFO.test.js` wasn't converted because typescript compile scrambles function comments on transpilation. To make our test suite simpler, the `reflection.functionToRFO.test.ts`, calls .js test, enabling it also being considered a TS test
- [x] Check about typings of HaikuSummonable/EventHandler
  - HaikuSummonable was implemented. EventHandler wans't used by `haiku-formats`, so I preferred not to implement it
- [x] Check about using migration.ts to test typing on non conformant bytecode
   - It would be lot of trouble to mix compile time and runtime test, I opted to avoid it. (Migration itself should be covered by its tests)
- [x] Enable `noImplicitAny` in packages/haiku-formats/tsconfig.json, remove all any types, and fix lint errors.
   - DONE!! :metal: :metal:.  Note that  some `haiku-formats` functions (generally ones interfacing BodyMoving, there wasn't typescript already made definitions - DefinitelyTyped -, so I left some of them as any)
- [ ] Per above, carefully trace the path of bytecode through the HaikuComponent constructor and down to all the other things that happen with this.bytecode. As you go, strike out bytecode-descended things that are explicitly or implicitly typed as any and replace them with their actual types
   - Wasn't able to fix types in available time.

jonaias:
BUILDME!

packages/@haiku/core/src/api/HaikuBytecode.ts
```diff
@@ -0,0 +1,259 @@
...
+  // timeline; useful in headless
+  freeze?: boolean,
+
+  // Whether we should loop the animation, i.e. restart from 
+  // the first frame after reaching the last
+  loop?: boolean,
+
+  // Optional function that we will call on every frame, provided 
+  // for developer convenience
+  frame?: (() => void)|null,
```
stristr, line 175: Question: is the `|null` union type necessary here? I believe the `?` means "nullable", so `|null` is unnecessary here and below.

packages/@haiku/core/package.json
```diff
@@ -54,6 +57,7 @@
     "@types/tape": "^4.2.30",
     "async": "^2.5.0",
     "chokidar": "^1.7.0",
+    "cross-env": "^5.1.4",
```
stristr: Hmmm, why is this dependency added here? It doesn't seem to be used anywhere.

packages/@haiku/core/package.json
```diff
@@ -26,18 +26,21 @@
...
-    "test:perf": "yarn compile && tape \"test/perf/**/*.test.js\" | tap-spec",
-    "test:api": "yarn compile && tape \"test/api/**/*.test.js\" | tap-spec",
-    "test:render": "yarn compile && tape \"test/render/**/*.test.js\" | tap-spec",
-    "test:unit": "yarn compile && tape \"test/unit/**/*.test.js\" | tap-spec",
-    "test-report": "tape \"test/**/*.test.js\" > test-result.tap",
+    "test:perf": "yarn compile-full && tape \"~test/perf/**/*.test.js\" | tap-spec",
+    "test:api": "yarn compile-full && tape \"~test/api/**/*.test.js\" | tap-spec",
+    "test:render": "yarn compile-full && tape \"~test/render/**/*.test.js\" | tap-spec",
+    "test:unit": "yarn compile-full && tape \"~test/unit/**/*.test.js\" | tap-spec",
+    "test-report": "tape \"~test/**/*.test.js\" > test-result.tap",
```
stristr: Couple of things:

1. Let's go ahead and make just one `yarn test` command that only compiles once (`tape "~test/**/*.test.js"`). Will be doing this across other projects with multiple "suites" set up like this. No need for `test:perf`, …, `test:unit`….
2. Let's also have one `yarn lint` which lints both source files and tests in a single command.
3. Let's clean up `~test` after running tests, like the `rm -rf ~test` in `haiku-formats`…
4. …and in case something crashes, please add `~test` to `.gitignore` so it doesn't sneak into source control.

packages/@haiku/core/src/Layout3D.ts
```diff
@@ -5,6 +5,74 @@
 import computeMatrix from './layout/computeMatrix';
 import computeSize from './layout/computeSize';
 
+
+export type LayoutSpecType = {
```
stristr: Nit/naming: it seems a bit extra to call a type for "thing that is a layout spec" `LayoutSpecType`. `LayoutSpec` is a better, more concise name. See also: `BytecodeStateType` (maybe `BytecodeStateValue` is better?)….

We could even change `PrimitiveType` to `Primitive` or similar, come to think of it.

Also: I would like to move this and all additional typings to the `src/api` "namespace" for future plans.

packages/@haiku/core/test/TestHelpers.ts
```diff
@@ -0,0 +1,179 @@
+const jsdom = require('jsdom');
```
stristr, line 1: To complete the migration to TypeScript, we should rewrite all `require()` statements as `import` statements.

Please don't do this now; just filing a note for later.

packages/haiku-formats/src/exporters/bodymovin/bodymovinTypes.ts
```diff
@@ -13,6 +20,24 @@ export type BodymovinProperty = {
...
-  inheritFromParent: boolean,
+  parentId: string;
+  inheritFromParent: boolean;
 };
+
+export type BodymovinShape = any;
+export type BodymovinFill = any;
+
+export type BodymovinTransform = any;
+// TODO: It is commented due: https://github.com/Microsoft/TypeScript/issues/13042,
```
stristr: Actually, this should work using:

```
[key in TransformKey.TransformOrigin]: {…}
```
etc.

That specific issue is about using `index in` typing for integer enums. String enums should be fine with `key in` typing.

packages/haiku-formats/package.json
```diff
@@ -33,6 +34,10 @@
     "lodash": "^4.17.4"
   },
   "devDependencies": {
+    "@types/async": "^2.0.49",
+    "@types/bezier-easing": "^2.0.0",
+    "@types/bluebird": "^3.5.20",
```
stristr: Hmm…why is `@types/bluebird` added here when `bluebird` is not a direct (dev or non-dev) dependency of `haiku-formats`? I think the places it needs to be added are the dependencies `bluebird` is coming from. (Use `yarn why bluebird` to see where it's coming from.)

This happens to work because of how everything is set up using yarn workspaces, but in isolation, this dependency is in the "wrong place".

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -5,11 +5,16 @@ import {
...
-import * as mapKeys from 'lodash/mapKeys';
+
+// TODO: Use import after `haiku-serialization` port to typescript. 
+// `haiku-serialization` is imported using `require` to avoid:
+// "TS7016: Could not find a declaration file for module..."
+// tslint:disable-next-line:variable-name
+const Template = require('haiku-serialization/src/bll/Template');
+// tslint:disable-next-line:variable-name
+const LoggerInstance = require('haiku-serialization/src/utils/LoggerInstance');
+
```
stristr: [See 9106dd832 for the solution to the problem solved temporarily here]

stristr (commented):
Looking very nice!

Just a few additional change requests here, and this should be good to merge later today 👍.

packages/@haiku/core/src/api/HaikuBytecode.ts
jonaias: If I understood correctly, according to "Optional parameters and properties" from, https://www.typescriptlang.org/docs/handbook/advanced-types.html, ? means "undefinable".

Anyway, I agree with you... It is much better left it "undefinable" than "nullable".

packages/@haiku/core/package.json
jonaias: yep, removed it

packages/@haiku/core/package.json
jonaias: Accept your challenge (and suggestion) and now we are using ts-node. As you forecasted, it was pretty straightforward.

packages/@haiku/core/src/Layout3D.ts
jonaias: I agree with LayoutSpecType, very redundant, fixed it.. About BytecodeStateType, if we change to BytecodeStateValue, it could means its value instead its type, idk..

packages/haiku-formats/package.json
jonaias: Fixed it.. the indirect dependency was `haiku-sdk-creator`.

packages/@haiku/core/test/TestHelpers.ts
jonaias: Some of them are related to https://github.com/HaikuTeam/mono/commit/9106dd832d9b0ad0339a063d56e297a1e05b1239.  Ok.

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
jonaias: :+1:

packages/haiku-formats/src/exporters/bodymovin/bodymovinTypes.ts
jonaias: I thought the last reply in the commented link (a PR closing the issue) would enable something like [key in keyof TransformKey]. Or at least make commented code work.

Accepted the suggestion.

packages/haiku-formats/src/exporters/index.ts
```diff
@@ -5,7 +5,7 @@ import {BodymovinExporter} from './bodymovin/bodymovinExporter';
 import {HaikuStaticExporter} from './haikuStatic/haikuStaticExporter';
 
 // @ts-ignore
-import LoggerInstance from 'haiku-serialization/src/utils/LoggerInstance';
+import * as LoggerInstance from 'haiku-serialization/src/utils/LoggerInstance';
```
stristr: :+1:, sorry about that!

stristr:
Please note changes in 507558922e40f59e2cdaefc91ae8c4d19e5dc8fa:
 - [nit] removed some redundant usages of `yarn` in command specs.
 - [important] access `tape` binary at `./node_modules/.bin/tape` instead of `../../../node_modules/tape/bin/tape`. The reason for this is `@haiku/core` is an [open source repository](https://github.com/HaikuTeam/core) that people should be able to pull and contribute to like any other, but `mono` is not. The only reason you find the `tape` binary at that location is the way we're using [Yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) in `mono` specifically. The symlink from `./node_modules/.bin/tape` -> `../../../../../node_modules/tape/bin/tape` -> `../../../node_modules/tape/bin/tape` will work in both contexts.

stristr:
LGTM! Very, very nice work here 👍 😺 🥇!

Merging.

stristr (approved):

PR #481, opened 2018-05-18
State transitions

jonaias:
OK to Merge.

Short review.

Summary of changes:

- Implement state transitions (tween)

Regressions to look for:

- If not using state transitions,  none. (Its a tangential code path)

Notes for reviewers:

Completed checkin tasks:


packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -147,6 +150,9 @@ export default class HaikuComponent extends HaikuElement {
     this._states = {}; // Storage for getter/setter actions in userland logic
     this.state = {}; // Public accessor object, e.g. this.state.foo = 1
 
+    // Instantiate StateTransitions. Resposible to store and execute any state transition.
```
jonaias: typo

stristr:
Please note failing test.

@jonaias, can you see into Jenkins to note which test is failing? (It's one you wrote 😸!)

packages/@haiku/core/demo/projects/statetransitions/code/main/angular-dom.js
```diff
@@ -0,0 +1,3 @@
+var HaikuAngularAdapter = require('@haiku/core/dom/angular')
```
stristr: Note: we have been in the habit of only committing the files we actually need for the demos to run.

This would be:

```
statetransitions
└── code
    └── main
        ├── code.js
        ├── dom.js
        ├── react-dom.js
        └── vue-dom.js
```
in this case (so no designs etc.)

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -332,16 +338,23 @@ export default class HaikuComponent extends HaikuElement {
     return this.state[key];
   }
 
-  setState(states) {
-    if (!states) {
-      return this;
-    }
-    if (typeof states !== 'object') {
+  setState(states: StateValues, transitionParameter: StateTransitionParameters) {
```
stristr: Nit: `transitionParameter:` should be `transitionParameter?:`

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -147,6 +150,9 @@ export default class HaikuComponent extends HaikuElement {
     this._states = {}; // Storage for getter/setter actions in userland logic
     this.state = {}; // Public accessor object, e.g. this.state.foo = 1
 
+    // Instantiate StateTransitions. Responsible to store and execute any state transition.
+    this.stateTransitions = new StateTransitions(this.state, this.getClock());
```
stristr: Nit: Based on how it works, a better name for this class would be `StateTransitionManager`.

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -332,16 +338,23 @@ export default class HaikuComponent extends HaikuElement {
...
+    if (!states || typeof states !== 'object') {
       return this;
     }
-    for (const key in states) {
-      this.set(key, states[key]);
+
+    /* If has transition parameter, it should be treated as a transition */
+    if (transitionParameter) {
+      this.stateTransitions.createNewTransition(states, transitionParameter);
+    }else {
```
stristr: *`} else {` (note space)…fixing linter this weekend!

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -332,16 +338,23 @@ export default class HaikuComponent extends HaikuElement {
...
   }
 
-  setState(states) {
-    if (!states) {
-      return this;
-    }
-    if (typeof states !== 'object') {
+  setState(states: StateValues, transitionParameter: StateTransitionParameters) {
+
+    /* Do not set any state if invalid */
```
stristr: Nit: please use `//` comments inside methods.

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
...
+  clock: HaikuClock;
+
+  // Store running state transitions
+  transitions: RunningStateTransition[];
+
+  constructor(states: StateValues, clock: HaikuClock) {
+    this.states = states;
+    this.clock = clock;
+    this.transitions = [];
+  }
```
stristr: Here's some cool stuff you can do in TypeScript:

```
export default class StateTransitions {
  private transitions: RunningStateTransition[] = [];

  constructor(private readonly states: StateValues, private readonly clock: HaikuClock) {}
  // ...
}
```

1. No need to use constructor to initialize properties with a known initial value.
2. The empty constructor body is not a mistake—no need to assign a variable named `states` to a property named `states`. Just name them in the constructor, and they will be automatically assigned. This is the only time you _should_ use the otherwise implicit `public` keyword—if you have the rare constructor argument that shouldn't be private (not the case here).
3. Use `private` and `readonly` whenever you can, especially for things like powerful objects that are passed into the `constructor` method. The compiler will enforce it and help you build good encapsulation into your code.

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
...
+
+
+  /**
+   * Create a new state transition.
+   */
+  createNewTransition(target: StateValues, parameter: StateTransitionParameters) {
+
+    // Copy current states as transition origin (needed to calculate interpolation)
+    const origin = {};
```
stristr: Nit: since `origin` is an overloaded concept used in our layout system, we might want to choose a better name for this variable like `transitionStart`.

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
+import Transitions from './Transitions';
+import HaikuClock from './HaikuClock';
+import justCurves from './vendor/just-curves';
+
+
+export type StateTransitionParameters = { curve: string, duration: number};
+
+export type StateValues = {[stateName: string]: any};
```
stristr: Didn't we invite the type `BytecodeState` instead of `any`, and call `StateValues` `BytecodeStates` in our last sprint? Let's reuse those types where they apply (I think this is one case where they do).

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Should be called on every tick. It cleans expired state transitions 
+   * and execute interpolation of running state transitions.
+   */
+  tickStateTransitions(): void {
+    const currentTime = this.clock.getTime();
+
+    // Helper functions 
+    const isExpired = (transition) => currentTime >= transition.endTime;
+    const isNotExpired = (transition) => currentTime < transition.endTime;
```
stristr: Consider declaring these at the module level (but not exporting them, so they're functionally private), with `currentTime` made into an argument.

This should help with GC dumps.

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.states[key] = states[key];
+    }
+  }
+
+  /**
+   * Should be called on every tick. It cleans expired state transitions 
+   * and execute interpolation of running state transitions.
+   */
+  tickStateTransitions(): void {
+    const currentTime = this.clock.getTime();
```
stristr: I see an opportunity to return early here if `this.transitions.length === 0`. We should probably use it—perf is always something to keep an eye on.

packages/@haiku/core/src/StateTransitions.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    // Copy current states as transition origin (needed to calculate interpolation)
+    const origin = {};
+    for (const key in target) {
+      // Ignore state if it doesn't pre exist
+      if (key in this.states) {
+        origin[key] = this.states[key];
+      } else {
+        // delete state from target if it doens't pre exist
+        delete target[key];
```
stristr: Hmmm…but is this necessary? I might be missing some subtlety here, but if it's fine to leave it, we could make this code a lot more readable.

stristr (commented):
Looking nice so far! Left some comments. Is this still considered WIP, or ready for review?

packages/@haiku/core/demo/projects/statetransitions/code/main/angular-dom.js
jonaias: :+1:

packages/@haiku/core/src/HaikuComponent.ts
jonaias: :+1:

packages/@haiku/core/src/HaikuComponent.ts
jonaias: Makes more sense!

packages/@haiku/core/src/StateTransitions.ts
jonaias: What a good lecture! Thanks!

FYI, I didn't set states as `readonly`, as write on states from within StateTransitionManager.

packages/@haiku/core/src/StateTransitions.ts
jonaias: :+1:

packages/@haiku/core/src/StateTransitions.ts
jonaias: The problem doing that is I won't be able to call filter, making the function less readable. (I could bind it, but would be slower and less readable). The alternative is to make function less readable by not using filter. (As readable is completely subjective, if you prefer I can change it)

After you saying about GC dumps, I became concerned and I went to research (Mostly for my curiosity). From what I understood, V8 Turbofan inline small functions like these at JIT compilation: 

> The inlining in V8 is driven by a bunch of heuristics. First of all, the engine puts limits on the size of the inlinee, that is the --max-inlined-nodes flag (which defaults to 200) is the upper bound for the number of AST nodes (as determined by parsing), and in addition the --max-inlined-nodes-cumulative flag (which defaults to 400) limits the total number of AST nodes that can be inlined into an optimization unit, and the --max-inlining-levels flag (which defaults to 5) limits the inlining depth.

https://docs.google.com/document/d/1VoYBhpDhJC4VlqMXCKvae-8IGuheBGxy32EOgC2LnT8/edit

packages/@haiku/core/src/StateTransitions.ts
jonaias: Well spotted! :+1:

packages/@haiku/core/src/StateTransitions.ts
jonaias: Nice comment. I improved the comment there to make more clear why this is needed. 

This is done to avoid math rounding due to timing error, for example, imagine a transition to be expired on 5000ms, the `tickStateTransitions` is called on 4999ms and next called `tickStateTransitions` on 5001ms, so to make sure final state value is transitionEnd, transitionEnd is set to states when deleting expired transitions.

jonaias:
What I've done:
- Implemented State Transitions in a new Class(`StateTransitionManager.ts`) to make it modular (State Transition can be easily reused)
- Unit and Functional (using dom) tests for State Transitions
- Moved interpolation stuff from `Transitions.ts` to newly created `Interpolate.ts` (As Interpolate now is used by Transitions and StateTransitionManager. If I shouldn't had done it, I can revert it without any problem). 
- Added typing to Interpolate.ts.
- Moved Curve definition to newly created `@haiku/core/src/api/Curve.ts` from previous `haiku-common` (Also, no problem reverting)
- Added Curve type to HaikuBytecode.
- Created function on `HaikuComponent.ts` to reset states to initial value using transitions.

jonaias:
BTW, as soon as it gets merged, I'll update docs with the new setState parameter.

packages/@haiku/core/src/HaikuContext.ts
```diff
@@ -296,6 +296,10 @@ export default class HaikuContext extends HaikuBase {
     // Only continue ticking and updating if our root component is still activated and awake;
     // this is mainly a hacky internal hook used during hot editing inside Haiku Desktop
     if (!this.component.isDeactivated && !this.component.isSleeping) {
+
+      // Execute state transitions
+      this.component.tickStateTransitions();
```
tsortwehttam: Due to multi-component, `tickStateTransitions` also needs to be called on every component in the component hierarchy. You can visit every `HaikuComponent` in the hierarchy with `HaikuComponent#visitGuestHierarchy`.

tsortwehttam (changes requested):
Made one change request, otherwise this is looking good to me.

packages/@haiku/core/src/HaikuContext.ts
jonaias: Thanks for heads up! Done!

packages/@haiku/core/src/StateTransitions.ts
stristr: Note: I think you actually can use `readonly` for `states`, because they are not _reassigned_ inside of this class.

 - Not allowed for `readonly x`: `x = …`, x += …`, etc.
 - Allowed (I think—check if it compiles!) for `readonly x`: `x.push(...)`, `delete x[...]`, etc.

packages/@haiku/core/src/StateTransitionManager.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Delete every running transition
+   */
+  deleteAllStateTransitions() {
+    this.transitions = [];
+  }
+
+  getNumRunningTransitions() {
```
stristr: Notes:
 - For "dumb getter" methods like this, prefer ES6 getters, i.e. instead of `getNumRunningTransitions()` use `get numRunningTransitions()` and access as `instance.numRunningTransitions` instead of `instance.getNumRunningTransitions()`.
 - That said, since apparently this method is only used in tests, it should move to our test suite (to keep down our bundle size).

packages/@haiku/core/src/StateTransitionManager.ts
```diff
@@ -0,0 +1,116 @@
...
+    // If have no running transitions, optimize it by using early retrun
+    if (this.transitions.length === 0) {
+      return;
+    }
+
+    const currentTime = this.clock.getTime();
+
+    // Helper functions 
+    const isExpired = (transition) => currentTime >= transition.endTime;
+    const isNotExpired = (transition) => currentTime < transition.endTime;
```
stristr: Continuing conversation from before…

Note: we aren't only targeting V8 (`@haiku/core` is an open source package that targets "the browser"—theoretically we should play nice with everything that can draw SVG), so we should be confident that we're micro-optimizing around everything we target. Most modern browsers have some form of inlining like what you linked, _but_…I'm not sure inlining addresses the issues of runtime allocations happening inside the `tickStateTransitions` method which gets called on every RAF. I'm pretty sure these functions need to get allocated (and dumped when we exit the function scope) every time we call `tickStateTransitions`. Since there's a `this`-binding from the current clock time, the way to doing this with one-time allocations would be to build into the class:

```
private expiredFilter = (transition: RunningStateTransition) => this.clock. getTime()  >= transition.endTime;
// etc.
```

Now, with all that said…having looked carefully at how this routine is written I think we don't need filters and there's a far more elegant way to write everything. Check this (I haven't checked it with the compiler, so there may be subtle bugs):

```
  private isExpired(transition: RunningStateTransition, currentTime: number) {
    return this.clock.getTime() >= transition.endTime;
  }

  /**
   * Should be called on every tick. It cleans expired state transitions 
   * and execute interpolation of running state transitions.
   */
  tickStateTransitions(): void {

    const currentTime = this.clock.getTime();
    const interpolatedStates = {};

    for (let index = 0; index < this.transitions.length; index++) {
      const transition = this.transitions[index];

      if (this.isExpired(transition, currentTime)) {
        // If expired, simulate it was calculated exactly on endTime.
        Object.assign(
          interpolatedStates,
          Interpolate.interpolate(
            transition.endTime, transition.parameter.curve, transition.startTime,
            transition.endTime, transition.transitionStart, transition.transitionEnd,
          ),
        );
        // Remove expired transitions and rewind cursor.
        this.transitions.splice(index, 1);
        index--;
        continue;
      }

      // Calculate interpolated states.
      Object.assign(
        interpolatedStates,
        Interpolate.interpolate(
          currentTime, transition.parameter.curve, transition.startTime,
          transition.endTime, transition.transitionStart, transition.transitionEnd,
        ),
      );
    }

    this.setStates(interpolatedStates as StateValues);
  }
```

I like this for a couple of reasons, but mainly for overall loop reduction. One call to `this.setStates` at the end, so only one loop through defined properties. All logic contained inside a single loop instead of filtering expired, doing something special, then reassigning non-expired, looping again…just remove things that should be removed as we go, adjust the cursor to offset, and continue.

Thoughts?

tsortwehttam (approved):

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -1086,6 +1100,26 @@ export default class HaikuComponent extends HaikuElement {
...
+   * Execute state transitions.
+   */
+  tickStateTransitions(): void {
+    this.stateTransitionManager.tickStateTransitions();
+  }
+
+  /**
+   * Reset states to initial values by using State Transitions. Default to linear
+   */
+  resetStatesToInitialValuesWithTransition(duration: number, curve: Curve = Curve.Linear) {
```
stristr, line 1108: Is this used anywhere?

packages/@haiku/core/src/StateTransitionManager.ts
```diff
@@ -0,0 +1,116 @@
...
+
+      // Set interpolated values
+      this.setStates(interpolatedStates as StateValues);
+    }
+  }
+
+  /**
+   * Delete every running transition
+   */
+  deleteAllStateTransitions() {
```
stristr, line 129: Also unused except in tests. Is there a way to make tests more repetitive to restrict the `src/` code to only include runtime-required functionality?

packages/@haiku/core/src/HaikuComponent.ts
jonaias: This was a request made during team meeting:
"2. Option to reset values on loop-end with transition: It’d be interesting to have an option at the end of the timeline to “auto-animate elements to starting values over a specified number of ms”. Maybe it’d be “not part of the timeline”. It's a convenience for users who don't want to worry about matching myriad starting values up for seamless looping, and It’d certainly make most of our auto-looping share-pages look a lot nicer."

From what I understood, it might be used by @matthewtoast current work on controlFlow.repeat

packages/@haiku/core/src/StateTransitionManager.ts
jonaias: dito

packages/@haiku/core/src/HaikuComponent.ts
stristr: Ah, cool, thanks for clarifying.

packages/@haiku/core/src/StateTransitionManager.ts
stristr: :+1:

packages/@haiku/core/src/api/Curve.ts
```diff
@@ -0,0 +1,46 @@
...
+    EaseInOutQuint = 'easeInOutQuint',
+    EaseInOutSine = 'easeInOutSine',
+    Linear = 'linear',
+  }
+
+/** 
+ * Defines a normalized curve, to be used in BytecodeTimelineValue and also in 
+ * state transition
+ */ 
+export type CurveFunction = ((percent :number) => number);
```
stristr: Nit: a curve function should be `(time: number) => number`. The values we pass in aren't actually percents.

stristr: Also: can we just add the bezier notation `[number, number, number, number]` now?

It already is supported in `just-curves`…using `vendor/just-curves/internal/cubicBezier`.

packages/@haiku/core/test/unit/stateTransitions.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    varObject:{varString: 'string', var: 5},
+    varNull:null,
+    varBool:true,
+  };
+  const haikuClock = new HaikuClock({}, {});
+  
+  const stateTransitionManager = new StateTransitionManager(states, haikuClock);
+
+
+  stateTransitionManager.createNewTransition({var1:10, var2:10}, {duration: 4000, curve: 'Linear' as Curve});
```
stristr: I think in this test, you meant to use `Curve.Linear`, no?

packages/haiku-formats/src/exporters/curves.ts
```diff
@@ -323,7 +323,7 @@ export const decomposeCurveBetweenKeyframes = (timelineProperty:BytecodeTimeline
   const getKeyframe = (normalizedTime: number) => Math.floor(denormalizeValue(normalizedTime, inKeyframe, outKeyframe));
   const getValue = (normalizedPosition: number) => denormalizeValue(normalizedPosition, from as number, to as number);
 
-  getBezierBreakpointsForDecomposableCurve(curve)
+  getBezierBreakpointsForDecomposableCurve(curve as Curve)
```
stristr: Why do we need to typecast here with `as`? It seems to indicate the typing isn't quite right…

Would you mind unwinding this a bit? Seems like we shouldn't have to given that `BytecodeTimelineProperty` is now pointing to `CurveFunction`….

packages/@haiku/core/src/api/Curve.ts
stristr: (Also, was just looking through…we're calling this structure `InterpolationPoints` in `haiku-formats`. It's okay if you want to leave this for now.)

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -332,16 +339,23 @@ export default class HaikuComponent extends HaikuElement {
...
+    if (!states || typeof states !== 'object') {
       return this;
     }
-    for (const key in states) {
-      this.set(key, states[key]);
+
+    // If has transition parameter, it should be treated as a transition
+    if (transitionParameter) {
+      this.stateTransitionManager.createNewTransition(states, transitionParameter);
+    } else {
```
stristr: Final thing: I think here, we _need_ to blow out the transition queue for all state names provided here. In other words, if I call `component.setState({foo: 15})` 500ms after calling `component.setState({foo: 1000, {duration: 1000, curve: 'linear'})`, the in-progress transition (and any queued after it) should be canceled more or less immediately. (Maybe raise in #product if you strongly disagree to get others' take on this.)

My preferred way to handle would just be to put everything through the `StateTransitionManager`, treating this case as an "immediate" transition with queueing disabled.

packages/@haiku/core/src/api/Curve.ts
jonaias: Well, i think it might be percent, because we cant do an interpolation only with current time.. we would need start and end time to a function like that. 

I also got to this conclusion by looking at Interpolate::interpolateValue
``` typescript
function interpolateValue(v0: number, v1: number, t0: number, t1: number, tnow: number, curve: CurveFunction) {
  let pc = percentOfTime(t0, t1, tnow);
  if (pc > CENT) {
    pc = CENT;
  }
  if (curve) {
    pc = curve(pc);
  }
  return valueAtPercent(v0, v1, pc);
}
```

stristr: Oh, I think those are weird variable/function names too :smile:

If you look at the signatures, they return a number in `[0, 1]`, not a number in `[0, 100]` (which is technically what a percent is).

Anyway, just a nit, you can leave it all 😸 👻 🔢

jonaias: Yep, you are right.. percent should be [0,100]

Just found that just-curves use offset as name for it.. adopting ehhehe

packages/@haiku/core/src/HaikuComponent.ts
jonaias: Yep, My plan was to ask on #product after merging this PR. But it makes sense what you explained and as we can easily change it afterwards (It won't break StateTransitionManager api), I decided to implement so we can merge feature complete.

packages/@haiku/core/test/unit/stateTransitions.test.ts
jonaias: :+1:

packages/haiku-formats/src/exporters/curves.ts
jonaias: The CurveFunction is there to allow setting a custom curve on state transition.

About typecasting, I also don't like typecast and definitely they are a code smell . When I finished HaikuBytecode, I left curve type as `any` because I was aware of these divergences on `haiku-formats`..  There are some places that uses same `BytecodeTimelineValue`, but with a different curve types. 

As `haiku-formats` is now set to `noImplicitAny` to tsconfig.json, it will complain. The easiest way is to revert this and leave `curve: any` on HaikuBytecode. Another option (the most correct) would be implement another  `BytecodeTimelineValue` clone with different curve type, to be used exclusively on `haiku-formats`.

jonaias:
Hey Sasha, thanks for your patience.. We end up having a way more flexible state transitions, including queued state transition thanks to you.

packages/haiku-formats/src/exporters/curves.ts
stristr: Yeah, I think there is a better way–which is to fix all internal APIs to match what we want to allow on our public API.

No need to fix this now, but I don't want to lose sight of it, so I added a TODO commit to this branch mentioning that we need to fix this sometime soon.

stristr (approved):
LGTM, merge when ready!

PR #470, opened 2018-05-10
New project modal from anywhere

roperzh:
OK to merge.

Short review.

Summary of changes:

- Add a new shortcut to create a project: `⌘ + n`.
- Allow the new project modal to be invoked from the project detail page (both from the top menu and with `⌘ + n`).
- Fix a [bug](https://app.asana.com/0/670246721316721/629632357785540) regarding max limit on project names and duplicated projects.
- Added an intermediary step between creating the project on inkstone and setting the project locally in order to properly display errors coming from inkstone.

[Here's a video of the new workflow ](https://github.com/HaikuTeam/mono/files/2023217/newproject.mov.zip)

Regressions to look for:

- New project creation, duplication and fork.

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [x] Ran `$ yarn lint-all` and fixed all formatting issues
- [x] Ran `$ yarn test-all` and all tests passed
- [ ] Added measurement instrumentation (Mixpanel, etc.)
- [ ] Wrote an automated test covering new functionality

PR #485, opened 2018-05-21
fix: make new projects public by default

roperzh:
OK to merge.

Short review.

Notes for reviewers:

Ideal scenario would be to merge #470 which also includes this fix, but it is way more dangerous (it touches a lot of code related to creating/duplicating/forking projects).

As you can see in the diff, this is a very small change so I created a separate PR in case we want to merge ASAP and include in the next release.

Completed checkin tasks:


stristr:
LGTM, merging.

PR #488, opened 2018-05-22
Cache File.readMana

tsortwehttam:
OK to merge.

Short review.

Summary of changes:

- Break off separate `Cache` class so usage can occur outside of `BaseModel`
- Add cache to the `File.readMana` call. As `File.readMana` normally (1) reads from disk, (2) parses XML, and (3) performs `svgo` pre-pass, the savings here are considerable. A cache hit here only incurs the cost of cloning the pre-processed mana object.

Regressions to look for:

- Merging design changes on stage
- Instantiating things
- Transform-related actions that depend on the former `cacheFetch` super method

Completed checkin tasks:


stristr:
RECHECK!

stristr:
Note: checkstyle results aren't being published for some reason. But this branch has lint errors.

tsortwehttam:
It's not actually a lint error, it's a problem with the lint fixer. It's doing this:

```
             if (relpath !== this.getActiveComponent().fetchActiveBytecodeFile().relpath) {
-              return
+
             }
```

For some reason it doesn't like that empty return and tries to replace it with a blank line.

stristr:
Wack. Excited to bring in `tslint` and make this issue irrelevant.

stristr:
This LGTM. Do you want it in the RC? If so, we should merge to `rc-3.3`.

PR #489, opened 2018-05-23
Regression: No error when changing the name of a state referenced in an expression

jonaias:
OK to merge.

Short review.

Summary of changes:

- …

Regressions to look for:

- …

Notes for reviewers:

- …

Completed checkin tasks:

- [X] Did manual testing of interrelated functionality

### DIFF
--- a/packages/haiku-common/src/sustained-checker/SustainedWarningChecker.ts
+++ b/packages/haiku-common/src/sustained-checker/SustainedWarningChecker.ts
@@ -4,11 +4,7 @@
  * 
  * TODO: Move this file to haiku-creator
  */
-
-// TODO: component constructor parameter is a HaikuComponent, but haiku-common 
-// is compiled first than haiku-core, giving a compilation error. Fix it when
-// moving this file to haiku-creator
-// import HaikuComponent from '@haiku/core/lib/HaikuComponent';
+import HaikuComponent from '@haiku/core/lib/HaikuComponent';
 
 
 export enum SustainedWarningKind {
@@ -35,7 +31,7 @@ export default class SustainedWarningChecker {
   // checking warning types for method like isIdentifierMissing() )
   private identifierNotFoundWarnings: SustainedWarning[] = [];
   
-  constructor(private readonly component) {}
+  constructor(private readonly component: HaikuComponent) {}
 
   /**
    * Transverse injected HaikuBytecode and check if all injected 
@@ -60,7 +56,7 @@ export default class SustainedWarningChecker {
             if (typeof maybeFunc.value === 'function') {
               const funcParams = maybeFunc.value.specification.params;
               
-              funcParams.forEach((param) => {
+              funcParams.forEach((param: string) => {
                 // If function parameter param is not found in injectables list, create a 
                 // SustainedWarning with type SustainedWarningKind.IdentifierNotFound
                 if (injectables.indexOf(param) < 0) {
--- a/packages/haiku-common/src/sustained-checker/SustainedWarningChecker.ts
+++ b/packages/haiku-common/src/sustained-checker/SustainedWarningChecker.ts
@@ -1,11 +1,15 @@
 /**
  * @file Check and store SustainedWarning from HaikuComponent 
  *       (eg. identifier not found on an expression)
+ * 
+ * TODO: Move this file to haiku-creator
  */
 
+// TODO: component constructor parameter is a HaikuComponent, but haiku-common 
+// is compiled first than haiku-core, giving a compilation error. Fix it when
+// moving this file to haiku-creator
+// import HaikuComponent from '@haiku/core/lib/HaikuComponent';
 
-import HaikuComponent from './../HaikuComponent';
-import {BytecodeTimelines} from './../api/HaikuBytecode';
 
 export enum SustainedWarningKind {
     IdentifierNotFound,
@@ -31,7 +35,7 @@ export default class SustainedWarningChecker {
   // checking warning types for method like isIdentifierMissing() )
   private identifierNotFoundWarnings: SustainedWarning[] = [];
   
-  constructor(private readonly component: HaikuComponent) {}
+  constructor(private readonly component) {}
 
   /**
    * Transverse injected HaikuBytecode and check if all injected 
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -1138,19 +1138,24 @@ export default class Creator extends React.Component {
 
     // Check sustained warnings.
     this.getActiveComponent().on('sustained-check:start', () => {
+      const activeComponent = this.getActiveComponent()
+
+      // If activeComponent is null, delete any identifiersNotFound notice and skip it
+      if (!activeComponent) {
+        this.deleteIdentifierNotFoundNotice()
+        return
+      }
+
       // Check sustained warnings
-      this.getActiveComponent().checkSustainedWarnings()
+      activeComponent.checkSustainedWarnings()
 
       // Get current indenfiers not found
-      const currentIdentifiersNotFound = this.getActiveComponent().sustainedWarningsChecker.notFoundIdentifiers
+      const currentIdentifiersNotFound = activeComponent.sustainedWarningsChecker.notFoundIdentifiers
 
       // If changed, delete current notice and display a new notice if num identifier not found > 0
       if (!lodash.isEqual(currentIdentifiersNotFound, this.identifiersNotFound)) {
         // Delete old notice
-        if (this.identifiersNotFoundNotice) {
-          this.removeNotice(undefined, this.identifiersNotFoundNotice.id)
-          this.identifiersNotFoundNotice = undefined
-        }
+        this.deleteIdentifierNotFoundNotice()
 
         // Create new notice if has any not found indentifier
         if (currentIdentifiersNotFound.length > 0) {
@@ -1405,6 +1410,9 @@ export default class Creator extends React.Component {
   }
 
   teardownMaster ({shouldFinishTour, launchingProject = false}, cb) {
+    // Delete identifier not found notice on teardown
+    this.deleteIdentifierNotFoundNotice()
+
     // We teardownMaster FIRST because we want to close the websocket connections before
     // destroying the webviews, which leads to EPIPE/"not opened" crashes.
     // Previously we were relying on dropped connections to deallocate websockets,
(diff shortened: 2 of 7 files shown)


jonaias:
@matthewtoast, I implemented basically using the ideas you gave. 

Instead of creating property sustainedWarnings in ActiveComponent, I opted to create a class (`packages/@haiku/core/SustainedWarningChecker`), so in the future we might use in raw code editor all sustained warnings.

I had to use HaikuComponent (instead of ActiveComponent) as a constructor parameter because it has `getInjectables()` method. A positive side effect is that we can use `SustainedWarningChecker` on core.

About when to check, your suggestion was https://github.com/HaikuTeam/mono/blob/master/packages/haiku-serialization/src/bll/ActiveComponent.js#L2707, but the problem is that every process (`haiku-glass`, `haiku-creator` and `haiku-timeline`) would perform this check while necessary only for `haiku-creator`. Also as `SustainedWarningChecker` uses bytecode from HaikuComponent, so a reload is necessary, and is performed only on  `ActiveComponent::performComponentWork` callback, inside `ActiveComponent::deleteStateValue`

The check is done used cached parsed functions, as you suggested.

Is it how you imagined it? How is this direction going?   

TODO:
- [ ] Check event functions (eg click) for deleted states
- [ ] Add tests

PS. The algorithm is suboptimal (traverses whole timelines bytecode property), but as it is called only when deleting states once, I think we can live with that. Maybe the unique way to optimize it would be creating a data structure to map each state to everywhere it is currently used on bytecode.

Screenshot:
screenshot_20180523_034831

tsortwehttam:
The error check and typing for the sustained warning both look solid. Your thinking ahead to reuse this functionality in the raw code editor is also a great idea.

> I opted to create a class (packages/@haiku/core/SustainedWarningChecker) [...]
> Also as SustainedWarningChecker uses bytecode from HaikuComponent, so a reload is necessary

I'm thinking that this code _probably_ belongs in `haiku-serialization/src/bll` instead of in `core`, for three reasons:

1. `core`'s concern is runtime and rendering, but `haiku-serialization`'s concern is the app.
2. `core` should be as lightweight as possible.
3. Given that the check has a logical dependency on `reload`, which is a behavior of `ActiveComponent`, this module should be associated with `haiku-serialization`.

However, I don't have a strong feeling about this. I am open to being persuaded to the other side. I would love to keep this module in TypeScript, for example.

I'm curious if @stristr has any thoughts here.

tsortwehttam:
The trickiest part of this feature implementation is _knowing when the warning needs to be shown, and knowing when it can be dismissed_. Have you thought about how to handle the edge cases for that? For example:

- What if the user manually edits the bytecode (`code/main/code.js`) and deletes a state. How do we detect inside the app that the reference is broken via an action that it didn't know about? Should we put this check in `ActiveComponent#softReload` maybe? Or would that be too bad for perf?

- What if the user sees the warning and edits the expression to fix the problem? (Note: That edit occurs in Timeline process, but the warning in your implementation would have originated in the Creator process.) How do we automatically dismiss the warning in this case?

tsortwehttam:
One other thing: "Check event functions (eg click) for deleted states" -- I feel like this can wait until later. For now, displaying the warning on expressions is most important.

jonaias:
ActiveComponent#softReload sounds like the right place (its is called when deleting a state, renaming a state or external code edit). 

As we talked, to avoid performance issues, an event is emitted inside ActiveComponent#softReload (within a debounce), and is executed on haiku-creator. Now every time a softReload is called ( change state, delete state, external file editing), it triggers a checking expressions for missing identifiers.

A toast is displayed showing all missing identifiers. If it is closed (clicking on got it), it will reappear only if missing identifiers change. If missing identifier is fixed (inside haiku or externally editing), it will update (or close if none missing) toast.

Demo:
https://youtu.be/juUZnLCrRzI

jonaias:
BTW, can we temporally let the implementation "@haiku/core/src/helpers/SustainedWarningChecker.ts" there? We can move it later to `haiku-creator` when it becomes Typescript..

Pretty sad to remove typings from implementation, just to add afterwards.. :cry:

tsortwehttam:
> BTW, can we temporally let the implementation "/src/helpers/SustainedWarningChecker.ts" there?

How about we move it to `haiku-common`? (I'd prefer we keep `core` clean as possible.)

> Demo

Awesome!

tsortwehttam:
When you feel like this is ready to merge, can you...:

- Rebase to master, so the PR only contains your commits
- Run linter and fix formatting

Then I will do a final review before merge. 👍

stristr:
Ping! Is this PR ready for review?

packages/haiku-common/src/sustained-checker/SustainedWarningChecker.ts
```diff
@@ -0,0 +1,109 @@
...
+ * @file Check and store SustainedWarning from HaikuComponent 
+ *       (eg. identifier not found on an expression)
+ * 
+ * TODO: Move this file to haiku-creator
+ */
+
+// TODO: component constructor parameter is a HaikuComponent, but haiku-common 
+// is compiled first than haiku-core, giving a compilation error. Fix it when
+// moving this file to haiku-creator
+// import HaikuComponent from '@haiku/core/lib/HaikuComponent';
```
stristr: @jonaias—the reason this was happening was that the necessary `devDependencies` declaration (`haiku-common` -> `@haiku/core`) was not provided. The compile order is determined by explicit dependencies declared in `package.json`, which are essential when introducing cross-module deps. (It would be in `devDependencies` and not `dependencies` because it's only used for typing here).

Unrelated but relevant: in a separate commit now in `master` (a610fe4a2059d631e3e287659ce7f331c5f04e86) we brought in `@haiku/core` as a regular `dependencies` dependency, so you should now be fine to use the type. You'll also want to re-check because implicit `any` is now disabled in this project.

jonaias: Done!

jonaias:
It appers that on master (as of a2da894df74b0791d08a26d9d2e3b008d335ab0c), when you edit a state on state inspector, it isn't updating ActiveComponent bytecode. Tomorrow I'll investigate. Otherwise, it is ready to review.

BTW, I was always targeting development, should I target master? (fork from master / merge to master instead of fork from master/ merge to development)

tsortwehttam:
@jonaias Correct, the workflow is now to branch from `master` and merge to `master`.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -1133,6 +1136,41 @@ export default class Creator extends React.Component {
       subComponents: this.state.projectModel.describeSubComponents()
     })
 
+    // Check sustained warnings.
+    this.getActiveComponent().on('sustained-check:start', () => {
```
tsortwehttam: We may need to unsubscribe from this event when switching between components, otherwise I think we will end up with multiple subscriptions. For example, what happens if we switch back and forth between the 'Main' component and the 'Foo' component many times?

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -2248,6 +2254,15 @@ class ActiveComponent extends BaseModel {
...
       const timelineTime = this.getCurrentTimelineTime()
       this.$instance = this.createInstance(bytecode, instanceConfig)
+
+      // Sustained warnings checker (eg. injected function identifier not found, etc)
+      this.sustainedWarningsChecker = new SustainedWarningChecker(this.$instance)
+
+      // Use debounce to emit event to trigger sustained warnings check on haiku-creator
+      this.emitDebouncedCheckSustainedWarning = lodash.debounce(() => {
+        this.emit('sustained-check:start')
+      }, CHECK_SUSTAINED_WARNINGS_DEBOUNCE_TIME)
```
tsortwehttam: We may want to specify the options `{leading: false, trailing: true}` here.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -1133,6 +1136,41 @@ export default class Creator extends React.Component {
...
+
+      // Get current indenfiers not found
+      const currentIdentifiersNotFound = activeComponent.sustainedWarningsChecker.notFoundIdentifiers
+
+      // If changed, delete current notice and display a new notice if num identifier not found > 0
+      if (!lodash.isEqual(currentIdentifiersNotFound, this.identifiersNotFound)) {
+        // Delete old notice
+        this.deleteIdentifierNotFoundNotice()
+
+        // Create new notice if has any not found indentifier
```
tsortwehttam, line 1169: **Nit:**

IMHO this comment is maybe unnecessary.

My rule of thumb is: Comments should explain _why_. For example:

Bad:

```
// Assign variable foo to the value 1
var foo = 1

// Increment foo
foo++
```

Good:

```
// I wrote it this way because the number two on my keyboard is broken
var foo = 1
foo++
```

tsortwehttam (changes requested):

tsortwehttam:
> when you edit a state on state inspector, it isn't updating ActiveComponent bytecode

I also see an issue sort of like this. But it's very weird:

- The state IS getting changed on disk (in code/main/code.js)
- The state IS updated in the in-mem bytecode in the different views:

```
// After creating a state "moo" in the state inspector, this value is present:
window.timeline.getActiveComponent().getReifiedBytecode().states.moo
```

- But the state IS NOT included as a reference when you type inside an expression (it's like the state is missing even though it is in the in-mem bytecode.)

packages/haiku-creator/src/react/Creator.js
jonaias: You are right. Created an event 'componentDeactivating' that is emitted before changing active component.

packages/haiku-serialization/src/bll/ActiveComponent.js

packages/haiku-creator/src/react/Creator.js
jonaias: I deleted some useless comments.

packages/haiku-serialization/src/bll/Project.js
```diff
@@ -614,6 +614,12 @@ class Project extends BaseModel {
           Lock.LOCKS.FilePerformComponentWork,
           Lock.LOCKS.ActionStackUndoRedo
         ], () => {
+          // Useful to stop haiku-creator listeners when deactivating ActiveComponent
+          const currentActiveComponent = this.getCurrentActiveComponent()
+          if (currentActiveComponent) {
+            currentActiveComponent.emit('update', 'componentDeactivating')
+          }
```
jonaias, line 621: Is it the right place? I couldn't find a better place to emit componentDeactivating event.

tsortwehttam (approved):

packages/haiku-serialization/src/bll/Project.js
tsortwehttam: This looks like a good place to me.

tsortwehttam:
LGTM, merging.

PR #491, opened 2018-05-24
Envoy error handling

roperzh:
OK to merge.

Short review.

Summary of changes:

As I was implementing Figma API calls trough Envoy, I discovered that we aren't handling errors in promises executed in handlers; if the handler fails there's no way for clients to know what happened.

This chains into the pipeline errors that occur when client-executed methods are invoked.

Regressions to look for:

- Envoy in general

Completed checkin tasks:

### DIFF
--- a/packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
+++ b/packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
@@ -291,7 +291,7 @@ export default class EnvoyClient<T> {
         const success = new Promise<any>((acceptInner: Function, rejectInner: Function) => {
           this.outstandingRequests.set(datagram.id, (data) => {
             if (data && data.error) {
-              return rejectInner(new Error(data.error));
+              return rejectInner(data.error);
             }
 
             return acceptInner(data);
--- a/packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
+++ b/packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
@@ -8,6 +8,8 @@ import {
   EnvoyOptions,
   RequestOptions,
   Schema,
+  ClientRequestCallback,
+  EnvoySerializable,
 } from '.';
 
 import EnvoyLogger from './EnvoyLogger';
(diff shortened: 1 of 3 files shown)


packages/haiku-sdk-creator/src/envoy/index.ts
```diff
@@ -17,7 +17,7 @@ export interface Datagram {
   channel: string;
   method?: string;
   params?: string[];
-  data?: string;
+  data?: any;
```
roperzh: This looks bad, but now reflects what is actually happening in the code, and was hidden because parameters without type declaration are implicitly set to `any`, check out [here](https://github.com/HaikuTeam/mono/blob/db7df2cd542b082d043794b146da45cf270c3c6b/packages/haiku-sdk-creator/src/envoy/EnvoyServer.ts#L177-L185).

packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
```diff
@@ -36,7 +36,7 @@ export default class EnvoyClient<T> {
     this.isConnected = false;
     this.websocket = this.options.WebSocket;
     this.datagramQueue = [];
-    this.outstandingRequests = new Map<string, Promise<void>>();
+    this.outstandingRequests = new Map<string, Function>();
```
stristr: Two things, since I'm doing this throughout codebase as soon as we get this release out:
 - When possible, can we initialize in the class spec instead of constructor? i.e., prefer:

```
private blah = 'foo';
```

over

```
private blah;

constructor() {
  this.blah = 'foo';
}
```

 - Can we use a more specific type than `Function`? It seems like this is a very specific function, of type `(data: EnvoySerializable) => void` (also see below).

packages/haiku-sdk-creator/src/envoy/index.ts
stristr: Nit, but preferred:

```
export type EnvoySerializable = any;

data?: EnvoySerializable;
```

Type aliases are nifty because they allow us to enforce type safety at a high level, even if we're not enforcing it at a low level. (Also, we really should type this, once we know what it is…but this is a step in the right direction, even if we don't want to do the work now.)

packages/haiku-sdk-creator/src/envoy/EnvoyClient.ts
```diff
@@ -288,21 +288,23 @@ export default class EnvoyClient<T> {
         const timeout = this.generateTimeoutPromise(mergedOptions.timeout);
 
         // TODO: Fix Bluebird typing error by removing any
-        const success = new Promise<any>((acceptInner: any) => {
-          this.outstandingRequests.set(datagram.id, acceptInner);
+        const success = new Promise<any>((acceptInner: Function, rejectInner: Function) => {
```
stristr: Nit, these functions do have types…I think in this case it's `acceptInner: (PromiseLike<EnvoySerializable>) => void, rejectInner: (error: any) => void`. Generally trying to avoid the types `Function`, `Object`, and similar.

stristr (commented):
LGTM. I sent some nits, which I tend to do, but I think this is fine to merge.

roperzh:
Thank you very much for the review. Fixed nits, merging.

PR #494, opened 2018-05-24
dev: enable react devtools in development for all views

roperzh:
OK to merge.

Short review.

Summary of changes:

As the title says, this should help a big deal with the visibility of the components, especially for people that is new to the app.

Notes to reviewers:

I switched from `electron-react-devtools` to `electron-devtools-installer` because the latter is more versatile, allowing us to install any extension we want.

screen shot 2018-05-24 at 10 40 50

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
     "dotenv": "^5.0.1",
     "electron": "1.8.2",
     "electron-builder": "20.0.8",
-    "electron-react-devtools": "^0.5.3",
     "electron-rebuild": "^1.6.0",
     "fs-extra": "^4.0.2",
     "git-branch": "0.3.0",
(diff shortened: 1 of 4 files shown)


roperzh:
cc: @taylorpoe 🙌

stristr:
Drive-by: no major complaints because this is bound to the `DEV=1` variable which I never use (preferring to manually open inspector as-needed, and spending more time with my IDE's debugger), _but_ I have noticed major performance issues related to react dev tools.

Other, more important thing: I think this might crash the release, because it will transpile into code that `require()`s a `devDependency`, which won't be included in `node_modules`.

Possible remedy is conditional import…

roperzh:
Thanks! I will send a PR in short fixing the crash somehow.

PR #495, opened 2018-05-24
fix: conditionally import react devtools to avoid crashes in build

roperzh:
OK to merge.

Short review.

Summary of changes:

Conditionally import react devtools to avoid crashes in build.


tsortwehttam:
LGTM, merging.

PR #497, opened 2018-05-25
proposal for allowing project launch to be interrupted with ESC

roperzh:
Needs discussion.

Short review.

Summary:

This is a _possible_ solution to interrupting project launch when pressing the `ESC` key, boils down to:

- If the user presses `ESC` during project launch, set `this.state.launchingProject` to `false` in Creator, and immediately hide the reticulating splines screen.
- After every async step of launching a project, check if `this.state.launchingProject` has been set to `false`, if yes, early return and call `teardownMaster`.

The idea is to check if there are any fundamental flaws with this approach because this code path is delicate, if this idea is good to go, I will clean up and add documentation. 

Caveats:

- "Can only update a mounted or mounting component" may appear for the `Timeline` and `Glass` components because the views are abruptly unmounted, should we worry about this or it's expected? 

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -84,6 +84,7 @@ export default class Creator extends React.Component {
     this.onTimelineUnmounted = this.onTimelineUnmounted.bind(this)
     this.onNavigateToDashboard = this.onNavigateToDashboard.bind(this)
     this.disablePreviewMode = this.disablePreviewMode.bind(this)
+    this.handleProjectLoaderKeyDown = this.handleProjectLoaderKeyDown.bind(this)
     this.clearAuth = this.clearAuth.bind(this)
     this.layout = new EventEmitter()
     this.activityMonitor = new ActivityMonitor(window, this.onActivityReport.bind(this))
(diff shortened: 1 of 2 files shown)


tsortwehttam:
The general approach sounds fine to me, but this part

> "Can only update a mounted or mounting component" may appear

concerns me a bit. Generally it's a sign that we're initiating an async action at the wrong time, or not properly cancelling it/cleaning it up, just crossing our fingers that it doesn't cause a problem. I'd like to make sure we track down what's causing it and make sure the action is handled correctly if not cancelled.

roperzh:
> concerns me a bit. Generally it's a sign that we're initiating an async action at the wrong time, or not properly cancelling it/cleaning it up, just crossing our fingers that it doesn't cause a problem. I'd like to make sure we track down what's causing it and make sure the action is handled correctly if not cancelled.

Yes, it makes sense, but I don't think we can control that, to be more specific these errors are happening due to the `Timeline` and `Glass` webviews being shut down without waiting until they are fully loaded.

A possible option could be to wait until a point in which we consider the webviews are safe to be closed, but feels weird as a user to press `ESC` and wait a couple of seconds for the "reticulating splines" screen to close.

I will take another look because I may be confused and maybe there's a simpler solution.

cc: @stristr do you have any ideas?

roperzh:
I'm going to close this and try to come up with a better idea.

PR #500, opened 2018-05-29
Raw code editor

jonaias:
OK to merge.

Short review.

Summary of changes:

 - Implemented raw code editor. Design spec: https://haiku.quip.com/pLmCAnw3uCzY

Notes for reviewers:

 - Not sure if css transitions (haiku-creator/src/react/components/Stage.js) is implemented in the most clean way.

Completed checkin tasks:


jonaias:
Code editor working:
peek 2018-06-04 02-52

Checking if bytecode can be required before saving.
peek 2018-06-04 02-57

I had a lot of troubles with race conditions while implementing save file. Had to start over more than once. I'm happy with resulting implementation (Much cleaner than I expected)

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,125 @@
...
+      currentEditorContents: '',
+      currentBytecodeErrorString: '',
+      showBytecodeErrorPopup: false
+    }
+  }
+
+  componentWillMount () {
+    if (this.props.projectModel) {
+      // Reload monaco contents on component load (eg. changing active component, loading a new project, ..)
+      this.props.projectModel.on('update', (what, ...args) => {
```
tsortwehttam: We probably need to unsubscribe from this event to prevent multiple subscriptions: Consider what happens when:

1. User opens project A
2. User navigates back to dashboard
3. User opens project B
4. User navigates back to dashboard
5. User opens project A again

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,125 @@
...
+
+    this.state = {
+      currentComponentCode: '',
+      currentEditorContents: '',
+      currentBytecodeErrorString: '',
+      showBytecodeErrorPopup: false
+    }
+  }
+
+  componentWillMount () {
```
tsortwehttam: Should this be `componentDidMount` (to avoid `Cannot call setState on unmounted component` errors)?

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,125 @@
...
+    } catch (error) {
+      this.setState({showBytecodeErrorPopup: true,
+        currentBytecodeErrorString: `${error.name}: ${error.message}`})
+      // If cannot valide it, display to user
+      return
+    }
+
+    // Save contents to file
+    activeComponent.fetchActiveBytecodeFile().flushContentFromString(currentEditorContents)
+    // Force module reload
```
tsortwehttam: Can you document why forcing a reload is needed here?

packages/haiku-creator/src/react/components/CodeEditor/MonacoEditor.js
```diff
@@ -0,0 +1,163 @@
...
+import Palette from 'haiku-ui-common/lib/Palette'
+
+// monaco is on global namespace as it uses vscode loader scheme
+// import * as monaco from 'monaco-editor';
+
+class MonacoEditor extends React.Component {
+  constructor (props) {
+    super(props)
+    this.containerElement = undefined
+    this.__current_value = props.value
```
tsortwehttam: Nit: Rename to `this.currentValue`, `this.preventTriggerChangeEvent`?

packages/haiku-creator/src/react/components/CodeEditor/MonacoEditor.js
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Update monaco editor dimensions
+   */
+  updateDimensions () {
+    console.log('Update dimensions')
+    this.editor.layout()
+  }
+
+  destroyMonaco () {
+    if (typeof this.editor !== 'undefined') {
```
tsortwehttam: Nit: This could probably just be `if (this.editor) { ... }`

packages/haiku-creator/src/react/components/StageTitleBar.js
```diff
@@ -561,10 +576,10 @@ class StageTitleBar extends React.Component {
             ]}>
             <ComponentIconSVG color={this.getConglomerateComponentButtonColor()} />
           </button>
-          : ''}
+        }
 
-        {this.isEventHandlersEditorAvailable()
-          ? <button
+        {this.isEventHandlersEditorAvailable() &&
```
tsortwehttam, line 581: 👍

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -2089,8 +2089,10 @@ class ActiveComponent extends BaseModel {
   reload (reloadOptions, instanceConfig, cb) {
     const runReload = (done) => {
       if (reloadOptions.hardReload) {
+        console.log('HardReload on ', this.project.getAlias(), ', file:', this.fetchActiveBytecodeFile().relpath)
```
tsortwehttam: If you want to keep this log, use `logger.info(...)` here instead. (We might want to remove it for `softReload` since that happens a lot.)

packages/haiku-serialization/src/bll/ModuleWrapper.js
```diff
@@ -357,6 +357,55 @@ ModuleWrapper.doesRelpathLookLikeInstalledComponent = (relpath) => {
...
+  filename = filename || ''
+
+  opts.appendPaths = opts.appendPaths || []
+  opts.prependPaths = opts.prependPaths || []
+
+  if (typeof code !== 'string') {
+    throw new Error('code must be a string, not ' + typeof code)
+  }
+
+  var paths = Module._nodeModulePaths(path.dirname(filename))
```
tsortwehttam: Nit: Avoid all use of `var` and please use `let`/`const` instead.

packages/haiku-serialization/src/bll/ModuleWrapper.js
```diff
@@ -357,6 +357,55 @@ ModuleWrapper.doesRelpathLookLikeInstalledComponent = (relpath) => {
   return parts[0] === '@haiku'
 }
 
+/**
+ * Enable loading module from string. Used to check if currently editing file can be required.
+ */
+ModuleWrapper.requireFromString = (code, filename, opts) => {
```
tsortwehttam: We have very similar functionality written here: https://github.com/HaikuTeam/mono/blob/master/packages/haiku-creator/src/react/components/ProjectPreview.js#L14 (Please especially note L23-25) Maybe we should consolidate the other logic to use your method?

tsortwehttam (changes requested):

tsortwehttam:
@jonaias I'm going to review the functionality locally. If I find bugs I will just file them in Asana instead of in here, ok?

tsortwehttam:
Did some brief, general testing of the feature. The full team should QA it this week, but it works pretty well for me 👍 It's really cool to (e.g.) edit values in the timeline and see them change in the code, and vice-versa. 🎆 I filed a few small bugs in Asana.

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
jonaias: :+1:

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
jonaias: You are right. Fixing it.

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
jonaias: Nice spot. This was one of race conditions while switching currently active component.

packages/haiku-creator/src/react/components/CodeEditor/MonacoEditor.js
jonaias: :+1:

packages/haiku-creator/src/react/components/CodeEditor/MonacoEditor.js
jonaias: :+1:

packages/haiku-serialization/src/bll/ActiveComponent.js
jonaias: I intentionally left there because as softReload and hardReload are *very* expensive functions, always printing them make pretty clear they are being executed, and in the future might bother someone to remove some unnecessary reload calls. 

About softReload, I'd also like them.

jonaias: BTW, someone talked about some logging facilities on team call.. Something along the lines of android log tags `Log.v("tag", "message");` (they can be enable/disable according to tag) would be very helpful to debug and develop. Just a thought..

packages/haiku-serialization/src/bll/ModuleWrapper.js
jonaias: The 'your' is too much ehheheh I copied from https://github.com/floatdrop/require-from-string, left the comment there to honor the correct author. Yep, we might use this method instead.

I'll test if everything works correctly.

packages/haiku-serialization/src/bll/ModuleWrapper.js
jonaias: :+1:

jonaias:
Using requireFromString to load project miniatures works!.  I rebased and cleaned commit history.

packages/haiku-serialization/src/bll/ModuleWrapper.js
```diff
@@ -357,6 +358,62 @@ ModuleWrapper.doesRelpathLookLikeInstalledComponent = (relpath) => {
   return parts[0] === '@haiku'
 }
 
+/**
+ * Enable loading module from string. Used to check if currently editing file can be required.
+ * Heavily based on https://github.com/floatdrop/require-from-string
+ */
+ModuleWrapper.requireFromString = (code, filename, opts) => {
```
stristr: See also `packages/haiku-creator/src/react/components/ProjectPreview.js`. Can we consolidate here?

yarn.lock
```diff
@@ -7957,6 +7957,10 @@ monaco-editor@^0.12.0:
   version "0.12.0"
   resolved "https://registry.yarnpkg.com/monaco-editor/-/monaco-editor-0.12.0.tgz#cd8621017526b57746245104d764bbf52ad42283"
 
+monaco-editor@^0.13.1:
+  version "0.13.1"
+  resolved "https://registry.yarnpkg.com/monaco-editor/-/monaco-editor-0.13.1.tgz#6b9ce20e4d1c945042d256825eb133cb23315a52"
+
```
stristr: Looks like this PR brings in an additional copy of `monaco-editor`. Can you please be sure to align the dependency version? This could either be by downgrading the version you added to `haiku-creator` or by upgrading the version in `haiku-glass` (both probably require testing if there are breaking API changes).

jonaias: I took a look at monaco-editor changelog.. We don't use any of breaking changes. Updated it.

packages/haiku-serialization/src/bll/ModuleWrapper.js
jonaias: Sorry I didn't follow. By consolidating you mean using this same method on `packages/haiku-creator/src/react/components/ProjectPreview.js`? 

If so, Matthew also mentioned it, implemented in aa7c0efe96b1d12b792d897c0ed4a673f8f3efbb.

stristr: Yep! Sorry, missed that.

tsortwehttam (approved):

tsortwehttam:
Code LGTM, merging into `rc-3.5.0`.

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,136 @@
...
+      ModuleWrapper.testLoadBytecode(currentEditorContents, absPath)
+    } catch (error) {
+      this.setState({showBytecodeErrorPopup: true,
+        currentBytecodeErrorString: `${error.name}: ${error.message}`})
+      // If cannot valide it, display to user
+      return
+    }
+
+    // Save contents to file
+    activeComponent.fetchActiveBytecodeFile().flushContentFromString(currentEditorContents)
```
stristr, line 87: Will need to fix.

 - Leaky abstraction—no way should a content flush be happening from React.
 - More importantly, nothing other than `Master`-controlled AC should be triggering a writeout because the filesystem lock used for all other writeouts would not apply here. We'll need to handle this via IPC.

PR #504, opened 2018-05-30
Bring back the public/private toggle to the share modal

roperzh:
OK to merge.

Short review.

Summary of changes/notes:

~As the title says, this brings back the public/private toggle to the publish modal, but it doesn't switch licenses when toggling.~

~I remember that this topic was discussed in the past, and maybe it was one of the reasons behind moving the public/private toggle to the new project modal. In light of this, should I bring back those changes too? cc: @matthewtoast @zackbrown~

Removing `wip` label, this was discussed in Slack, in general lines: it's not a good idea to handle the license switching in the app, so this is good to go now.

Completed checkin tasks:


roperzh:
Removing `wip` label, updated original post, this is good to review + merge now.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
```diff
@@ -4,6 +4,8 @@ import {RevealPanel} from '../RevealPanel';
 import {ProjectShareDetails} from './ProjectShareDetails';
 import {EmbedList} from './EmbedList';
 import {EmbedDetails} from './EmbedDetails';
+import {Project} from 'haiku-sdk-creator/lib/bll/Project';
+import {inkstone} from '@haiku/sdk-inkstone';
```
stristr, line 8: This dependency is not actually available to `haiku-ui-common` :smile: Pitfall of yarn workspaces…the dependency resolves locally, but isn't actually supposed to be available.

And…this is a good thing. As with Figma, we need to isolate API calls on the back end, both because proxy settings require it and because it's a good practice. Good candidate for the new Services channel! (Please merge in latest from `typescriptification` before you get into `haiku-sdk-creator` if that's the approach you take.)

stristr: Wait, ignore everything I just said since you're not actually using it except as a devDependency, and the call is already in Envoy. Sorry for not reading down before posting!

So: to fix, just make sure `@haiku/sdk-inkstone` is a `devDependency`.

stristr: (I made this fix, since I'm about to upgrade this module.)

stristr:
LGTM, merging.

packages/haiku-ui-common/src/react/ShareModal/index.tsx
roperzh: You are awesome, thank you!

PR #511, opened 2018-06-01
perf: avoid re-rendering the whole timeline twice every click

roperzh:
OK to merge.

Short review.

Summary of changes:

Doing research on timeline perf, I noticed that the whole timeline
is rendered twice on every click.

This is caused because we are setting state on mousedown and mouseup,
and setting styles in one of the root elements based on this.

This changes the logic to modify the styles via plain old JS and
use attributes in the component itself instead of states.

Regressions to look for:

As @stristr mentioned in the past, some bugs may be hidden because we are re-rendering the timeline way too many times. After implementing this, I have found one totally unrelated bug, which is also solved. (see comment below)

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/RowSegments.js
+++ b/packages/haiku-timeline/src/components/RowSegments.js
@@ -95,7 +95,7 @@ export default class RowSegments extends React.Component {
                   keyframe={keyframe} />
               )
             }
-            if (keyframe.isSoloKeyframe()) {
+            if (keyframe.isSoloKeyframe() || !keyframe.hasNextKeyframe()) {
               segmentPieces.push(
                 <SoloKeyframe
                   id={`keyframe-${keyframe.getUniqueKey()}-${this.props.scope}-solo-keyframe`}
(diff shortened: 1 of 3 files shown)


packages/haiku-timeline/src/components/RowSegments.js
```diff
@@ -95,7 +95,7 @@ export default class RowSegments extends React.Component {
                   keyframe={keyframe} />
               )
             }
-            if (keyframe.isSoloKeyframe()) {
+            if (keyframe.isSoloKeyframe() || !keyframe.hasNextKeyframe()) {
```
roperzh, line 98: This fixes a bug when selecting the last keyframe of a row that was hidden before because the timeline was re-rendered on mouseup.

roperzh:
BUILDME!

stristr:
I put this through the timeline QA process on a fresh project, then clicked around on `metapoem2` and found no regressions. LGTM and great find. Merging.

PR #515, opened 2018-06-04
Proposal: replace radium with emotion

roperzh:
OK to merge.

Short review.

Summary of changes:

We have discussed moving away from Radium many times, and I think our long term goal is to move all the CSS away from the JS into CSS modules.

This proposal is a step forward in that direction, which allows us to move fast into a more CSS oriented timeline, without having to rewrite/move a lot of code.

`emotion` is a so-called CSS-in-JS library, it exposes a function (`css`) which takes almost anything (including objects, which fits well for us), and converts that into CSS which is injected via a style tag into the head, and returns a class name.

While this adds a bit of JS scripting time as well, I think it's a step forward because it has some advantages:

- Styles are injected only once, not re-calculated every time the component is rendered
- All of CSS is supported, without hacks, including pseudo-selectors( `:before`, `:after`, etc) and media queries
- `:hover` and friends are managed by CSS, not more JS hacks.
- As you can see in the diff, very low action is required in our end, this whole migration took me ~15 minutes.

Completed checkin tasks:


packages/haiku-timeline/package.json
```diff
@@ -27,15 +27,20 @@
...
       "react-app"
+    ],
+    "plugins": [
+      ["emotion", { "hoist": true }]
     ]
   },
   "dependencies": {
     "@haiku/core": "3.3.4",
     "archy": "^1.0.0",
+    "babel-plugin-emotion": "^9.1.2",
```
stristr: Important: this should be in `devDependencies`.

roperzh:
*Update*

As discussed in the team meeting:

- Apply styles via `emotion` only to components using Radium
- Use the babel plugin only as a dev-dependency

Notes:

After looking at the diff in `yarn.lock`, knowing that perf gains aren't that good, and knowing that we want to migrate away from babel, I'm a bit hesitant of merging this, but I'll leave the decision into the hands of the reviewer(s).

I briefly looked into CSS modules, but seems like that's part of a bigger issue because we need a bundler such as webpack or similar in place.

tsortwehttam:
LGTM, merging.

packages/haiku-timeline/package.json
```diff
@@ -44,12 +45,12 @@
     "lodash": "^4.17.4",
     "numeral": "^2.0.6",
     "qs": "6.4.0",
-    "radium": "0.18.1",
     "raven-js": "^3.18.1",
     "react": "15.6.2",
     "react-beautiful-dnd": "^6.0.1",
     "react-dom": "15.6.2",
     "react-draggable": "2.2.3",
+    "react-emotion": "^9.1.3",
```
stristr, line 53: Note: I don't think this plugin is actually used in this project. Should it just be removed?

stristr: (and/or can it?)

stristr:
Hmmm, yes, this is weird that `emotion` itself brings in actual babel dependencies as production dependencies. Will take a careful look at this later.

PR #516, opened 2018-06-04
feat: enable public/private to be toggled during publish

roperzh:
OK to merge.

Short review.

Summary of changes:

- The title says all, this was disabled because we needed to switch licenses, which is no
longer the case.

Completed checkin tasks:


tsortwehttam:
LGTM.

PR #520, opened 2018-06-05
Finish queued state transitions behavior implementation

jonaias:
OK to merge.

Short review.

Summary of changes:
The behavior details of queued state transitions was discussed on slacks's #product on May 29th:

1- A 'packed' setState transition,

```
setState({var1: 10, var2: 10}, {duration: 4000, curve: 'linear'}
```

is implemented essentially as a short for

```
setState({var1: 10}, {duration: 4000, curve: 'linear'}
setState({var2: 10}, {duration: 4000, curve: 'linear'}
```
 
2 - If you pass the parameter `queue: true`

```
setState({var1: 10}, {duration: 4000, curve: 'linear', queue:true}
```

It will queue (or execute right away if no other transition is executing on that state) the state transition. If a normal setState is called (without any transition parameter) or a not queue setState(eg. `setState({var1: 10}, {duration: 4000, curve: 'linear'}` or `setState({var1: 10})` ), it will cancel any state transition on that state and execute the non queued state transition. 

3- Transition origin (startTime and trasitionStart) for queued transitions is calculated when transition is about to start

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/StateTransitionManager.ts
+++ b/packages/@haiku/core/src/StateTransitionManager.ts
@@ -19,6 +19,7 @@ export interface RunningStateTransition {
   transitionParameter: StateTransitionParameters;
   startTime: number;
   endTime: number;
+  duration: number;
 }
 
 export default class StateTransitionManager {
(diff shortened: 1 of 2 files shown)


tsortwehttam:
LGTM, merging.

PR #536, opened 2018-06-08
Enable interpolation for boolean only. Fix (703436957231272)

jonaias:
OK to merge.

Short review.

Summary of changes:

Booleans weren't interpolatable(_sic_) in interpolation function. Because setting state was generalized inside  stateTransitionManager, booleans were being skipped.

Additionally, for the hypothetical user who wants to do a boolean interpolation, this patch enables it.. It assumes 0 for false and 1 for true, and if interpolation result is >= 0.5, it is true.

The ideal solution (eg, if we enable number->boolean transition, if we enable boolean -> boolean transition) is discussed on #product on 8th June

Regressions to look for:

- …

Notes for reviewers:

- …

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/StateTransitionManager.ts
+++ b/packages/@haiku/core/src/StateTransitionManager.ts
@@ -34,11 +34,21 @@ export default class StateTransitionManager {
    */
   setState (transitionEnd: StateValues, parameter?: StateTransitionParameters) {
 
+    // If not a transition, execute it right away
+    if (!parameter) {
+      for (const key in transitionEnd) {
+        delete this.transitions[key];
+      }
+      this.setStates(transitionEnd);
+      return;
+    }
+
     // Get current time
     const currentTime = this.clock.getTime();
 
-    // If paramers isn't set, process like a non queued SetState
-    const transitionParameter = parameter ? parameter : {duration:0, curve: Curve.Linear, queue: false};
+    // Set default values than assign parameters
+    const transitionParameter = {duration:0, curve: Curve.Linear, queue: false};
+    Object.assign(transitionParameter, parameter);
 
     // Copy current states as transition start (needed to calculate interpolation)
     for (const key in transitionEnd) {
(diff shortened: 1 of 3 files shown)


packages/@haiku/core/test/unit/stateTransitions.test.ts
```diff
@@ -6,7 +6,7 @@ import * as tape from 'tape';
 tape(
   'Test state transitions',
   (t) => {
-    t.plan(31);
+    t.plan(34);
```
stristr: Suggestion (my other feelings about whether interpolating booleans aside): instead of doing`t.plan(N = number of tests)`, you can just remove this and put `t.end()` at the end of your tests.

(`t.plan()` is only truly necessary when you have an asynchronous test that might not finish.)

jonaias: Thanks for the tip

jonaias:
Another thing that i'd like to discuss is if interpolation cannot be done, if we can return `destination` instead of `origin` if `now >= ends`;

Basically change https://github.com/HaikuTeam/mono/pull/536/files#diff-9d13f95c09501e5b206b02a2629a2bbeR90 to `return now >= ends ? destination : origin;` instead of `return origin;`

packages/@haiku/core/src/StateTransitionManager.ts
```diff
@@ -118,11 +118,13 @@ export default class StateTransitionManager {
           // Remove expired transition.
           this.transitions[stateName].splice(0, 1);
 
-          // Update next queued state transition
+          // Update next queued state transition or delete empty transition vector for performance reasons
```
stristr, line 127: I think there is a better performance fix for the actual issue reported higher up.

```
setState (transitionEnd: StateValues, parameter?: StateTransitionParameters) {
  if (!parameter) {
    for (const key in transitionEnd) {
      delete this.transitions[key];
    }
    this.setStates(transitionEnd);
    return;
  }

  // Get current time
  const currentTime = this.clock.getTime();
```

See what I mean?

stristr:
Overall, I think we just need to simplify the behavior when there is no transition parameter.

I don't have a strong opinion on if/how boolean interpolation should be defined, if we define it at all, so I think the rest is fine. But we should be jumping out of this logic entirely for the straightforward case that is broken in [setState() either doesn't appear to change state](https://app.asana.com/0/687739386153745/703436957231272/f). These will be the most common use cases so it's worthwhile to short-circuit the interpolation logic and save the CPU cycles....

packages/@haiku/core/src/StateTransitionManager.ts
jonaias: Implemented!

jonaias:
The tests got a subtle bug when changed into `Object.assign(interpolatedStates, transition.transitionEnd);` on transition expiration.

PR #542, opened 2018-06-09
Fix write races

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes problem where we commit an empty `code/<scene>/code.js`.
- Shores up other potential weak points.

Regressions to look for:

- Any ungood changes to Git behavior or files not being committed.
- Any possibility of getting infinite locks.

Notes for reviewers:

- This PR is intentionally to `master` (still only good at building from `master` r.n.)

The race resulting in _commits_ of an empty `code.js` this should fix is:
  1. some change A occurs -> file lock granted -> commit A of relpath is pushed into commit queue -> file lock released
  1. some other change occured during above -> file lock granted -> write begins -> file is truncated
  1. commit A is completed -> file write completes -> file lock released

Other changes include:
  1. vacate all locks before Master can be torn down
  1. get file locks before trying to bootstrap scene files
  1. clean up unused `discardProjectChanges`, `doesProjectHaveUnsavedChanges`, and friends.

The gist is: use more I/O locks on bytecode files in places where they weren't before.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ModuleWrapper.js
+++ b/packages/haiku-serialization/src/bll/ModuleWrapper.js
@@ -168,7 +168,7 @@ class ModuleWrapper extends BaseModel {
             this.file.maybeFlushContentForceSync()
           }
           release()
-          return cb()
+          return cb(null, this.exp)
         })
       }
 
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -137,10 +137,7 @@ export default class Creator extends React.Component {
 
     // fileOptions is something of a misnomer, but we pass these into Project so they
     // can be forwarded to file. They're also used to configure the Project model itself.
-    this.fileOptions = {
-      doWriteToDisk: false,
-      skipDiffLogging: true
-    }
+    this.fileOptions = {}
 
     // Callback for post-authentication
     this._postAuthCallback = undefined
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -141,10 +141,7 @@ export class Glass extends React.Component {
       this.props.websocket,
       window,
       this.props.userconfig,
-      { // fileOptions
-        doWriteToDisk: false,
-        skipDiffLogging: true
-      },
+      {}, // fileOptions
       this.props.envoy,
       (err, project) => {
         if (err) throw err
--- a/packages/haiku-plumbing/src/Master.js
+++ b/packages/haiku-plumbing/src/Master.js
@@ -15,7 +15,6 @@ import logger from 'haiku-serialization/src/utils/LoggerInstance'
 import MockWebsocket from 'haiku-serialization/src/ws/MockWebsocket'
 import { EventEmitter } from 'events'
 import EmitterManager from 'haiku-serialization/src/utils/EmitterManager'
-import * as Git from './Git'
 import Watcher from './Watcher'
 import * as ProjectFolder from './ProjectFolder'
 import MasterGitProject from './MasterGitProject'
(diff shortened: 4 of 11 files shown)


tsortwehttam:
Code LGTM.

packages/haiku-serialization/src/bll/ModuleWrapper.js
```diff
@@ -168,7 +168,7 @@ class ModuleWrapper extends BaseModel {
             this.file.maybeFlushContentForceSync()
           }
           release()
-          return cb()
+          return cb(null, this.exp)
```
stristr, line 171: Note: added this for correctness. I don't think it _really_ matters because everything seems to use `this.exp` directly where it matters.

stristr:
(@matthewtoast, added some commits here for your review)

stristr:
Note: in 74e5d9c, I changed a call to `File.upsert({...})` to `this.project.upsertFile` in `ActiveComponent` beneath the comment `// There is a race where the file might not be ready, so we upsert`. Although I don't have any logs to prove it or an indication of exactly how this can occur (it seems like a `File` model would have to somehow be destroyed), my best guess for how commit `38afe7f` happened in @taylorpoe's `tog` project was that this race led a renderer process to have a `File` with `doWriteToDisk` set to `true`. For good measure, I changed the prior setup (`doWriteToDisk` is default `true` and `skipDiffLogging` is `false` by default, and each is manually disabled in each renderer process) to one that made more sense to me (`doWriteToDisk` are default `false` and `skipDiffLogging` is default `true`, and only in main are they switched on/off respectively).

From my mostly static analysis, I feel pretty good about this as our write race explanation despite failing to get a repro after trying for a day. I'm still not fully satisfied because I don't understand when/how the `File` instance is getting destroyed. At worst the bug still exists and this is a noop.

stristr:
Revised thoughts on 74e5d9c—nothing in the logs actually indicates this is what's happening (there's no log line `[info] [file] writing code/main/code.js to disk` except from `main`).

Still a good change IMO, but likely _not_ the explanation we seek.

tsortwehttam:
Code LGTM. Also did a quick review locally.

I do like this theory, but I agree it doesn't quite satisfy without repro.

A few observations that may help shine a light:

This code pathway looks very suspicious:

```
ActiveComponent#upsertSceneByName
  ActiveComponent#upsertComponentBytecodeToModule <~ bytecode arg is null here!
    ActiveComponent#fetchActiveBytecodeFile
    if (!bytecode) {
      file.mod.update({}) <~ empty object is provided here
        Bytecode.reinitialize(bytecode) <~ empty template, base bytecode recreated
    }
```

If `ActiveComponent#upsertSceneByName` is called before the virtual File is ingested, _and_ if we are in a process that writes to disk (master), then we are halfway to explaining the problem.

The other side of the problem is: "Under what circumstances is a disk-write ever requested?" And there are only two circumstances when this happens:

1. `File#requestAsyncContentFlush` - which is only called via `ActiveComponent#handleUpdatedBytecode`, which itself is only called after ActiveComponent#performComponentWork` and `ActiveComponent#conglomerateComponentActual`.

2. `File#maybeFlushContentForceSync` - which is only called if we reach the exception block of the try/catch block in `ModuleWrapper#reload`.

Of those, #2 seems most suspicious. The intent of the logic is to ensure the file is populated if the module is missing, i.e. if this method gets hit before the correct contents are on disk. But it assumes that an error in `ModuleWrapper#load` must be file not found, when it could be something else.

Things that could cause `ModuleWrapper#load` to throw:

- `overrideModulesLoaded(...)` throws
  - ?
- `require()` throws
  - File not found
  - Syntax error
  - ?

stristr:
Yeah, I'm with you, except for understanding under under which circumstances actually would occur. (I'm hopeful the `[active component (${folder})] fetchActiveBytecodeFile called before file was upserted` log will give an additional clue.

Going to merge this, but with less hope than at first that all write races are resolved.

PR #547, opened 2018-06-12
small fixes for actions UI

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes https://app.asana.com/0/680508544654148/700874348545631

- Don't show main as having actions if it only has timeline actions
- Change the shade of blue of the bolt icon

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/Timeline.js
+++ b/packages/haiku-timeline/src/components/Timeline.js
@@ -90,7 +90,6 @@ class Timeline extends React.Component {
     )
 
     this.handleRequestElementCoordinates = this.handleRequestElementCoordinates.bind(this)
-    this.showEventHandlersEditor = this.showEventHandlersEditor.bind(this)
     this.showFrameActionsEditor = this.showFrameActionsEditor.bind(this)
     this.mouseMoveListener = this.mouseMoveListener.bind(this)
     this.mouseUpListener = this.mouseUpListener.bind(this)
@@ -1207,8 +1206,8 @@ class Timeline extends React.Component {
                                 getActiveComponent={() => {
                                   return this.getActiveComponent()
                                 }}
-                                showEventHandlersEditor={() => {
-                                  this.showEventHandlersEditor()
+                                showEventHandlersEditor={(...args) => {
+                                  this.showEventHandlersEditor(...args)
                                 }}
                                 />
                             </div>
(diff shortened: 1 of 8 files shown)


roperzh:
screen shot 2018-06-12 at 13 55 02

I am seeing this ☝️ but the branch is rebased against `rc-3.4.5.2` 🤷‍♂️

stristr:
I think the problem is `rc-3.4.5.2`, which is apparently not rebased against `master`…will check later.

roperzh:
Update: pushed a commit  to also fix https://app.asana.com/0/680508544654148/697640433446165

stristr:
LGTM (so far)! Except for bad Git history, but let's just squash-merge this and really fix in `rc-3.4.5.2`.

roperzh:
Update II: pushed another commit to fix https://app.asana.com/0/680508544654148/704795620573338

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -1207,8 +1206,8 @@ class Timeline extends React.Component {
                                 getActiveComponent={() => {
                                   return this.getActiveComponent()
                                 }}
-                                showEventHandlersEditor={() => {
-                                  this.showEventHandlersEditor()
+                                showEventHandlersEditor={(...args) => {
```
stristr, line 1209: I think the fix we want here is simply:

```showEventHandlersEditor={this.showEventHandlersEditor}```

When we typescriptify, I'm adding the `jsx-no-lambda` rule, so we'll actually be forbidding this. In TypeScript, the preferred way will be:

```
private boundShowEventHandlers = (...args) => this.showEventHandlers(...args);
```

In JS, the `.bind(...)` pattern is fine.

Sorry for mixed messages here! We still want to prefer lambda functions, but _not_ create them inside `render()` method for performance reasons.

stristr: (Docs if interested:

> Creating new anonymous functions (with either the `function` syntax or ES2015 arrow syntax) inside the `render` call stack works against _pure component rendering_. When doing an equality check between two lambdas, React will always consider them unequal values and force the component to re-render more often than necessary.

)

stristr:
Merging.

packages/haiku-timeline/src/components/Timeline.js
roperzh: kk sounds good! 👍

PR #552, opened 2018-06-13
Hack item scaling to prevent zero scale.

ghost:


ghost:
Not really sure why all those health tests are failing. Linting passes on my machine...

stristr:
There's a lint error (you can click through _Details_ to see it).

One `yarn lint-all` from `mono` root should fix it, or one `yarn fix` from `haiku-serialization` root. :smile:

stristr:
BUILDME!

stristr:
(The test failing seems to be a flake/false negative that sometimes happens when lint errors are found.)

ghost:
Ah thanks, I didn't expect linting to return 0 and modify my code if there were errors.

So what do you think about the placement of this thing to prevent zero scale? It will negate the early return referenced in the comment, but isn't it true that we never want scale to equal zero because the element will be impossible to interact with, impossible to see, and full of matrix multiplication issues?

packages/haiku-serialization/src/bll/ElementSelectionProxy.js
```diff
@@ -1381,6 +1381,12 @@ ElementSelectionProxy.computeScalePropertyGroup = (
   const targetLayout = element.getComputedLayout()
   // Opportunity to return early if we have a downstream "division by 0" problem. Scaling _from_ 0 is not supported (and
   // the UI should make it impossible.
+
+  // --- hack ---
+  if (targetLayout.scale.x === 0) targetLayout.scale.x = 0.0001
+  if (targetLayout.scale.y === 0) targetLayout.scale.y = 0.0001
+  // --- /hack ---
+
   if (targetLayout.scale.x === 0 || targetLayout.scale.y === 0) {
```
stristr: Note: with hack above, this block is never entered, so could/should just be removed.

stristr (commented):
Exactly.

So, the way this is happening at all is because [here](https://github.com/HaikuTeam/mono/blob/8781f1d1252741413636d92d7fde406bdc98e9f1/packages/haiku-serialization/src/bll/ElementSelectionProxy.js#L1512) we round to three significant digits. This PR does seem to remove the "transient rounded scale of 0" problem. It also has the theoretical side effect of causing the point you're scaling from to not perfectly track the cursor due to micro-offsets adding up over time…but I'm not sure how real of a concern this is.

I think this is fine for now (with change requested below). LGTM.

PR #549, opened 2018-06-12
RC 3.4.5.3

stristr:
For build.

PR #554, opened 2018-06-15
IOC refactor for #549

stristr:
OK to merge.

Should fix known template hash insertion point related races and associated integrity-check crashes in dev.

Mainly simplifies IoC around File upsertion and cleans up the resulting mess—only Project upserts file, unless an ActiveComponent is creating a subcomponent.

Also rolls a nifty util for checking if XML is meaningfully different, which fixes some tests that flaked.

Testing health check before merging to #549.


stristr:
Merging to #549.

PR #563, opened 2018-06-18
Brush Clearing

tsortwehttam:
OK to merge.

Summary of changes:

- Add an E2E test harness based on Spectron.
- Refactor to remove explicit property enumerations which bloat @haiku/core, make support for recognized APIs more labor intensive, and make supporting custom properties (e.g. `controlFlow.repeat`) more convoluted than it needs to be.
- Refactor to allow for property application during rendering to occur in distinct structural and presentational phases such that `controlFlow.repeat` is able to implemented in a straightforward manner.

Completed checkin tasks:


stristr:
This PR hangs `yarn test-report`. Almost certainly due to e2e test stuff attached to it. Confirming/fixing….

stristr:
Code LGTM. Have confirmed distro changes `dom.bundle.min.js`: 215K -> 161K (46K gzipped, down 10K from 56K). Ain't bad!

Assuming you checked demos, so please merge when ready.

packages/@haiku/core/src/Migration.ts
```diff
@@ -179,8 +184,8 @@ export const runMigrations = (component: HaikuComponent, options: any, version:
 
   if (requiresUpgrade(coreVersion, UpgradeVersionRequirement.OriginSupport)) {
     component.visit((element) => {
-      if (schema[element.tagName] && schema[element.tagName]['origin.x']) {
-        // We only need to upgrade elements whose schemas support origin.
+      // We only need to upgrade elements whose schemas support origin.
+      if (ELEMENTS_THAT_SUPPORT_ORIGIN[element.tagName]) {
```
stristr, line 188: Note: technically every "dimensionable" supports origin, which is a much longer list (per the now-deleted schema). This should be fine in terms of legacy migrated bytecode still working…but we do have examples where the migration is now broken. (e.g. `glass-primitives` no longer displays correctly.) Up to you if you want to change this or leave it…I was never very attached to `glass-primitives` displaying correctly since `percy`, `Moto`, and `Move` are all fine.

If you do care, should be fine (+/- migration-time perf for some now very-old projects) to just remove the conditional entirely.

PR #551, opened 2018-06-13
feats: include Figma components and ids for all assets to facilitate ungroup

roperzh:
⚠️ OK to merge when #550 is merged.

Short review.

Summary of changes:

- Include Figma components in the `/groups` folder ([task](https://app.asana.com/0/680508544654148/704595780151982))
- Include a parameter when requesting SVGs to include id attributes for all SVG elements, this should help with ungroup if I'm right (cc: @stristr)

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [x] Wrote an automated test covering new functionality

PR #550, opened 2018-06-12
feat: include Figma frames during import

roperzh:
OK to merge.

Short review.

Summary of changes:

What the title says, related task: https://app.asana.com/0/680508544654148/704595780151981

I decided to include this in rc-3.5.0 to keep the hotfix release as lean as possible.

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [x] Wrote an automated test covering new functionality

PR #535, opened 2018-06-07
Fix When project creation fails, new project form is not retracted 

jonaias:
https://app.asana.com/0/687739386153745/693868196194064

OK to merge.

Short review.

Summary of changes:

- …

Regressions to look for:

- …

Notes for reviewers:

- …

Completed checkin tasks:

- [ ] Did manual testing of interrelated functionality
- [ ] Added measurement instrumentation (Mixpanel, etc.)
- [ ] Wrote an automated test covering new functionality

PR #532, opened 2018-06-07
Better Commit Messages (682873242829705) [for RC-3.5.0]

tsortwehttam:
OK to merge into `rc-3.5.0`.

Short review.

Summary of changes:

- Have commit messages be more descriptive.

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality

PR #566, opened 2018-06-18
RC 3.5.0 (redux)

stristr:
Not ready to merge yet, but ready for testing.

Medium review.

Summary of changes:

- Raw code editor (#500)
- Replace `radium` with `emotion` in timeline (#515)
- Show an error when changing the name of a state referenced in an expression (#489)
- Add RCM option in timeline to "send to t0" (#528)
- More descriptive commit messages (#532)
- Retract new project form when project creation fails (#535)
- Include Figma frames in import (#550)
- Include Figma components, and include IDs on all imported SVGs (#551)
- [refactor] Experiments are either always on or always off

Regressions to look for:

- TBD.

Completed checkin tasks:


stristr:
BUILDME!

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,145 @@
...
+      ModuleWrapper.testLoadBytecode(currentEditorContents, absPath);
+    } catch (error) {
+      this.setState({showBytecodeErrorPopup: true,
+        currentBytecodeErrorString: `${error.name}: ${error.message}`});
+      // If cannot valide it, display to user
+      return;
+    }
+
+    // Save contents to file
+    activeComponent.fetchActiveBytecodeFile().flushContentFromString(currentEditorContents);
```
stristr: @jonaias: this introduces a potential write race that was turned into an error since this code was written—if `doWriteToDisk` is not enabled, `File#write` will throw an error (`[file] illegal write requested`) and not write. So as of now, the Code Editor is not actually writing to disk.

To fix, we should use the standard `Lock`s and ensure that writing is done in the Main process.

Please look into fixing this ASAP, since it blocks 3.5.0 (unless we disable the raw code editor in 3.5.0).

packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -0,0 +1,145 @@
...
+    // This module reload is necessary because when switching active component (eg. component
+    // tab clicking), the file is saved in the disk but contents aren't loaded into memory. By forcing
+    // a module reload, we guarantee that the file is synced with memory
+    activeComponent.moduleReplace(() => {});
```
stristr: In general, we should avoid the "leaky abstraction" of having this React component handle so much important bytecode-specific logic.

A better way to write this entire block would be:

```
const activeComponent = this.props.projectModel.getCurrentActiveComponent();
if (!activeComponent) {
  return;
}

activeComponent.replaceBytecode(this.state.currentEditorContents, (err) => {
  if (err) {
    // (handle error)
  }
});
```

The `ActiveComponent#replaceBytecode` method should maintain responsibility for:
 - Wrapping everything in `this.project.updateHook` so the net effect of changes is synced across processes.
 - Making sure the bytecode is require'able.
 - Calling `handleUpdatedBytecode` with the valid result of `require`'ing the bytecode (you don't need to reload from disk for this, which is what `moduleReplace()` does). Note that this _will_ flush to disk—but only from the main process, which will receive the `updateHook` from `Project` as desired.
 - Note that as long as there are no errors, you can safely assume that everything made it onto disk okay, but if you want, you can subscribe to a special method hook from `main` so you know the actual save took.

stristr:
BUILDME!

stristr:
I have reviewed and tested this PR and, except for issues with bytecode writeout noted in comments above, this LGTM.

stristr:
BUILDME!

stristr:
LGTM, merging.

PR #568, opened 2018-06-19
Fix code editor save to file

jonaias:
WIP.

Short review.

Summary of changes:

- …

Regressions to look for:

- …

Completed checkin tasks:

- [ ] Did manual testing of interrelated functionality
- [ ] Wrote an automated test for existing and modified functionality
- [ ] Added measurement instrumentation (Mixpanel, etc.)

PR #569, opened 2018-06-19
Details rc 3.5

jonaias:
OK to merge.

Short review.

Summary of changes:

- …

Regressions to look for:

- …

Completed checkin tasks:


jonaias:
Hey @sasha and @matthewtoast , can you take a look to see if it is fine d95b110b63526d03be4488264e41294da2423c4e and 4d4c5d1a0ab48a6f8fdb447f23ed18d045659fc3 ?

About 4d4c5d1a0ab48a6f8fdb447f23ed18d045659fc3, I know that `AtiveComponent` ideally should be synced across every process, but I wasn't able to do it without `moduleSync`. From what I understood,  for `component:reload` (triggered for example, by external file change), glass already has a specific behavior (which is pretty similar to what `reloadBytecode`) and was used as template. But as we already have the updated byecode due `updateBytecode`, I used `moduleSync` instead `moduleReplace`.

tsortwehttam:
@jonaias: You are saying that `component:reload` is not working, which is why you had to add these lines:

- https://github.com/HaikuTeam/mono/pull/569/files#diff-42ab4d87cc0a0abdd17166d500964d13R117
- https://github.com/HaikuTeam/mono/pull/569/files#diff-faac8754effb1942994db60572619863R308
- https://github.com/HaikuTeam/mono/pull/569/files#diff-31168cf166e6f9f90d1f23ecbe13c945R251

Is that correct?

jonaias:
To avoid any race conditions (Sasha helped me there), now is called `replaceBytecode` on `ActiveComponent` instead issuing a module reload (https://github.com/HaikuTeam/mono/pull/568/files#diff-83a177c7138bb8bdd19668b5eb4026e0R79). 

So, code editor save is no longer issuing a component:reload while updating bytecode by editing it on code editor. These lines are there to simulate a component:reload (from the perspective of displaying code editor changes in state editor, timeline and glass) when calling `replaceBytecode` on `ActiveComponent`

stristr:
I think it would be much cleaner (and is probably necessary to cover all edge cases) to put the call to `moduleSync` in `replaceBytecode` itself.

Pretty sure this would fix all of these cases without a need for special-casing.

```
        try {
          this.handleUpdatedBytecode(ModuleWrapper.testLoadBytecode(currentEditorContents, absPath))
        } catch (requireError) {
          // If we cannot validate it, return an error.
          return cb(requireError)
        }
        release()
        fire()
        return this.moduleSync(cb)
```

jonaias:
Now they are ideally synced!!  Thanks for the tip (It was right in front of me)

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -4417,13 +4417,13 @@ class ActiveComponent extends BaseModel {
       return this.project.updateHook('replaceBytecode', this.getRelpath(), currentEditorContents, metadata, (fire) => {
         try {
           this.handleUpdatedBytecode(ModuleWrapper.testLoadBytecode(currentEditorContents, absPath))
-          cb(null)
         } catch (requireError) {
           // If we cannot validate it, return an error.
           cb(requireError)
```
stristr: @jonaias, think you need to `release()` here and return early. Else `cb` gets called twice.

stristr:
This LGTM. Merging to `rc-3.5.0`.

PR #574, opened 2018-06-21
fix: ensure that assets/bytecode are renamed on duplication ⇓

roperzh:
OK to merge.

Short review.

Summary of changes:

This fixes two issues that occur during project duplication:

1- Historically, we only had `.sketch` files to live in the designs folder
and some of this logic was assuming that we only wanted to rename them
during duplication. This changes the logic so all design files are
renamed accordingly.
2- Bytecode entries for "folder" and "project" where not being updated.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-plumbing/src/ProjectFolder.js
+++ b/packages/haiku-plumbing/src/ProjectFolder.js
@@ -418,14 +418,13 @@ export function duplicateProject (destinationProject, sourceProject, cb) {
     // Create a haiku config file.
     fse.mkdirpSync(destinationProject.projectPath);
 
-    const sourcePrimaryAssetPath = Project.getPrimaryAssetPath(sourceProject.projectName);
-    const sourceAssetBasename = path.basename(sourcePrimaryAssetPath);
-    const sourceAssetPathPattern = new RegExp(escapeRegExp(sourceAssetBasename), 'g');
+    // Use the project name as the folder basename
+    const sourceBaseName = sourceProject.projectName;
+    const sourcePathPattern = new RegExp(escapeRegExp(sourceBaseName), 'g');
+    const destinationBasename = destinationProject.projectName;
+    logger.info(`using ${sourceBaseName}, ${destinationBasename}, ${sourcePathPattern}`);
 
-    const destinationPrimaryAssetPath = Project.getPrimaryAssetPath(destinationProject.projectName);
-    const destinationAssetBasename = path.basename(destinationPrimaryAssetPath);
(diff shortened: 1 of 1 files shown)


roperzh:
I targeted `rc-3.5.0-fixes` because seems like something we want to include asap, but please feel free to change the base as you wish, I will take care of merge issues/rebase as they arise.

stristr:
`master` is a good target when we want in next release. I'll adjust accordingly.

stristr:
(Oh, looks like you used it as base branch. Changing back.)

stristr (commented):
LGTM. Merging.

packages/haiku-plumbing/src/ProjectFolder.js
```diff
@@ -418,29 +418,30 @@ export function duplicateProject (destinationProject, sourceProject, cb) {
...
 
-    const sourcePrimaryAssetPath = Project.getPrimaryAssetPath(sourceProject.projectName);
-    const sourceAssetBasename = path.basename(sourcePrimaryAssetPath);
-    const sourceAssetPathPattern = new RegExp(escapeRegExp(sourceAssetBasename), 'g');
-
-    const destinationPrimaryAssetPath = Project.getPrimaryAssetPath(destinationProject.projectName);
-    const destinationAssetBasename = path.basename(destinationPrimaryAssetPath);
-    logger.info(`using ${sourceAssetBasename}, ${destinationAssetBasename}, ${sourceAssetPathPattern}`);
+    // Use the project name as the folder basename
+    const sourceBaseName = sourceProject.projectName;
```
stristr, line 422: Actually, undo LGTM, sorry!

This seems to leave in something that I _think_ was already broken for projects with really long name—basically, we should be using `projectNameSafeShort`, not `sourceProject.projectName`, to ensure we're actually targeting the default files (see `Project.getProjectNameVariations` for details).

I would check if this is functional for a 32-character project name. I'm guessing it is not.

roperzh: Good catch!

> I would check if this is functional for a 32-character project name. I'm guessing it is not.

Yes, seems broken to me. To clarify in case someone else is reading: Sketch and Illustrator files are named with `projectNameSafeShort`, so this logic is not able to find and rename them resulting in the files keeping the name of the original project.

@stristr would `projectNameSafeShort` really help here? seems like we are going to get the same result, because the method `slice`s the first 20 characters, leaving projects with long names in the same situation. For example `projectWithReallyReallyLongName` and `projectWithReallyReallyCopy` both have the same `projectNameSafeShort` (`projectWithReallyRea`).

Aside of this we should keep the full project name in the bytecode entries for "folder" and "project", right?

PR #575, opened 2018-06-21
Direct selection finesse

ghost:
OK to merge.

Fixes and improvements getting direct selection ready for production + path morphing.

Summary of changes:

- Fixes for closed paths having too many control points.
- Fixed scale for direct selection vertices when zoomed.
- Fixed selection issues for strokes and fills not directly specified on the shape.
- Fixed inaccuracies that caused a vertex to drift from the cursor while dragging.
- Fixed direct selection incorrect when designs are reloaded.
- Enabled toggling of syncing between Sketch at the per-instance level.
    - *Note:* I tried to get a `<TooltipBasic>` working for the sync toggle button in the timeline UI, but couldn't get the z-index and offscreen rendering working, so I opted for the `title` attribute.

Regressions to look for:

- Syncing issues with source Sketch files.
- **Still need to enable the direct selection experiment.**

Completed checkin tasks:


ghost:
Guys I don't even know how to Jenkins

tsortwehttam:
RECHECK!

packages/haiku-serialization/src/bll/ActionStack.js
```diff
@@ -59,7 +59,7 @@ const INVERTER_ACCUMULATORS = {
   }
 }
 
-const shouldAccumulate = (method, params) => ACCUMULATORS[method] && !params[3].cursor
+const shouldAccumulate = (method, params) => ACCUMULATORS[method] && !params[4].cursor
```
stristr: Oof. Okay…I see this was an ugly hack to begin with. Is this the same as `params[params.length - 1]`? If so, let's use that so we really shouldn't rely on this very specific, frail assumption about the number of params.

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -3504,18 +3525,27 @@ class ActiveComponent extends BaseModel {
...
-    return this.project.updateHook('updateKeyframes', this.getRelpath(), Bytecode.serializeValue(keyframeUpdates), metadata, (fire) => {
+    return this.project.updateHook('updateKeyframes', this.getRelpath(), Bytecode.serializeValue(keyframeUpdates), options, metadata, (fire) => {
       for (const timelineName in keyframeUpdates) {
         for (const componentId in keyframeUpdates[timelineName]) {
           this.clearCachedClusters(timelineName, componentId)
         }
       }
 
+      if (options && options.setElementLockStatus) {
+        for (let elID in options.setElementLockStatus) {
```
stristr: Nit: this can/should be `for (const elID…`

packages/haiku-timeline/src/components/ComponentHeadingRow.js
```diff
@@ -54,6 +59,19 @@ export default class ComponentHeadingRow extends React.Component {
...
+    const locked = !this.props.row.element.isLocked()
+    this.props.component.updateKeyframes({}, {setElementLockStatus: {[this.props.row.element.getComponentId()]: locked}}, {from: 'timeline'}, () => {
+      if (!locked) {
+        const designs = {
+          [this.props.row.element.getStaticTemplateNode().attributes[HAIKU_SOURCE_ATTRIBUTE]]: true
+        }
+        this.props.component.project.mergeDesigns(designs, {from: 'timeline'}, () => {})
+      }
+      this.forceUpdate()
+    })
```
stristr: Could we make this `mergeDesigns` behavior a consequence of `options.setElementLockStatus` inside ActiveComponent instead of keeping it in the React app? (Typically the way this is handled in AC, to avoid lock conflicts, is to break `mergeDesigns` out into `mergeDesigns`, `mergeDesignsActual`, and `mergeDesignsImpl`—see `zShiftIndices`, `zShiftIndicesActual`, and `zShiftIndicesImpl` for a simple reference—which allows you to call `mergeDesignsImpl` without its duplicative `performComponentWork` and `Lock.request` behaviors.)

You're already calling `this.findElementByComponentId` to recover the `Element` model instance inside `ActiveComponent`, so there should be no added overhead from not already having the model resolved via `this.props.row.element`….

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -3504,18 +3525,27 @@ class ActiveComponent extends BaseModel {
...
 
-    return this.project.updateHook('updateKeyframes', this.getRelpath(), Bytecode.serializeValue(keyframeUpdates), metadata, (fire) => {
+    return this.project.updateHook('updateKeyframes', this.getRelpath(), Bytecode.serializeValue(keyframeUpdates), options, metadata, (fire) => {
       for (const timelineName in keyframeUpdates) {
         for (const componentId in keyframeUpdates[timelineName]) {
           this.clearCachedClusters(timelineName, componentId)
         }
       }
 
+      if (options && options.setElementLockStatus) {
```
stristr: Minor nit: it's less cognitive overhead/easier for future static typing to make `options` a required parameter and pass it as `{}` instead of `null` when no options are required. Then we can just use: `if (options.setElementLockStatus) { … }` here and below.

stristr (commented):
Looks very good. I left some comments—all simple changes except for the request to move the post-`updateKeyframes` `mergeDesigns` work from `<ComponentHeadingRow />` into `ActiveComponent`. But that's probably the most important change requested.

packages/haiku-serialization/src/bll/ActionStack.js
tsortwehttam: Not sure but maybe `params[params.length - 2]` (is the last param the callback?)

ghost: Duh good catch.

ghost:
@stristr thanks for the review! I've implemented everything but I'll need your help taking this the last mile with `mergeDesigns`.

1. When a design is merged, I'm getting a duplicate of all the elements within the design rather than completely replacing the contents (didn't happen before moving `mergeDesigns`).
2. I know `this.project.emit('remote-update', 'mergeDesigns')` doesn't belong here but I'll need some guidance on how to make sure that Glass is still notified of the event. https://github.com/HaikuTeam/mono/pull/575/commits/53597c1a62236457a195605d1dc35c3472dab5c1#diff-7433d312b35db871f4e4161899095cfeR1809

stristr:
Hey @winduptoy, so—the duplication issue is fixed by 0daa60d9da1504333afa0dde2f7f5a6441bdb18d (I also fixed, or made more correct, a broken update hook on `updateKeyframesAndTypes` in that commit).

The behavior I think you wanted is added by e663fd3a0. I'm not totally sure _why_ (and/or if) it was working before moving this stuff around, but probably not for great reasons. It looks like `this.removeChildContentFromBytecode` was not actually optional, though—this is the step that ensures the old template is removed before being maybe-merged.

stristr:
Unrelated, but see also 00e15a6—fixes a sneaky breaking change to undoing changes to `updateKeyframes` which happened due to changing the finicky # of arguments received in that update hook.

This led me to realize `updateKeyframesAndTypes` is not undoable, so you might want to fix that now. The dead-simplest way to handle this is by adding it to `SNAPSHOTTED_UNDOABLES` in `ActionStack`. Although not as performant as a surgical undoing, this will basically take a snapshot of bytecode to roll back to when we change a type.

This LGTM to merge, with the flag still off.

ghost:
@stristr Thanks for your help.

I'm now getting some errors and I can't determine if they're related to these changes or not.

1. Sporadic results when converting a primitive to a path. They seem to collapse in on themselves or make their way to 0,0. I assume this is from some invalid attributes, not a math thing. Any clues on where to start looking to solve this one?

sketch-haiku-sync

2. When converting a primitive to a path, I have gotten this in the timeline. It totally locks up Electron and the stack trace is completely unhelpful. The only info I have to go off of is `intercom-frame` in the bottom left. I have no clues as to what would have called `createOrUpdateUser`. Can't reproduce this one.

screen shot 2018-06-22 at 9 49 19 am

tsortwehttam:
@winduptoy Until @stristr gets online, it might be beneficial to try to isolate the change that caused these bugs. If I were you, I would locally reset to `9197960` (before the merge designs refactor) and see if you can reproduce this. At least then it will be easy to see when the break happened.

stristr:
Guessing any weirdness would be a result of merging 240 commits from `master` into this branch (
745b5d8)…taking a look but doubt it's due to any of the changes here, which are relatively isolated.

ghost:
RECHECK!

PR #578, opened 2018-06-26
RC 3.5.2

stristr:
RC.

Currently has `CodeEditor` experiment in disabled state while we diagnose/fix UI lockups during element dragging.


stristr:
BUILDME!

PR #579, opened 2018-06-26
Making movies

stristr:
OK to merge.

Short review.

Summary of changes:

- Basic functionality for exporting variable framerate GIFs and H.264 videos on-demand, all hidden behind an experiment flag.

Regressions to look for:

- None expected—although there is significant refactoring to the entry point for `haiku-formats` exporting. Suffices to confirm that on publish, `lottie.json` is written out (I have already done this).

Notes to reviewers:
 - The `ffmpeg` solution involves bringing in precompiled binaries via `ffmpeg-static`. (Decided against linking `ffmpeg` directly with a C++ add-on after consultation with @matthewtoast.) That package comes laden with 252MB of precompiled binaries; the expected behavior of the bundling process is to bloat the package size by only the one we're actually using (e.g. `node_modules/ffmpeg-static/bin/*darwin*/*x64*/ffmpeg`). We should make sure that's actually happening (via ad-hoc build) before merging this.

Completed checkin tasks:


stristr:
RECHECK!

stristr:
RECHECK!

stristr:
RECHECK!

stristr:
(plz)

stristr:
BUILDME!

packages/haiku-creator/oven.html
```diff
@@ -0,0 +1,62 @@
+<!doctype html>
```
tsortwehttam, line 1: Just want to say I appreciate your naming all throughout this diff.

packages/haiku-creator/src/bakery/electron.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Point of entry for request a PNG sequence from a bytecode file from the PNG bakery.
+ * @param {string} abspath
+ * @param {(outputDirectory: string) => void} cb
+ */
+export default (abspath: string, framerate: number, cb: () => void) => {
+  const bytecode = requireFromFile(abspath);
+  const haikuId = bytecode.template.attributes['haiku-id'];
+  const width = bytecode.timelines.Default['haiku:' + haikuId]['sizeAbsolute.x'][0].value;
+  const height = bytecode.timelines.Default['haiku:' + haikuId]['sizeAbsolute.y'][0].value;
```
tsortwehttam: Due to "auto"-sizing, you may want to test that width/height are actually numeric before continuing. It is possible that a user might want to set their main component to "auto"-size. To compute a size, you should be able to call `HaikuComponent#computeSize` (`HaikuElement#computeSize`) and friends.

packages/haiku-plumbing/src/Watcher.js
```diff
@@ -42,7 +42,8 @@ export default class Watcher extends EventEmitter {
     this.watcher = chokidar.watch(entry, {
       // - Avoid any git blobs (there are tons of these)
       // - Avoid any sketch tempfiles (these are ephemeral and not needed)
-      ignored: /(node_modules|bower_components|jspm_modules|\.git|~\.sketch)/,
+      // - Avoid files which we explicitly .gitignore
+      ignored: /(node_modules|bower_components|jspm_modules|\.git|~\.sketch|png\/*)/,
```
tsortwehttam: Maybe change the ignore pattern from `png` to `code/*/*.png` or whatever the destination folder of the PNGs is? We shouldn't exclude _all_ pngs. (In the near future, users will be able to import image assets and instantiate them; for that to work with our asset listing logic, we'll need the watcher to be aware of them.)

tsortwehttam (changes requested):

tsortwehttam:
Overall code looks good. Only one or two specific issues I think needs to be addressed before merging.

tsortwehttam:
@stristr -- I wonder if any of this logic could be reused for sake of component thumbnails? https://app.asana.com/0/713061630135205/686777755217976 I haven't thought much about how I'm going to implement, but grabbing a tiny PNG of the component seems at least one plausible way.

packages/haiku-creator/oven.html
stristr: Thank you! Also the name of this PR is a Talking Heads album. Also my prior statement is also a Talking Heads album reference.

packages/haiku-plumbing/src/Watcher.js
stristr: Ah, so glad you mentioned. This _was_ already done to a point (`png/` was excluded, not `png`), but I later changed the directory pattern to be `png-${framerate}` and hadn't refactored this.

Anyway, changing as requested to both be correct and trap the ignore pattern inside of a component folder.

packages/haiku-creator/src/bakery/electron.ts
stristr: Oof. Is this going to be a big problem for Lottie? Right now we do Lottie 100% server side, without a hot rendered component instance to give up the goods. Does this method always work without a window? Or do we need Glass/someone to transmit size on export requests now?

stristr:
@matthewtoast this should be good to go.

638699e is a major refactor (moving `computeLayout` out of `Layout3D`) so please merge when ready.

tsortwehttam (approved):

stristr:
Responding to thought about the PNG reuse for component thumbnails—we _could_, but wouldn't a full (overlaid) render of the component on hover be both sufficient and more awesome?

PR #588, opened 2018-07-02
chore: add title attributes to elements in the UI

roperzh:
OK to merge.

Very short review.

Summary of changes:

I did this during yarn installs, mostly motivated after presenting Haiku to a friend this weekend, he tried multiple times to hover over elements for the usual help tooltip, so I wondered that adding a couple of title tags here and there is good enough for now.

Please feel free to edit for grammar/wording or disregard this completely!

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/components/ProjectBrowser.js
+++ b/packages/haiku-creator/src/react/components/ProjectBrowser.js
@@ -353,7 +353,7 @@ class ProjectBrowser extends React.Component {
               option: 'show-changelog',
             });
           }}>
-          <span style={DASH_STYLES.popover.icon} title="Show Changelog">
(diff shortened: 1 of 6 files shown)


stristr:
Awesome idea, and I agree, let's do this. We really could do more with accessibility in general, and this is a good start—I would even consider using `aria-label` tags at some future point. I made some tiny changes to capitalization. I tried to consistently use the rule "only capitalize something if it's not a generic term", but ultimately it comes down to one's opinions…it was a little tough to make the call on some, but I went with:
 - changelog
 - project
 - file
 - element
 - Library
 - State Inspector
 - Actions

stristr:
LGTM, merging.

roperzh:
yeah! 🙌  thanks for taking care of fixing those.

PR #590, opened 2018-07-02
Preview user console tracer

jonaias:
OK to merge.

Short review.

One thing I'd like to have a second opnion is how logs are relayed through plumbing (`message.type==='log'`). I opted to use websock  to additionally have log messages on plumbing without hacky electron flag (`--enable-logging`). This relaying is also the foundation to enable application logging using categories (a.k.a. tags), more info on https://haiku.quip.com/aKQLAXVYIaiN

Summary of changes:

- …

Regressions to look for:

- …

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -768,6 +768,7 @@ export default class HaikuComponent extends HaikuElement {
     }
 
     try {
+      this.context.info('ACTIONS_FIRED', `Action ${eventName} on element ${eventsSelector}`, {});
       return handler.apply(this, eventArgs);
     } catch (exception) {
       consoleErrorOnce(exception);
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -355,9 +355,9 @@ export default class HaikuComponent extends HaikuElement {
 
   set (key, value) {
     this.traceInfo('STATE_CHANGES', `State ${key} changed from ${this.state[key]} to ${value}`,
-                      {state: key,
-                      from: this.state[key],
-                      to: value});
+      {state: key,
+        from: this.state[key],
+        to: value});
 
     this.state[key] = value;
     return this;
@@ -1347,10 +1347,18 @@ export default class HaikuComponent extends HaikuElement {
     return this;
   }
   $console = {
-    log: (params) => { this.traceInfo('CONSOLE', params, {}); },
-    info: (params) => { this.traceInfo('CONSOLE', params, {}); },
-    warn: (params) => { this.traceInfo('CONSOLE', params, {}); },
-    error: (params) => { this.traceInfo('CONSOLE', params, {}); },
+    log: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+    info: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+    warn: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+    error: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
   };
 
   // Set logger to add context info when logging while avoiding runtime overhead
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -156,7 +156,7 @@ export default class HaikuComponent extends HaikuElement {
     this.state = {}; // Public accessor object, e.g. this.state.foo = 1
 
     // Instantiate StateTransitions. Responsible to store and execute any state transition.
-    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock());
+    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock(),  this.context);
 
     // `assignConfig` calls bindStates because our incoming config, which
     // could occur at any point during runtime, e.g. in React, may need to update internal states, etc.
@@ -353,6 +353,11 @@ export default class HaikuComponent extends HaikuElement {
   }
 
   set (key, value) {
+    this.context.info('STATE_CHANGES', `State ${key} changed from ${this.state[key]} to ${value}`,
+                      {state: key,
+                      from: this.state[key],
+                      to: value});
+
     this.state[key] = value;
     return this;
   }
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -76,13 +76,14 @@ export default class HaikuComponent extends HaikuElement {
   CORE_VERSION;
   doAlwaysFlush;
   doesNeedFullFlush;
-  guests;
-  host;
+  guests: {[haikuId: string]: HaikuComponent};
+  host: HaikuComponent;
   playback;
   PLAYER_VERSION;
   registeredEventHandlers;
   state;
   stateTransitionManager: StateTransitionManager;
+  logger;
 
   constructor (
     bytecode: HaikuBytecode,
@@ -156,7 +157,7 @@ export default class HaikuComponent extends HaikuElement {
     this.state = {}; // Public accessor object, e.g. this.state.foo = 1
 
     // Instantiate StateTransitions. Responsible to store and execute any state transition.
-    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock(),  this.context);
+    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock(),  this);
 
     // `assignConfig` calls bindStates because our incoming config, which
     // could occur at any point during runtime, e.g. in React, may need to update internal states, etc.
@@ -353,7 +354,7 @@ export default class HaikuComponent extends HaikuElement {
   }
 
   set (key, value) {
-    this.context.info('STATE_CHANGES', `State ${key} changed from ${this.state[key]} to ${value}`,
+    this.traceInfo('STATE_CHANGES', `State ${key} changed from ${this.state[key]} to ${value}`,
                       {state: key,
                       from: this.state[key],
                       to: value});
@@ -773,7 +774,7 @@ export default class HaikuComponent extends HaikuElement {
     }
 
     try {
-      this.context.info('ACTIONS_FIRED', `Action ${eventName} on element ${eventsSelector}`, {});
+      this.traceInfo('ACTIONS_FIRED', `Action ${eventName} fired on element ${eventsSelector}`, {});
       return handler.apply(this, eventArgs);
     } catch (exception) {
       consoleErrorOnce(exception);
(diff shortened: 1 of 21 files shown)


packages/@haiku/core/src/HaikuTimeline.ts
```diff
@@ -273,7 +288,7 @@ export default class HaikuTimeline extends HaikuBase {
    * the falsy state timelines have when first constructed.
    */
   isExplicitlyPaused () {
-    return this._isPlaying === false;
+    return !this._isPlaying;
```
stristr: @jonaias, was this changed on purpose?

I don't believe it is correct to change it from what it was. Note the name of this method—it's meant to check if the timeline was _explicitly_ paused (e.g. by calling `.gotoAndStop()`, `.pause()`, etc.).

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -156,7 +157,7 @@ export default class HaikuComponent extends HaikuElement {
     this.state = {}; // Public accessor object, e.g. this.state.foo = 1
 
     // Instantiate StateTransitions. Responsible to store and execute any state transition.
-    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock());
+    this.stateTransitionManager = new StateTransitionManager(this.state, this.getClock(),  this);
```
stristr: Now that StateTransitionManager receives the entire component, does it need to receive `this.state` and `this.getClock()`? It looks like these are public (so can be derived from the `HaikuComponent` argument).

stristr: (Note: I think these can still be set to `readonly` in the class spec, as long as they are assigned in the constructor.)

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -1339,6 +1346,42 @@ export default class HaikuComponent extends HaikuElement {
...
+    info: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+    warn: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+    error: (params) => {
+      this.traceInfo('CONSOLE', params, {});
+    },
+  };
```
stristr: This is probably not needed for MVP, so I would prefer taking it out to "fixing" it, but…if we keep it, it really should [implement the methods](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v9/index.d.ts#L34) of the `Console` interface, which use different signatures (for example, `log(message?: any, ...optionalParams: any[]): void;`).

Again, since this is unused, I would just prefer to take it out.

packages/@haiku/core/src/HaikuTimeline.ts
```diff
@@ -32,18 +32,19 @@ export type PlaybackStatus = PlaybackSetting | number | string;
...
-  _localElapsedTime;
-  _localExplicitlySetTime;
-  _maxExplicitlyDefinedTime;
-  _isPlaying;
+  _globalClockTime: number;
+  _localElapsedTime: number;
+  _localExplicitlySetTime: number;
+  _maxExplicitlyDefinedTime: number;
+  _isPlaying: boolean;
+  _loopCounter: number;
```
stristr: Please refactor _new_ variables (not old ones) to avoid this deprecated pattern. (There is always a good chance the existing "_private" variables are used in unexpected places in our hosted services, so better to leave them alone. But prefer `private loopCounter` to `[public] _loopCounter`.

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -1339,6 +1346,42 @@ export default class HaikuComponent extends HaikuElement {
...
+  }
+
+  traceInfo (tag: string, message: string, attachedObject: any) {
+    const haikuRootComponent = this.getRootComponent();
+
+    if (haikuRootComponent.logger) {
+      attachedObject.sceneName = this.title;
+      haikuRootComponent.logger.traceInfo(tag, message, attachedObject);
+    }
+  }
```
stristr: Regarding the design of `setLogger` and `traceInfo`: it feels like we have hoisted `Haiku.app` concerns into `@haiku/core` when we didn't really need to. Personally, I would prefer an event-based model, which would provide both consistency and a lighter footprint. For example a lightweight approach could be:

```
  emitFromRootComponent(key: string, ...args: any[]) {
    this.getRootComponent().emit(key, {sceneName: this.title}, ...args);
  }
```

Then, for example, the `LOOP_COUNTER` event could instead look like:

```
this.component.emitFromRootComponent('loop', {count: `this._loopCounter`});
```

And over in `Haiku.app`, we could approach this in a number of ways. One nice way unlocked by how Matthew rewired handlers is declaratively:

```
# Preview.js

this.component = factory(
  this.mount,
  {
    ...,
    onLoop: ({sceneName}, {count}) => {
      const timeline = this.component.getDefaultTimeline();
      logger.traceInfo('LOOP_COUNTER', `Loop count ${count}`, {
        sceneName, // (Unless we want to call this something other than `sceneName`?)
        elapsedTime: timeline.getElapsedTime(),
        ...,
      });
    }
  }
);
```

This approach enforces a couple of boundaries that I thought should really be enforced:
 - Things like "formatted log message" (e.g. `Loop count 2`) should not be in Core—they have nothing to do with rendering/production runtime.
 - This approach forces you to use the _public_ data from the component/timeline, rather than leaking a bunch of "private" data like `_localElapsedTime`. A perfectly fine approach to resolving a bunch of private data through a public interface could be to define a public method on `HaikuTimeline`, something like `getDebugInfo()`—if it's needed or useful.

Curious if this approach was considered and if @matthewtoast has any comments on it.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -880,6 +880,11 @@ export default class Creator extends React.Component {
   handleInteractionModeChange (interactionMode) {
     this.setState({interactionMode});
     this.mixpanelReportPreviewMode(interactionMode);
+    // Clear all logs before starting preview on production. For development it might be useful to not clean them
+    if (this.state.interactionMode === InteractionMode.LIVE && interactionMode === InteractionMode.EDIT &&
+        process.env.NODE_ENV === 'production') {
```
stristr: Note: we have some constants in `haiku-common/lib/environments` which are slightly better than this.

packages/haiku-serialization/src/utils/LoggerInstance.js
```diff
@@ -1,5 +1,5 @@
 const fs = require('haiku-fs-extra')
-const Logger = require('./Logger')
+const Logger = require('./Logger').Logger
```
stristr: Preferred:

`const {Logger} = require('./Logger')`

Will also be easier to refactor when we convert to `import`/TypeScript.

packages/haiku-creator/src/react/components/LogViewer/LogViewer.js
```diff
@@ -50,33 +79,77 @@ class LogViewer extends React.Component {
...
+              this.refs.loglist.scrollTo(this.state.logMessages.length - 1);
+            }
+          });
+        }
+      });
     }
   }
 
   componentWillUnmount () {
     logger.removeListener('log', this.logHandler);
```
stristr: Guessing this is no longer needed?

packages/haiku-serialization/src/utils/Logger.js
```diff
@@ -37,13 +48,33 @@ const DEFAULTS = {
...
+  // We send to pumbling so we can log to pumbing console in a nice way
+  sendToPlumbing (message) {
+    if (this.websocket) {
+      this.websocket.send({
+        type: 'log',
+        from: message.view,
+        message
+      })
+    }
+  }
```
stristr, line 73: Regarding this maneuver: it's _okay_, but Envoy is our preferred transport mechanism for new code/features. We are planning to slowly deprecate all this direct websocket IPC in favor of explicit, typed, event-based handlers. The most complete/correct example we have of this is in `ExporterHandler`—see `packages/haiku-plumbing/src/envoy/attachListeners.ts` for where you could hook in and [here](https://github.com/HaikuTeam/mono/blob/fdfc16d98780298bfc3a6df95bb94d6747633837/packages/haiku-plumbing/src/Plumbing.js#L250) for how to bind a new Envoy handler on a new channel.

LMK if you have any questions about this—I think you might find it to be a more elegant solution!

stristr (commented):
Left some comments. Please LMK if any questions!

packages/@haiku/core/src/HaikuTimeline.ts
jonaias: Oh, my bad, i thought isPlaying (also by its naming) was a boolean member, that's why i tried to simplify. I will revert this.

packages/@haiku/core/src/HaikuComponent.ts
jonaias: I was fixing it while fixing state transition test. Thanks for remembering.

packages/@haiku/core/src/HaikuComponent.ts
jonaias: I'll take it out.. to do a clear console output, i guess it would necessary some changes. Eg passing a console instance to all event handlers in addition to other parameters. It was a proof of concept. (I agree `$console.log` isn't pretty)

packages/@haiku/core/src/HaikuTimeline.ts
jonaias: I added these types to increase typification of our source code, but got it, i'll leave them alone.

packages/@haiku/core/src/HaikuComponent.ts
jonaias: It a good option. The attachedObject parameter from Logger `traceInfo` definition can be used for this purpose. We can format messages using key (parameter tag)+attachedObject directly on Logger.

My idea as discussed in quip console design document  is to populate our code base with  traceInfo('TAG','message',attachedObject), so in addition to  preview user console, we also gain a easy toggable trace log. For core stuff, it easily could be only `trace('TAG',attachedObject)`, but we would lose ability to debug Haiku component without Haiku application.. for me either way is fine, the most important thing is to always keep each log message at least a tag+attachedObject

packages/haiku-creator/src/react/Creator.js
jonaias: :+1:

packages/haiku-serialization/src/utils/LoggerInstance.js
jonaias: :+1:

packages/haiku-creator/src/react/components/LogViewer/LogViewer.js
jonaias: :+1:

packages/haiku-serialization/src/utils/Logger.js
jonaias: I opted to use websocket because i tend to choose the simpler (In my often-wrong-opnion) solution. I agree  it might be more elegant as envoy has named channel for each type of message. Can you detail a little more how you would think its the best way to implement? 

For example, I need to create a LoggerHandler.ts under `packages/haiku-sdk-creator/src/logger`, then on Logger.js LogForwarderTransport  I would send using stuff from  `import {LOGGER_CHANNEL, LoggerHandler} from 'haiku-sdk-creator/lib/logger';`?

packages/@haiku/core/src/HaikuTimeline.ts
stristr: Sorry—the types (even if incorrect) don't carry any serious risk. I'm just suggesting that we not make any new variables starting with `_`, and prefer `private` keyword instead.

packages/haiku-serialization/src/utils/Logger.js
stristr: Yes, that's exactly what I meant—you'd make a new channel `LOGGER_CHANNEL = 'logger'` and bind a new handler to it.

I don't think it's essential to make this change in order to get this PR merged—we can come back to it if it's a lot of extra work.

stristr:
LGTM! Merging.

PR #591, opened 2018-07-04
Sizing, Snapping, Distribute, Align, and Misc

tsortwehttam:
Not ready to merge.

- Snapping, distribute, align merged in via #592 -- see Zack's note about known issues.
- Groundwork toward support for primitives, which is closely tied to layouts/sizing
- Core improvements that play better with layouts/sizing/multi-component/primitives
- Work toward on-stage sizing that respects size-vs-distortive-scale (feature-flagged)
- Handful of bug-fixes related to property application and multi-component
- Weekend hack-a-thon groundwork toward supporting bitmap images (feature-flagged)

If you're reviewing this code, you want commit cd15d91.

PR #592, opened 2018-07-04
Snapping, align & distribute

zackbrown:
OK to merge.

Medium/long review.

Summary of changes:

- Adds snapping for translate & scale
- Adds align & distribute

Known issues:

- Align/distribute operations are very fast, but when triggered from the Creator process there's high latency.  (When triggered from Glass, latency is low — you can see this by pressing the magic keys "k" and "l", which are each hard-wired to a specific align/distribute operation.)
- When an element is ROTATED and you are SCALING from an EDGE (non-corner,) there's a case where two corners NEARLY snap to respective horizontal and vertical lines, but in fact only one is actually snapped.  In this case, both horizontal & vertical lines show, when they should not.  I judged that the ratio of impact to fix-effort was small enough to let this slide for now.
 - When DISTRIBUTING tops & bottoms TO THE STAGE, sometimes elements spill slightly outside of the stage.  I'm unsure if this is actually the correct math or not (it does work as expected when distributing centers,) so I decided this was good enough to ship.
 - The align panel UI will not update its x/y position when it's active and the component button shows up or disappears.  Since I'm not sure this is even the ideal place/design for the align panel (would be nice for @taylorpoe to take a pass at some point,) I left this for now.

### DIFF
--- a/packages/haiku-creator/src/react/components/StageTitleBar.js
+++ b/packages/haiku-creator/src/react/components/StageTitleBar.js
@@ -10,7 +10,7 @@ import {BTN_STYLES} from '../styles/btnShared';
 import Toggle from './Toggle';
 import {ShareModal} from 'haiku-ui-common/lib/react/ShareModal';
 import {
-  ComponentIconSVG, ConnectionIconSVG, DangerIconSVG, EventsBoltIcon, PublishSnapshotSVG, WarningIconSVG,
+  ComponentIconSVG, ConnectionIconSVG, DangerIconSVG, EventsBoltIcon, PublishSnapshotSVG, WarningIconSVG, AlignDistributeIcons,
 } from 'haiku-ui-common/lib/react/OtherIcons';
 import {ExporterFormat} from 'haiku-sdk-creator/lib/exporter';
 import * as Element from 'haiku-serialization/src/bll/Element';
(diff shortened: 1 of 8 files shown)


tsortwehttam:
RECHECK!

packages/haiku-creator/src/react/components/StageTitleBar.js
```diff
@@ -577,6 +614,18 @@ class StageTitleBar extends React.Component {
...
+  isAlignPanelAvailable () {
+    // This would show the align panel only when elements are selected:
+    // const proxy = this.fetchProxyElementForSelection();
+    // return proxy && !proxy.hasNothingInSelection();
+    // But it feels better (in zb's opinion at authoring time) to keep it shown at all times glass is shown
+    return this.props.showGlass;
+  }
+
+  shouldShowAlignPanel () {
+    return this.state.alignPanelShown && !this.state.showSharePopover && !this.props.isPreviewMode && this.props.showGlass;
```
roperzh, line 626: heads up/question: since we are hiding this for the sharePopover, should we hide it also for the actions editor?

packages/haiku-serialization/src/bll/Artboard.js
```diff
@@ -266,6 +270,102 @@ class Artboard extends BaseModel {
...
+    snapLines.push({
+      direction: 'HORIZONTAL',
+      positionWorld: (topWorld + bottomWorld) / 2,
+      elementId: 'STAGE_HORIZONTAL_MID'
+    })
+
+    const rootElement = this.getElement()
+    const topLevelElements = rootElement.children
+
+    _.forEach(topLevelElements, (elem) => {
```
roperzh, line 321: I think it's completely safe to use `[].forEach` here, death to lodash! 🔥

tsortwehttam:
I've merged this into my PR, #591 since there will be conflicts/issues to resolve w/r/t this snapping work vs. sizing modes and layouts. Going to close this PR in favor of that one so we can have the work consolidated.

PR #597, opened 2018-07-10
Path morphing

ghost:
OK to merge.

Core and Lottie export implementation for morphing of single-polygon paths.

Summary of changes:

- Path morphing utilities. 
- ValueParser synchronizes path structures where necessary when provided a `d` attribute.
- Added VSCode launchers for debugging `regenerate-goldens` and `test` scripts in `haiku-formats` package.

Regressions to look for:

- Anything that touches ValueParser?
- Incorrect Lottie path exports?

Completed checkin tasks:


stristr:
@winduptoy, "failing tests" include a parse error in `haiku-formats`—looks like it crashes in the middle of running the `haikuStatic` tests.

Also failures in `ActiveComponent.2.ativo`, `ActiveComponent.3.svg-pattern`, and `ActiveComponent.3.svg-figma-pattern` in `haiku-serialization`—likely flakes but worth taking a look.

packages/haiku-formats/scripts/regenerate-goldens.ts
```diff
@@ -1,9 +1,9 @@
-import {BodymovinExporter} from '@/exporters/bodymovin/bodymovinExporter';
-import {HaikuStaticExporter} from '@/exporters/haikuStatic/haikuStaticExporter';
 import {each} from 'async';
 // @ts-ignore
 import {readdir, writeFile} from 'fs-extra';
 import {basename, join} from 'path';
+import {BodymovinExporter} from '../src/exporters/bodymovin/bodymovinExporter';
+import {HaikuStaticExporter} from '../src/exporters/haikuStatic/haikuStaticExporter';
```
stristr: For consistency, please change `'../src/…` to `'@formats/...`.

packages/haiku-formats/test/goldens/bodymovin/percy.json
```diff
@@ -41129,7 +41138,7 @@
               "ks": {
                 "a": 0,
                 "k": {
-                  "c": false,
+                  "c": true,
```
stristr: @winduptoy—it looks like the goldens regeneration converted all paths to closed paths. Have you checked if this causes any visual regressions when running `yarn demos`? I would expect it does in some cases, but dunno.

stristr:
Re: the crash.

Looks like you need to run (or add output files from) `yarn regenerate-goldens` against the new golden `AnimatedPath.json`?

packages/@haiku/core/src/helpers/PathUtil.ts
```diff
@@ -0,0 +1,334 @@
+/* tslint:disable:max-line-length */
+
+import {CurveDef, CurveSpec} from '@haiku/core/src/vendor/svg-points/types';
```
stristr: Note: all imports inside a TS package should be relative. Usually this only works because you have `@haiku/core` self-linked due to how demos work—i.e. won't work in prod.

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -816,15 +824,35 @@ export class BodymovinExporter extends BaseExporter implements ExporterInterface
...
+            this.decorateShape(segmentTimeline, shapeSegment);
+            if (shapeDescriptor.closed) {
+              // Force closed if specified on the shape segment
+              shapeSegment[ShapeKey.Vertices][PropertyKey.Value][PathKey.Closed] = true;
+            }
+            groupItems.push(shapeSegment);
+          });
+          // Decorate the original shape in case we need to manage a complex fill (e.g. gradient stops).
+          this.decorateShape(timeline, shape);
+          shape[ShapeKey.Vertices][PropertyKey.Value][PathKey.Closed] = true; // Force closed
```
stristr: Guessing this is where the `"c": true` leaked in the new exports. Why this?

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -816,15 +824,35 @@ export class BodymovinExporter extends BaseExporter implements ExporterInterface
...
-        const path = getPath(initialValue(timeline, 'd'));
-        const pathSegments = decomposePath(path);
-        pathSegments.forEach((shapeDescriptor) => {
-          const shapeSegment = {...shape};
-          this.decorateShape(shapeDescriptor.points, shapeDescriptor.closed, shapeSegment);
-          groupItems.push(shapeSegment);
-        });
-        // Decorate the original shape in case we need to manage a complex fill (e.g. gradient stops).
-        this.decorateShape(path, true, shape);
+        if (Object.keys(timeline.d).length > 1) {
```
stristr, line 1036: Hmmm, this fork is not really ideal—I'll leave it be until we can support tweening multiple closed shapes, but I'm not convinced we should break compound shapes if they're tweened.

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
```diff
@@ -1014,6 +1042,42 @@ export class BodymovinExporter extends BaseExporter implements ExporterInterface
     });
   }
 
+  /**
+   * Normalizes tweened curves to suppot path morphing.
+   */
+  private preprocessTweenedCurves () {
+    this.visitAllTimelineProperties((timeline, property) => {
```
stristr, line 1290: Couple of observations here:
 - `normalizeCurves` adds a curve to every transition, guaranteeing that `totalTweens ` is always going to be equal to `keyframes.length`.
 - In light of this, it feels like we should favor a non-mutating early return in the majority of cases where a path is _not_ tweened (i.e. as soon as we have `keyframes` and can check if `keyframes.length < 2`), if for no other reason than for better perf. `string` -> `CurveSpec[]` conversion is handled already inside `decomposePath`, which is the path the interpreter will follow for non-tweened paths, so I think this would simply entail changing the early return to:

```
if (property !== 'd' || Object.keys(timeline[property]).length < 2) {
  return;
}
```

And removing all the `totalTweens` stuff beneath.

packages/haiku-formats/scripts/regenerate-goldens.ts
ghost: Shucks, VSCode kept changing that to `../src` on me...

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
ghost: See the diff of the `decorateShape` call that was originally here:

```
this.decorateShape(path, true, shape);
```

where the second argument is `closed === true`. I believe that this replicates the same behavior?

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
ghost: I'm afraid I don't follow you. As I read it, this fork:

- If tweened, go straight to `decorateShape`
- Otherwise, break up the untweened compound shape (original behavior before this branch)

stristr: Oh, because a tweened compound shape isn't impossible (even though we haven't designed tweening for it yet, it still might occur). And when it does occur, the Lottie path will definitely be "wrong"…because this code path essentially skips all `M`s and `Z`s.

Still, let's leave it be until tweening multiple closed shapes is a thing.

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
ghost: Nice catch. However, if we return early there and don't do the `string` &rarr; `CurveSpec[]` parsing here, then we end up with an exception because of a `string` instead of a `CurveSpec[]` at the final call to `decorateShape` in the compound-path branch of `handleShape` (approx line 852, where we commented above). Ideas?

ghost:
Currently, I have these tests failing from the `haiku-serialization` package.

```
   ActiveComponent.2.ativo

      ✖ html output ok


    ActiveComponent.3.svg-pattern

      ✖ html output ok


    ActiveComponent.3.svg-figma-pattern

      ✖ html output ok
```

These tests are also currently failing on `master` for me. Maybe someone could confirm this on their end?

tsortwehttam:
Those tests pass for me in `master`. You may need to run `$ yarn compile-all` after switching to `master`.

tsortwehttam:
My guess is you want these two diffs: https://github.com/HaikuTeam/mono/commit/7492fe17ca7fd2678dbf3b1e86cb62e1b4ed6fc3#diff-a80acded10d2320afd55348d0221c102L40

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
stristr: Hmmm, yeah, there may be something else causing the issue (e.g. things getting passed around by reference). See my comment on `percy.json`—to fix we need `"c": false` not to be changing to `"c": true` in `test/goldens/bodymovin/*.json`.

stristr: (Confirmed that this isn't it—and also you can just remove this line, given that the value of this object is literally never read. It's only done so that the shape has a size, which in turn is only done so we can handle gradient stops and the like. Lottie is fun!)

packages/haiku-formats/src/exporters/bodymovin/bodymovinExporter.ts
stristr: Oh yeah.

I guess the path of least resistance (just so we don't parse the string into `CurveSpec[]`) _again_ is:

```
this.decorateShape({d: {0: {value: [].concat(...pathSegments)}}, shape);
```

stristr:
(fixed conflicts with #599)

ghost:
@stristr I implemented all the improvements you brought up above and now we've got 100% equality with the original lottie exports (percy is fixed). AFAIK this is ready to merge.

stristr:
LGTM, let's merge. (I have some more Lottie fixes I'd like to layer on top of this.)

PR #607, opened 2018-07-16
Sdk billing tweaks

roperzh:
OK to merge.

Short review.

Summary of changes:

- fix: change sdk-inkstone addCard to do a POST request
- fix: do not parse JSON by default in sdk-inkstone request, I couldn't figure out why at first glance, but the body of the request is already being parsed before reaching this method, causing an error when trying to parse a JS object.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/sdk-inkstone/src/index.ts
+++ b/packages/@haiku/sdk-inkstone/src/index.ts
@@ -1,7 +1,7 @@
 /* tslint:disable:no-namespace */
 import * as request from 'request';
 import {InkstoneConfig, inkstoneConfig} from './config';
-import {Endpoints, MaybeAuthToken, newDeleteRequest, newGetRequest, newPutRequest} from './services';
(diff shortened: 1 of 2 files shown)


stristr:
Ah, thanks for the fixes.

I have a hunch the JSON issue is not as straightforward as it seems…we swap `request`->`xhr` module at build-time in our web projects to drastically reduce bundle size, and it seems like maybe the `xhr` module handles `Content-Type: application/json` automatically while the `request` module does not…

Anyway, since these endpoints are not expected to be used outside `account` for now, I'll handle this the next time I'm working in `sdk-inkstone` (which will be this week).

LGTM, please merge after health checks pass.

roperzh:
Ah, that sounds tricky. Thanks for the review, and great job with the new methods to define endpoints, they are really nice.

stristr:
RECHECK!

PR #610, opened 2018-07-17
Path Auto-Keyframe

ghost:
OK to merge.

Keyframe behavior matches everything else when modifying directly-selected shapes.

Summary of changes:

- Enabled numeric parsing for appropriate keyframe values.
- When modifying a path or other directly-selected shape, keyframes are inserted at the current timeline time, just like all other properties.
- When converting a shape with keyframes to a path, corresponding keyframes for the path`d` attribute are created to give the shape the same animation. **Easing is omitted here, because multiple keyframe channels are merged.**

Regressions to look for:

- ?

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/Transitions.ts
+++ b/packages/@haiku/core/src/Transitions.ts
@@ -13,19 +13,6 @@ function numberize (n) {
   return parseInt(n, 10);
 }
 
-function sortedKeyframesCached (keyframeGroup) {
-  // Cache the output of this on the object since this is very hot
-  if (keyframeGroup.__sorted) {
-    return keyframeGroup.__sorted;
-  }
-
-  const sorted = sortedKeyframes(keyframeGroup);
-
-  // Cache the sorted keys
-  keyframeGroup.__sorted = sorted;
-  return keyframeGroup.__sorted;
-}
-
 function sortedKeyframes (keyframeGroup) {
   const keys = Object.keys(keyframeGroup);
   const sorted = keys.sort(ascendingSort).map(numberize);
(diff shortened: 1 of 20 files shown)


packages/@haiku/core/src/Transitions.ts
```diff
@@ -36,7 +23,7 @@ function sortedKeyframes (keyframeGroup) {
 // 2500: { value: { ... } }
 // 5000: { value: { ... } }
 function getKeyframesList (keyframeGroup, nowValue) {
-  const sorted = sortedKeyframesCached(keyframeGroup);
+  const sorted = sortedKeyframes(keyframeGroup);
```
stristr, line 26: @winduptoy, what is the rationale for doing away with the concept of the sorted keyframes cache?

I'm not the author of this optimization and I haven't really benchmarked to introspect its significance, but given how hot this code path is, I would expect it to matter at least somewhat.

tsortwehttam: I suggested this. It fixes an implementation detail making it unnecessary to check for the presence of `__sorted` in a key path, something I've wanted to get rid of for a long time. (My rationale is: We are going to do a perf pass next week at which point we can make a more informed decision about how to improve this bottleneck using our more modern caching utilities which are less race-/invalidation-bug-prone (and which weren't around when I added this hack back in the early days).)

stristr: Okay. I would expect this to lower perf (sorting has a cost, and this is essentially triggered on all properties on all mutable timelines on every tick), but agree there's a better way to handle this upstream.

stristr:
LGTM.

PR #614, opened 2018-07-18
Implement haiku protocol handler on windows and linux

jonaias:
WIP.

Implement haiku protocol handler on windows and linux.

Also implement option to pass haiku// protocol URI as parameter. (It works with yarn start <haiku URI> or with installed haiku), for example `haiku haiku://fork/taylor/world`

Completed checkin tasks:

- [X] Did manual testing of interrelated functionality on Linux
- [x] Did manual testing of interrelated functionality on Windows


scripts/distro-prepare.js
```diff
@@ -57,7 +57,7 @@ logExec(ROOT, `yarn install ${YARN_INSTALL_FLAGS} --production`);
 // Important: if we have any modules self-linked, get rid of them with extreme prejudice.
 log.log(`Clear any self-linked module`);
 rimraf.sync('packages/*/node_modules/@haiku');
-rimraf.sync('/*/node_modules/haiku-*');
+rimraf.sync('./*/node_modules/haiku-*');
```
stristr: This should probably just be removed. The next two lines cover all use cases.

index.js
```diff
@@ -12,6 +12,15 @@ if (!global.process.env.NODE_ENV || global.process.env.NODE_ENV === 'production'
...
 }
 
+// On Windows and Linux, custom protocol handler is passed as argument
+let haikuURI = null;
+for (const arg of process.argv) {
+  if (arg.startsWith('haiku://')) {
+    haikuURI = arg;
+    break;
+  }
+}
```
stristr: Nit: I believe a more succinct way to write this would be:

`const haikuURI = process.argv.find((arg) => arg.startsWith('haiku://'));`

stristr:
LGTM (but see comments), and thanks very much for the flexible command line options!

 - Rebase to `origin/master` to fix health checks.
 - Please remove `yarn test-url` and the script it triggers! It provides a less general (Mac-only) solution to the same problem you solved with the flexible command line option.

stristr:
(looks like this breaks build on Mac 😿)

jonaias:
RECHECK!

packages/@haiku/cli/package.json
```diff
@@ -8,8 +8,8 @@
     "lint-report": "yarn lint -t checkstyle -o checkstyle-result.xml",
     "fix": "yarn lint --fix",
     "ts": "cross-env NODE_ENV=test ts-node -r tsconfig-paths/register -P tsconfig.all.json",
-    "test": "yarn ts ./node_modules/.bin/tape 'test/**/*.test.ts' | tap-spec",
-    "test-report": "nyc yarn ts ./node_modules/.bin/tape 'test/**/*.test.ts' > test-result.tap",
+    "test": "yarn ts ../../../node_modules/.bin/tape 'test/**/*.test.ts' | tap-spec",
+    "test-report": "nyc yarn ts ../../../node_modules/.bin/tape 'test/**/*.test.ts' > test-result.tap",
     "compile": "tsc",
```
stristr: So, just to clarify the purpose of the postinstall script:
 - The packages are expected to stand on their own—`@haiku/cli`, `@haiku/core`, and someday many more repos are public on GitHub, and we want people to be able to clone them, check them out, run their tests, etc. This `../../..` won't fly because for those users, the binaries will be in the usual place (`./node_modules/.bin/.../`).
 - Unfortunately, `yarn` workspaces are not smart enough to provide the binaries from `haiku-testing` because `haiku-testing` is itself part of the `yarn` workspace. [Issue on GitHub.](https://github.com/yarnpkg/yarn/issues/4964)
 - The whole reason I did this is to avoid having to sync a bunch of dependencies across all the individual packages. But given the cross-env headache, the next best thing to do is probably that. The specific `devDependencies` we'd want in every package that currently depends on `haiku-testing` are:

```
    "cross-env": "^5.1.6",
    "istanbul-reporter-cobertura-haiku": "^1.0.1",
    "nyc": "^11.8.0",
    "tap-spec": "^4.1.2",
    "tape": "^4.9.0",
    "ts-node": "^6.1.0",
    "tsconfig-paths": "^3.3.2",
    "tslint": "^5.10.0",
    "tslint-config-haiku": "^1.0.15",
    "typescript": "^2.9.1"
```

At that point, these specific dependencies could also move to `devDependencies` for `haiku-testing` as well—and the whole `bin` directory can be cleared out.

stristr:
LGTM. :shipit:

PR #617, opened 2018-07-19
feat: add support for placeholders in Angular adapter.

stristr:
OK to merge.

Short review.

Summary of changes:

- Adds support for placeholders in the Haiku Angular adapter. Like React, placeholders can be regular DOM or other Angular-adapted components. Unlike React, I skipped building support for generic CSS selectors because this really wouldn't make sense to do in core.
 - As noted inline, Angular Is Special™, so we have to specify placeholders like this:

```
<my-component>
  <my-placeholder-component #placeholder></my-placeholder-component>
  <strong #placeholder>Blah</strong>
</my-component>
```

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/adapters/angular-dom/HaikuAngularDOMAdapter.ts
+++ b/packages/@haiku/core/src/adapters/angular-dom/HaikuAngularDOMAdapter.ts
@@ -3,8 +3,10 @@
  */
 
 import {
+  AfterContentInit,
   AfterViewInit,
   Component,
+  ContentChildren,
   ElementRef,
   EventEmitter,
   Input,
@@ -13,9 +15,11 @@ import {
   OnChanges,
   OnDestroy,
   Output,
+  QueryList,
   SimpleChanges,
   ViewChild,
 } from '@angular/core';
+import {BytecodeNode} from '../../api';
 import HaikuComponent from '../../HaikuComponent';
 import {randomString} from '../../helpers/StringUtils';
 
@@ -24,7 +28,7 @@ export default function HaikuAngularDOMAdapter (
   selector, haikuComponentFactory, optionalRawBytecode?): ModuleWithProviders {
   @Component({
     selector,
-    template: `<div #mount class="haiku-angular-wrapper" [id]="randomId"></div>`,
+    template: `<div #mount class="haiku-angular-wrapper" [id]="randomId"><ng-content></ng-content></div>`,
     styles: [`
       div {
         position: relative;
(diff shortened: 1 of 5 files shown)


packages/@haiku/core/src/adapters/angular-dom/HaikuAngularDOMAdapter.ts
```diff
@@ -95,24 +111,67 @@ export default function HaikuAngularDOMAdapter (
...
+
     haikuConfig: {[key in string]: any} = {
       onComponentWillInitialize: (component) => this.componentWillInitialize.emit(component),
       onComponentDidInitialize: (component) => this.componentDidInitialize.emit(component),
       onComponentDidMount: (component) => this.componentDidMount.emit(component),
       onComponentWillUnmount: (component) => this.componentWillUnmount.emit(component),
+      vanities: {
+        // #FIXME: in Angular, surrogates can currently only be selected by numeric index. To support CSS queries (which
+        // Angular provides robust utilities for), we should really move adapter-specific surrogate selection into the
+        // adapter and not ask host component to yield the correct surrogate. This is a fair amount of refactoring, so
```
tsortwehttam, line 130: 👍

packages/@haiku/core/src/adapters/angular-dom/HaikuAngularDOMAdapter.ts
```diff
@@ -95,24 +111,67 @@ export default function HaikuAngularDOMAdapter (
...
+            const div = document.createElement('div');
+            node.parentNode.replaceChild(div, node);
+
+            node.style.visibility = 'hidden';
+            if (surrogate.nativeElement) {
+              div.appendChild(surrogate.nativeElement);
+            }
+
+            window.requestAnimationFrame(() => {
+              element.__placeholder.surrogate = surrogate;
```
tsortwehttam, line 156: Note to self: In my branch I have moved `__placeholder` and such fields into a sub-object.

tsortwehttam:
LGTM and merging, but:

I think we should make it a high priority to get at least one Angular placeholder demo project into `demo/*`. For contrast, the React placeholder logic can be brittle, so it's almost always necessary to review all of its placeholder-related demos when any substantial change is made in core.

stristr:
Yeah, agreed…unfortunately Angular 2+ vanilla JS support is was abandoned before 2.0.0 came out of beta, and you can't really build an Angular code base in a sane way without TypeScript support + `@angular/cli`. I haven't really looked into this beyond the point where anyone would recoil in horror.

I do think Google provides some UMD bundles (e.g. in `node_modules/@angular/core/bundles/core.umd.min.js`), so the most straightforward solution is likely to hack into these.

PR #618, opened 2018-07-19
chore: move actions editor to creator 🙌 

roperzh:
OK to merge.

Short review.

Summary of changes:

The title says it all, I know I have a ton on my plate but I _needed_ to do this before the code locking of Friday so I took a bit of my free time.

Regressions to look for:

- Actions editor in general, specially opening and closing it and how it interacts with the rest of the UI.

Completed checkin tasks:


stristr:
This is great!

From some light testing, everything looks and works great _except_ it seems that whenever I try to add a frame event handler, the modal thinks I'm trying to add an action on the component root element. Confirmed this on an existing and a new project. It even opens the "Main actions" when I click an existing frame event handler on `percy`.

Also: is there any reason for `monaco` to still be in Glass? At a glance, it looks like we can rip out that `haiku-glass` dependency and the convoluted `monaco` bootstrapping routine in `packages/haiku-glass/index.html` as well. Apologies if I'm mistaken, as I'm not very familiar with our sordid history with `monaco`…but I have had some "fun" trying to wrangle it into a webpack build, and the fewer places it needs to be, the better!

stristr:
Additional random UX musing (no action required/expected/suggested, but we might want to backlog these if you agree): I really want `esc` to close the modal and keep thinking it's frozen when it doesn't. Developer audience :smile:

roperzh:
Thank you very much for taking the branch for a spin! 

> I try to add a frame event handler

Good catch, this should be fixed now with afff9d6 

> At a glance, it looks like we can rip out that haiku-glass dependency and the convoluted monaco bootstrapping routine in packages/haiku-glass/index.html as well. 

You are 100% right, the monaco is gone away from glass 🎵 

> I really want esc to close the modal and keep thinking it's frozen when it doesn't. 

I absolutely agree, and I really want this for every modal in the app, I'd really like to use the `<Modal>` component everywhere and centralize the behavior to open/close it.

roperzh:
Just to overcommunicate, all the fixes are pushed to this branch, so it should be good to merge unless anyone finds something else. :shipit:

taylorpoe:
@roperzh 👏 👏  Sweet! Not being able to close the Actions editor with the timeline open (and large) was probably our most egregious ux blemish!

PR #627, opened 2018-07-24
tweak: on expressions editor navigate before committing value

roperzh:
OK to merge.

Short review.

Summary of changes:

- On expressions editor navigate before committing value

Regressions to look for:

- Not aware of any, we are caching the row and the ms value before navigating so everything should work exactly as before.

Completed checkin tasks:

- [ ] ~Added measurement instrumentation (Mixpanel, etc.)~ (already done)

### DIFF
--- a/packages/haiku-timeline/src/components/Timeline.js
+++ b/packages/haiku-timeline/src/components/Timeline.js
@@ -1277,7 +1277,8 @@ class Timeline extends React.Component {
             <GaugeTimeReadout reactParent={this} timeline={timeline} />
           </div>
         ),
-        (<SimplifiedFrameGrid
+        (
+          <SimplifiedFrameGrid
             key="frame-grid"
             timeline={timeline}
             timelineOffsetPadding={TIMELINE_OFFSET_PADDING}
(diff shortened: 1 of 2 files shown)


roperzh:
Please note that the health check will fail due to a lint error present in `master`, I will wait until another PR with the lint fix is merged and then rebase.

roperzh:
(rebased with the lint fix)

packages/haiku-timeline/src/components/Timeline.js
```diff
@@ -1665,9 +1665,7 @@ class Timeline extends React.Component {
           reactParent={this}
           component={this.getActiveComponent()}
           timeline={this.getActiveComponent().getCurrentTimeline()}
-          onCommitValue={(committedValue) => {
-            const row = this.getActiveComponent().getFocusedRow();
-            const ms = this.getActiveComponent().getCurrentTimeline().getCurrentMs();
+          onCommitValue={(committedValue, row, ms) => {
```
stristr: Reminder (from `jsx-no-lambda` lint rule):

> Creating new anonymous functions (with either the `function` syntax or ES2015 arrow syntax) inside the `render` call stack works against _pure component rendering_. When doing an equality check between two lambdas, React will always consider them unequal values and force the component to re-render more often than necessary.


For sanity, I left the `jsx-no-lambda` rule off `*.js`, on when we switch to on `*.tsx`. But it would be ideal to clean these up as we cross their paths.

Preferred:

```
// (in constructor)
this.onCommitValue = (commitedValue, row, ms) => { ... }
```

```
// (in render() call)
  onCommitValue={this.onCommitValue}
```

roperzh: thanks for the reminder, I'll be on the watch!

stristr:
LGTM for merge after health checks pass.

roperzh:
Thanks for the review!

PR #630, opened 2018-07-25
Fixed error caused by CST when using return in an action handler.

ghost:
OK to merge.

Fortunately cst had a nice option for this!


stristr:
LGTM! Merging.

PR #631, opened 2018-07-25
Figma: handling invalid SVG file names

roperzh:
OK to merge.

Short review.

Summary of changes:

There are reports of crashes on Figma SVGs with slashes on their name
in current prod (yeah...), due to the work done in 44d41821cff02fb296599bf05e5edbff637cec91 the app no
longer crashes, but it doesn't show the asset in the library neither.

This is a step forward, now we are replacing all invalid characters
with the ~`!` character~ `-` character.

What about Illustrator and Sketch?

- Illustrator: sometimes Adobe is not that bad, they are sanitizing
arboard names for us
- Sketch: elements with slashes will not appear in the library, I
haven't tested other invalid combinations, for now this is difficult to
solve because sketchtool doesn't provide an option to sanitize file
names before exporting. We *could* change the import logic to:
  - First do a dump of the sketch file contents to JSON.
  - Then parse the JSON file, sanitize file names and export items
  providing the desired file name.

Regressions to look for:

- Figma imports

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/components/library/AssetList.js
+++ b/packages/haiku-creator/src/react/components/library/AssetList.js
@@ -2,7 +2,7 @@ import * as React from 'react';
 import * as Radium from 'radium';
 import AssetItem from './AssetItem';
 
-class AssetList extends React.PureComponent {
+class AssetList extends React.Component {
   render () {
     return (
       <div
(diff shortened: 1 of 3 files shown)


stristr:
LGTM, merging.

(Despite the scientific fact that every time code is added to `haiku-serialization/src/utils`, a puppy dies.)

roperzh:
@stristr sorry I missed that one and I'm against the death of puppies, what is the best place to put this kind of logic from now on? I will try to be more careful

stristr:
(I was just kidding, especially since you were adding to an existing file! But in general `haiku-common` works for common utils these days.)

roperzh:
w00t! thanks! :)

PR #634, opened 2018-07-26
Ungrouped svg sync (669888452711074)

jonaias:
WIP.

There are some functions on HaikuNode from other approaches, I'll delete them before merging if they prove unnecessary.

There is still at least 2 bugs. 
- [x] When you resize a element A inside of a group on Figma (and other tools), on Haiku app, the root svg of ungrouped element A isn't resized.
- [ ] When you add an attribute on an element on Figma, sync with Haiku app, remove the attribute on Figma, sync with Haiku, Haiku won't delete this attribute (even if not marked as `edited: true`).


jonaias:
@matthewtoast @stristr , I'd like some input if implementation is reasonable (Mostly to check if I'm in the right path)

I've tried another approaches, found some insurmountable edge cases and had to start over.

tsortwehttam:
This looks pretty close to me, it seems like the right approach. A few notes:

1. The methods you added to `HaikuNode` _probably_ won't be needed in `@haiku/core` (since they are a concern of in-app editing), so I think we should move them into `haiku-serialization/src/bll/Template.js` or some other place. Perhaps we could move them into `haiku-sdk-creator` so they can have types? (Any input from @stristr on location?)

2. I'd love to see one or two unit tests for this. See `haiku-serialization/test/bll/01_ActiveComponent.test.js`. You could use the test `'ActiveComponent.prototype.mergeDesign[2](design)'` as a starting point. Having a unit test for this will help us feel a lot more confident about making changes here. (Please create a new test file, named something like ActiveComponent-grouping-merge-designs.test.js`.)

3. Since this is a particularly brittle part of the code, please test it with a variety of different design assets, and make sure to try it with projects such as `percy`, and `metapoem2`, to make sure we have covered the edge cases that we are likely to see.

stristr:
For TS, please put in `haiku-common/src/mana`. But I agree `Template.js` is a fine place to park this.

`haiku-sdk-creator` is now all about Envoy, so let's not put there.

(My go-to project is `Moto`, specifically its artboard.)

jonaias:
I'm pretty happy with my last commit, one more bug killed.. I'm finally seeing light in the end of the tunnel! never give up :metal: :metal:

roperzh:
> never give up 

💯 ! [`423kk25`](https://bit.ly/IqT6zt) is awesome!

jonaias:
The missing piece is updating bounding box of ungrouped svg elements. To do so, it's necessary to execute `<svg>.getBBox()`, just like is done on ungroup that runs on glass. The problem is that `mergeMana` runs on every process, so we cannot depend on DOM. 

I'm having some trouble figuring out how to best implement it. Sasha gave me some ideas (i.e. use `updateKeyframes` messages), but i'm not sure how to best implement them.. 

I might have a good idea how to solve it.. does updateHook messages happens after executing methods? (i.e. `mergeDesigns`on glass) If so, we can create a function on glass (we would call it on https://github.com/HaikuTeam/mono/blob/9b0bfdc5a1ce703cc0e935f562d36e29065264a8/packages/haiku-glass/src/react/Glass.js#L310-L312)  that traverse all elements with something like:
```
recalculateUngroupedSvgsBB = () => {
    let updateBytecode = false
    for (element of allElements) {
      if (element.isUngroupedRootSvgElement() ){
        const boundingBox = element.getHaikuElement().target.getBBox()

        const svgRootAttributes = {
          width: boundingBox.width,
          height: boundingBox.height,
          style: {overflow: 'visible'},
        }

        element.updateAttributes(svgRootAttributes)
        updateBytecode = true
      }
    }

    // propagate change to all processes
    if (updateBytecode) {
      activeComponent.replaceBytecode(getCurrentBytecode())
    }
  }
}
``` 

Will that work? There are some points that are not clear how to implement, for example, how the equivalent of element `updateAttributes` might reflect to current bytecode



Anyway, can we merge this PR so it can be tested on SLAM? The unique problem (the one described below) is that for ungrouped elements, the bounding box will be wrong on content change (i.e. increasing size)

PR #635, opened 2018-07-26
Deselect other state fields when editing.

ghost:
OK to merge.

Fixes https://app.asana.com/0/743996399535511/757707017946664

### DIFF
--- a/packages/haiku-creator/src/react/components/StateInspector/StateInspector.js
+++ b/packages/haiku-creator/src/react/components/StateInspector/StateInspector.js
@@ -66,6 +66,7 @@ class StateInspector extends React.Component {
       sceneName: 'State Inspector',
       statesData: null,
       addingNew: false,
+      editingStateName: null,
     };
   }
 
@@ -188,12 +189,12 @@ class StateInspector extends React.Component {
 
   openNewStateForm () {
     if (!this.state.addingNew) {
-      this.setState({addingNew: true});
+      this.setState({addingNew: true, editingStateName: 'new-row'});
     }
   }
 
   closeNewStateForm () {
-    this.setState({addingNew: false});
+    this.setState({addingNew: false, editingStateName: null});
   }
 
   getHeadingText () {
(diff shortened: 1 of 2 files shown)


packages/haiku-creator/src/react/components/StateInspector/StateRow.js
```diff
@@ -281,18 +279,24 @@ class StateRow extends React.Component {
       <form key={`${this.props.stateName}-state`}
         onMouseOver={() => this.setState({isHovered: true})}
         onMouseOut={() => this.setState({isHovered: false})}>
-        {!this.state.isEditing && !this.props.isNew
+        {!this.props.isEditing && !this.props.isNew
           ? <div style={STYLES.stateWrapper}>
             <div
-              onDoubleClick={() => this.setState({isEditing: true, didEscape: false, editingTarget: 'name'})}
```
roperzh, line 287: Could you please move this and the callback below into a constructor function? for the why and details: https://github.com/HaikuTeam/mono/pull/627#discussion_r204841886

packages/haiku-creator/src/react/components/StateInspector/StateInspector.js
```diff
@@ -188,12 +189,12 @@ class StateInspector extends React.Component {
 
   openNewStateForm () {
     if (!this.state.addingNew) {
-      this.setState({addingNew: true});
+      this.setState({addingNew: true, editingStateName: 'new-row'});
```
roperzh: nit: could you please make an enum with all state names and compare based on that? something like:

```js
const editingStates = {
  NOT_EDITING: 0,
  NEW_ROW: 1,
}
```

roperzh (commented):
LGTM! only nit functionality wise: if you click the `+` button twice there's a small flicker, but not a big deal

packages/haiku-creator/src/react/components/StateInspector/StateInspector.js
ghost: These aren't actually the states, it's the name of the state property that the user defined, with the exception of `new-row` which was set by someone on line 240.

Maybe `editingStateName` isn't a good variable name, but my intention was for it to be `null` when the user is not editing anything, and `"myCustomStateProp"` when the user is editing `myCustomStateProp`.

roperzh: ah, that makes sense, feel free to disregard this

ghost:
@roperzh I simplified the state and moved those callbacks into the constructor, let me know if that looks right to you.

I think the flicker is caused by `handleClickOutside`, which is a can of worms I don't feel like opening right now...

roperzh:
LGTM! :shipit:

PR #637, opened 2018-07-27
Ergonomics (brush clearing)

stristr:
OK to merge.

Short review.

Summary of changes:

- Upgrade `yarn`.
- Implement interface `any` types in API.
- Consolidate 3D migration (oddly set at `3.5.2` to the existing `3.5.1` migrations.
- Expand bytecode shorthand during pre-migration.

Regressions to look for:

- None. Mostly just typing.

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality

PR #638, opened 2018-07-28
Ergonomics

stristr:
OK to merge.

Short review.

Summary of changes:

- Serialize timeline shorthand during ASTification of bytecode.
- Expand timeline shorthand during pre-migration (actually implemented in [earlier PR](https://github.com/HaikuTeam/mono/pull/637/commits/4afcc22f3de0aa71b9a4a5aa723df83b85e2a7a4#diff-d32838e6880700273b8fb0c4a084addeR124)).

Regressions to look for:

- Many seem possible, but this seems to work great. I was concerned about Lottie export, but it actually runs off ActiveComponent as source of truth, which synchronously mounts/migrates bytecode before it's exposed via `getReifiedDecycledBytecode()` and friends. Still should rewrite Lottie export to accept bytecode in this format, but seems fine to skip for now.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/ast/expressionToOASTComponent.js
+++ b/packages/haiku-serialization/src/ast/expressionToOASTComponent.js
@@ -1,4 +1,4 @@
-function expressionToOASTComponent (exp, key) {
+function expressionToOASTComponent (exp, key, keyChain) {
   if (exp === undefined || exp === null) {
     return {
       type: 'NullLiteral'
@@ -40,7 +40,7 @@ function expressionToOASTComponent (exp, key) {
   if (exp.__function) {
     return RFOToFunctionAST(exp.__function, key)
   }
-  if (exp.__value) return expressionToOASTComponent(exp.__value, key)
+  if (exp.__value) return expressionToOASTComponent(exp.__value, key, keyChain)
   if (exp.__reference) {
     return {
       type: 'Identifier',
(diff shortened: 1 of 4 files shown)


stristr:
[Screenshot](https://cl.ly/0f2u0y3t173I)!

tsortwehttam:
One comment: `objectToOAST` and friends are sort of intended to be _generic_ ways of processing ASTs into runtime values and vice versa: I should be able to throw any object/AST at them and get a sensible result. So, putting `keyChain` and `LAYOUT_3D_SCHEMA` into these utilities feels like going astray of that general-purposeness in a way that could surprise. There may be code paths that call these functions where this behavior isn't wanted, so I'm wondering if we could have these accept an options object like `{doAssumeIsBytecodePropertyAndConvertToShorthand: true}`.

But if you've grepped for possible code paths and it seems safe w/o the options object, I wouldn't block merge on this note.

stristr:
So…I thought the same, but then figured the generic AST generator is `babylon.parse`, which we're not using in this code path at all, and that therefore we don't _really_ serve generic AST needs through this code path anyway. i.e. in our app, `bytecodeObjectToAST` is always the wrapping call to `objectToOAST` (including recursive calls that would also receive the same options), so `objectToOAST` would never be called _without_ `doAssumeIsBytecodePropertyAndConvertToShorthand`. I am confident it's not broken due to _these_ reasons (check the comment starting [here](https://github.com/HaikuTeam/mono/pull/638/files#diff-2dcb63055b05efdfebd286528c290035R12) for a full explanation).

(I also figured since the introduction of `keyChain` as an array of keys is O(1) overhead and carefully designed the early exit on `keyChain.length !== 4 || keyChain[0] !== 'timelines'` to get us out as soon as possible.)

PR #639, opened 2018-07-29
Zen

stristr:
OK to merge.

Short review.

Summary of changes:

- A smorgasbord of perf-targeting changes.
- [x] Do not initialize layout when it is not needed.
- [x] Do not set width when it doesn't do anything.
- [x] Keep `cachedTransform` in memory instead of using `isEqualTransformString`.
- [x] Optimize patch routine.

Regressions to look for:

- Check demos carefully :smile:

Completed checkin tasks:


stristr:
(flaky test failures expected)

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -1166,7 +1164,7 @@ export default class HaikuComponent extends HaikuElement implements IHaikuCompon
           continue;
         }
 
-        const propertiesGroup = timelineDescriptor[behaviorSelector];
+        const propertiesGroup = mutableTimelineDescriptor[behaviorSelector];
```
stristr, line 1164: @matthewtoast…I feel like we discussed this line of code and you said "I realize this is wrong, but weird things started to happen when it wasn't there." Can you check if this is still the case? It's working fine for me.

stristr:
WIP stats from this branch re: core perf, looking at share.haiku.ai front page for reference:
- In prod, share-page is running at 60 FPS.
- On `master`, share-page is running at 28 FPS.
- On `zen`, share-page is running at 48 FPS.

Still a lot of room for improvement, but I think we should merge this chunk of work ASAP.

packages/@haiku/core/src/renderers/dom/assignStyle.ts
```diff
@@ -29,9 +25,6 @@ export default function assignStyle (domElement, style, component, isPatchOperat
     const previousProp = domElement.haiku.explicitStyles[key];
     if (previousProp !== newProp) {
       domElement.haiku.explicitStyles[key] = style[key];
-      if (key === 'overflow') {
-        continue;
-      }
```
tsortwehttam, line 34: Did you check this against `butterblast` perf?

packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -2772,123 +2745,199 @@ export const getVanity = (elementName: string, propertyName: string) => {
...
+
+  apply();
+};
+
 export const LAYOUT_3D_VANITIES = {
   // Layout has a couple of special values that relate to display
   // but not to position:
   shown: (_, element, value) => {
-    element.layout.shown = value;
+    withLayout(element, () => {
```
tsortwehttam: Vanities are _very_ hot, and ideally we should keep them minimal in terms of CPU and especially garbage creation. (The way this is written, we will be creating a lot of `()=>{}` every frame.) My preference would be to move `withLayout` call back into `hydrateNode` where it used to be part of `initializeNodeAttributes` which is called before properties are applied. Is there any reason we can't/shouldn't do that?

packages/@haiku/core/src/layout/applyCssLayout.ts
```diff
@@ -20,6 +19,40 @@ function hasExplicitStyle (domElement, key) {
...
+  feTurbulence: true,
+  filter: true,
+  foreignObject: true,
+  image: true,
+  mask: true,
+  pattern: true,
+  rect: true,
+  svg: true,
+  use: true,
+};
```
tsortwehttam, line 54: I see the MDN doc link, but note that all elements with `display: block` or `display: inline-block` do respond to to width/height. I wonder if we should check that instead of using this whitelist?

tsortwehttam (changes requested):
Overall good changes and I can feel the perf difference when running locally.

A few things are broken in the demos that will need to be resolved:

  - shadows/blurs:
    - ShadNo
    - InTheSpace
    - comet-rotation
    - drop-shadow
    - filter-instances
    - mc-blur-shadow

  - offset/origin
    - cardmenu
    - heartstry4 [*and all demos that use it, incl. react-*]
    - moto7

  - auto-preserve-3d
    - threedee-perspective-just-work [see the .png in the folder for reference]

  - box size
    - clickable-square [interestingly, a red square now displays in this demo]
    - controlflow-placeholder-* [the spans are now half their original height]

packages/@haiku/core/src/renderers/dom/assignStyle.ts
tsortwehttam: Answering my own question, I just checked `butterblast` and perf is great. Plays fairly smooth in Chrome even with all three `butterblast` running together. Also this change should fix https://app.asana.com/0/743996399535511/757668381087320

packages/@haiku/core/src/layout/applyCssLayout.ts
stristr: Mmmm, not if they're inside an SVG. You don't think we're capturing all use cases by also whitelisting divs?

I would be okay with doing only these tags in an SVG scope, and all other tags outside an SVG scope. Make sense?

packages/@haiku/core/src/HaikuComponent.ts
stristr: So, those `()=>{}` are inlined and I don't think they actually create garbage.

But even if they do, I would say no, we really shouldn't do this in `hydrateNode`. The whole point of this optimization is only to set up layout if layout properties are actually applied.

I will instead refactor to:

```
ensureLayout(element);
element.layout.shown = value;
```

etc.

stristr:
Cool, so re: demos—

- All shadow/blurs and origin/offset issues addressed by 2f09ef6.
- `auto-preserve-3d` was actually broken in master (by me). I fixed it yesterday but hadn't rebased—see 46b23b2.
- under heading "box size", I didn't see any difference between `clickable-square` and `controlflow-placeholder-*` demos between `master` and my branch—are some things also broken in `master` maybe?

About 2f09ef6—it puts the perf issues back on `butterblast` (not better or worse than before). Good news is, I figured out why they're there, and can explain how to fix (without hacking overflow :smile:)!

Would prefer to merge at this commit if you're satisfied and fix it later. Separate doc coming on `butterblast` perf.

tsortwehttam (approved):

tsortwehttam:
LGTM.

PR #643, opened 2018-07-31
Warn losing transition on group (679402678954430)

jonaias:
OK to merge.

Tried to follow this algorithm: https://app.asana.com/0/743996399535511/679402678954430. The modification is that instead checking only for layout property, check for any property (If I understood correctly, not only LAYOUT_3D_SCHEMA would be lost on group/ungroup)

I used the same logic as action editor for displaying confirm group/ungroup popup.

screenshot_20180731_041104

For reviewers: 
- Is there any function/method to compose `elementHasTransitionOrExpression` method? 
- Is there a better way to set table attributes instead copying and pasting `style={STYLES.tableBorder}` for each cell? 

Completed checkin tasks:


stristr:
Hey @jonaias, I wanted to clarify that it **is** only layout properties, and not other properties, that are lost on group/ungroup. See [here](https://github.com/HaikuTeam/mono/blob/d575e3f7579a369d3c03ca5a51dedd09ba152091/packages/haiku-serialization/src/bll/ActiveComponent.js#L3922) for details on group.

(If that's not working, it's a bug, not a feature.)

I would also suggest that instead of listing the components that are going to lose transitions, we simply say: "Grouping will reset all layout (animations|expressions|animations and expressions) to their current state." Where we fill in the blank depending on what is being lost.

One other suggestion to improve UX is a checkbox that says "Don't show this again". (The Envoy User config would be a great place to store this preference—example [here](https://github.com/HaikuTeam/mono/blob/d575e3f7579a369d3c03ca5a51dedd09ba152091/packages/haiku-creator/src/react/Creator.js#L637).)

stristr:
Regarding `elementHasTransitionOrExpression`—I'm not aware of any existing utilities, but you could use `ac.$instance.mutableTimelines` (it's private in `HaikuComponent.ts`, so you'd want to make it public). It is limited to only things that have expressions, keyframes, or controlFlow, so it's much smaller and guaranteed to have all the things you're looking for. (Note also that it uses a more complete/correct definition of "things that might change"—see `isMutableProperty.ts` for details.)

packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -1945,6 +1945,42 @@ class ActiveComponent extends BaseModel {
...
+    const result = {hasTransition: false, hasExpression: false}
+
+    if (componentId in bytecode.timelines[timelineName]) {
+      const componentTimeline = bytecode.timelines[timelineName][componentId]
+
+      for (const propertyName in componentTimeline) {
+        const propertyTimeline = componentTimeline[propertyName]
+
+        // Check if property has more than one keyFrame ( component has transition )
+        if (propertyTimeline instanceof Object) {
```
stristr: 🖖, but why not use `instanceof Function` below?

jonaias:
I've avoided using `ac.$instance.mutableTimelines` to keep `elementHasTransitionOrExpression` alike other functions in ActiveComponent (e.g. `snapshotKeyframeUpdates`).

tsortwehttam:
_For now_ let's just have the modal show this text:

> Transitions or expressions may be lost when you group these elements. Proceed anyway?

And let's _not_ show the table until we can have some design love. (Less work for now, we can circle back to improve this later.)

jonaias:
I've removed it. It's in commit history if we'd like to restore someday: https://github.com/HaikuTeam/mono/pull/643/commits/1190afeb0840557973433248bd64d6531ba6ffdd#diff-1288142fdddf10ef14cbdd2aafbe88bbL57

stristr:
@jonaias, going to merge this after `offline`. Can you please fix conflicts first?

stristr:
LGTM, merging.

PR #644, opened 2018-07-31
Flutterfast

stristr:
OK to merge.

Short review.

Summary of changes:

- skip our layout system when we're inside an SVG and can directly address width/height
- simplify/speed up migrations (do more in pre-phase when possible)
- express styles/attributes as numbers during mana-ification when they look like numbers
- butterblast -> 60fps.
- share.haiku.ai -> 50fps (progress!)

Regressions to look for:

- Check demos :smile:

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [x] Wrote an automated test for existing and modified functionality

PR #645, opened 2018-08-01
Zennest

stristr:
OK to merge (base branch is `flutterfast` a.k.a. #644).

Medium review.

Summary of changes:

- Death to the virtual DOM of the virtual DOM! This PR carefully rewires child transclusion to happen on original nodes (at `node.__memory.children`, only if necessary).
- Rebirth of the `sortedKeyframeCache`, but with a cleaner approach. Parsed value clusters now look like `{keys: number[], parsee: <what it was before>}`. Huge perf boost from this.
- More types!
- Randomly VS Code rewrote a bunch of import paths from `'./../...'` to `'../...'`. Since this is a pet peeve of mine I was planning to write a lint rule for anyway, I let it slide.
- share.haiku.ai -> 59.9fps* (progress!)

To be added (either here or in follow-up PR), but not yet done here:
 - Get rid of `__memory.patched`, which is now _really_ no longer needed.
 - Do component cache differently. Right now, component cache key calculation and property access is the most expensive work on share page, and this is probably the solution to the last .1fps.
 - Bring back `flatManaTree` in some form, a.k.a. optimize the code paths involving `findMatchingNodesByCSSSelector`. This is a huge drag, especially how it slows down full flush render.

Regressions to look for:

- I'm not really sure how to test if auto-sizing is working, so that's the biggest area where I'm not confident.
- Please check demos :smile:

Completed checkin tasks:


stristr:
Also re: auto-sizing—please also confirm that everything works as expected in Creator.

tsortwehttam (dismissed):
These changes seem to break two things:

Animation of text content:
  gameyo
  events-internal
  controls-text

Procedural control of subcomponent:
  play-subc-from-consumer

stristr:
Thanks for review! `play-subc-from-consumer` was actually me breaking `querySelectorAll` by preventing `visitDescendants` from traversing inside a subcomponent…thought we _didn't_ want this, but was wrong.

Any concerns about my approach/simplification of the text content vanity issue (71a1c37)? I am honestly a little confused how it even worked before…because `expandCached()` was also…cached.

PR #649, opened 2018-08-02
Trost Aug 2

tsortwehttam:
OK to merge.

Assortment of late night improvements.


packages/haiku-plumbing/src/Master.js
```diff
@@ -558,9 +558,26 @@ export default class Master extends EventEmitter {
     return this.loadAssets(cb);
   }
 
+  getBaseFolder (abspath) {
```
jonaias, line 561: A good candidate to go into `haiku-plumbing/src/project-folder/ProjectDefinitions.ts`? I've moved some path/project pure functions there.. If it needs to be used by public stuff (core, cli)  it can be moved to `@haiku/sdk-client/src/ProjectDefinitions.ts`  (Plumbing automatically rexports ProjectDefinitions from sdk client, so we have a internal layer)

PR #651, opened 2018-08-03
fix: wire ESC key to close the actions editor

roperzh:
OK to merge.

Short review.

Summary of changes:

Wire ESC key to close the actions editor

Also, since the logic is implemented at `<ModalWrapper />` level,
this same functionality should be really easy to port into other
modals using it (offline, code save, etc)

Regressions to look for:

- Not aware of any

Completed checkin tasks:

### DIFF
--- a/packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
+++ b/packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
@@ -16,7 +16,7 @@ const STYLES = {
 
 export interface ModalWrapperProps {
   style: React.CSSProperties;
-  onClose: () => any;
+  onClose?: () => null;
 }
 
 const stopPropagation: React.MouseEventHandler<HTMLDivElement> = (event) => event.stopPropagation();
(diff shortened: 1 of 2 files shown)


packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
```diff
@@ -16,11 +16,30 @@ const STYLES = {
...
 
 const stopPropagation: React.MouseEventHandler<HTMLDivElement> = (event) => event.stopPropagation();
 
 export class ModalWrapper extends React.PureComponent<ModalWrapperProps> {
+  constructor () {
+    super();
+
+    // FIXME: I don't belong here, move me to the global scope once we centralize
+    // the key handlers
+    document.addEventListener('keydown', this.handleKeyEvents);
```
stristr, line 30: Is it guaranteed that this component will mount a new instance every time it's mounted?

If not, shouldn't the (temporary) location of this listener be `componentWillMount` and use a bound function instead of the direct function in the constructor?

(TBH the rules of React component reuse confuse me and I don't really know the answer to this question.)

roperzh: I _think_, it's guaranteed because of the pattern we are using to show/hide modals, which is something on this lines:

```jsx
{this.state.showMyModal &&
  <ModalWrapper>
    {contents}
  </ModalWrapper>
}
```

PR #652, opened 2018-08-03
Refactor creator interaction mode

jonaias:
OK to merge

Summary of changes:
Fixes
- Prompt to save when toggling from Code view to Preview Mode (759100037963903), 
- Add option to 'Save' from the "You have unsaved changes" dialog in Code Editor (767130344203355), 
- Loophole allows leaving Code Editor with unsaved changes(767130344203356), 
- User should be allowed to discard changes in code editor(766897892323203) 

Improvements:
 - Created `/home/jonaias/haiku/qa/packages/haiku-ui-common/src/interactionModes.ts` for creator/timeline/glass use (to avoid leaking implementation detail to @core, if it is not the correct place, lmk) :
``` typescript
import {  InteractionMode as CoreInteractionMode} from '@haiku/core/lib/helpers/interactionModes';

export namespace InteractionMode {
  export const GLASS_EDIT = CoreInteractionMode.EDIT; // Default mode, all is editable
  export const GLASS_PREVIEW = CoreInteractionMode.LIVE; // Preview mode
  export const CODE_EDITOR = 2; // Code editor mode
}
```
Now for example, timeline can check if in code editor mode or not. (before it was possible to distinguish only between EDIT or LIVE)

- Switching between modes is **way** more sane now. It can be easily extended for new modes besides GLASS_EDIT, GLASS_PREVIEW or CODE_EDITOR (Also mixpanel tracks code mode now =) )

- If you are in code mode, and change component, code mode is kept (can be easily changed).

- If you are in preview mode, neither code nor design button is highlighted

Regressions to look for:

- Changing modes(code|design|preview) and saving editor contents when changing tabs

Completed checkin tasks:


packages/@haiku/core/src/helpers/interactionModes.ts
```diff
@@ -8,8 +8,9 @@
  */
 // tslint:disable-next-line:no-namespace
 export namespace InteractionMode {
-  export const EDIT = 0; // Default mode, all is editable
-  export const LIVE = 1; // Preview mode
+  export const GLASS_EDIT = 0; // Default mode, all is editable
+  export const GLASS_LIVE = 1; // Preview mode
```
jonaias: I'm aware that these implementation details shouldn't be leaked to core. I'll fix it before OK to merge.

PR #656, opened 2018-08-06
Fix this.getDefaultTimeline autocomplete (700874348545630)

jonaias:
OK to merge.

Unfortunately to set autocomplete type(e.g. `monaco.languages.CompletionItemKind.Text`) , monaco global var should be called inside a function after monaco is loaded (cannot be inside a const, or monaco won't be defined)   

Btw, there is a nice "documentation" property that can be explored in action editor autocomplete, eg: https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-completion-provider-example. We can link ours docs individually by autocomplete (it accepts markdown according to monaco.d.ts).

Another useful feature (for code editor, not action editor) is https://microsoft.github.io/monaco-editor/playground.html#extending-language-services-folding-provider-example, so we can fold sections in a nice way

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/components/EventHandlerEditor/constants.js
+++ b/packages/haiku-creator/src/react/components/EventHandlerEditor/constants.js
@@ -64,5 +64,8 @@ export const AUTOCOMPLETION_ITEMS = [
     label: 'isFinished',
     insertText: 'this.getDefaultTimeline().isFinished()',
   },
-
+  {
+    detail: 'Returns default timeline.',
+    label: 'this.getDefaultTimeline()',
+  },
 ];


roperzh (changes requested):
Nice catch @jonaias ! I have some concerns/requests:

1- I don't think it's necessary to move `AUTOCOMPLETION_ITEMS`, because the object it's not necessarily bound to `monaco` and can be consumed by other resources. However, you have `monaco` available in the `EventHandler` class, and you can set the type in there.
2- I think we need to be consistent with the autocompletion `type`, why `this.getDefaultTimeline()` is type `text` and `this.getDefaultTimeline().isFinished()` type `function`?

IMO the easiest way to solve this, it's to use the same `type` for all functions (doesn't matter if is type `text` or `function`) and just add a new item in the `AUTOCOMPLETION_ITEMS` without the need to move it

roperzh:
Btw, folding sections would be awesome, nice find!

jonaias:
Hey Roberto, I've wanted getDefaultTimeline() autocomplete to be interpreted as text to avoid users getting confused by trying to use getDefaultTimeline() as a function in itself (it is, but for designers might be hard to understand that you receive a reference for default timeline).

But I think you have a good point. I just made it a function as other items.

roperzh (approved):
LGTM!

PR #659, opened 2018-08-06
fix: consolidate new component behavior

roperzh:
OK to merge.

Short review.

Summary of changes:

This includes the following fixes:

- fix: when nothing is selected, new component should be blank
- fix: change the rules to allow components to be created without selection
- feat: add a button to create new components next to the component tabs

Regressions to look for:

- New component creation

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ElementSelectionProxy.js
+++ b/packages/haiku-serialization/src/bll/ElementSelectionProxy.js
@@ -175,10 +175,6 @@ class ElementSelectionProxy extends BaseModel {
     this._originCache.groupOrigin = this.getOriginTransformed()
   }
 
-  canCreateComponentFromSelection () {
-    return this.selection.length === 0 || !this.isSingleComponentSelected()
-  }
-
   isSingleComponentSelected () {
     return this.selection.length === 1 &&
       this.selection[0] &&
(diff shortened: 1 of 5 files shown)


packages/haiku-serialization/src/bll/ElementSelectionProxy.js
```diff
@@ -176,7 +176,7 @@ class ElementSelectionProxy extends BaseModel {
   }
 
   canCreateComponentFromSelection () {
-    return this.selection.length > 0
+    return this.selection.length === 0 || !this.isSingleComponentSelected()
   }
```
roperzh: ⚠️ note for reviewer(s): this changes the logic to determine if a component can be created or not, I implemented this conditional based on what I inferred was the previous logic, but watch out in case I'm being dumb.

stristr: I believe for consistency with what was working before with added ability to conglomerate nothing, you should just remove `canCreateComponentFromSelection` entirely.

Although it seems kind of silly to make a higher-order component from a single component, I don't see a strong reason to forbid it (and we don't forbid it now, and IIRC that was changed on purpose).

Rest LGTM.

stristr: [(yep)](https://app.asana.com/0/680508544654148/704910095489688/f)

roperzh:
This is ready for another review, I have removed `canCreateComponentFromSelection` and friends.

packages/haiku-serialization/src/bll/ElementSelectionProxy.js
roperzh: Thanks for noticing that 👍

stristr:
LGTM, merging.

PR #661, opened 2018-08-08
Fix cover sizing (753779401816390)

jonaias:
OK to merge.

Explanations are given on asana: https://app.asana.com/0/752360003454336/753779401816390. Video summary:
Before: https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480079,
After: https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480077

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -2498,8 +2498,11 @@ const computeAndApplyPresetSizing = (element, container, mode): boolean => {
   const elementWidth = element.layout.sizeAbsolute.x;
   const elementHeight = element.layout.sizeAbsolute.y;
 
-  const containerWidth = container.layout.computed.size.x;
-  const containerHeight = container.layout.computed.size.y;
+  // Some browsers does not work correctly with matrix3d transforms on SVGs
+  // with resulting subpixel rendering, so let's round up the size to avoid
+  // browser problems
+  const containerWidth = Math.ceil(container.layout.computed.size.x);
+  const containerHeight = Math.ceil(container.layout.computed.size.y);
 
   // I.e., the amount by which we'd have to multiply the element's scale to make it
   // exactly the same size as its container (without going above it)
--- a/packages/@haiku/core/src/renderers/dom/getElementSize.ts
+++ b/packages/@haiku/core/src/renderers/dom/getElementSize.ts
@@ -3,9 +3,16 @@
  */
 
 export default function getElementSize (domElement) {
-  const rect = domElement.getBoundingClientRect();
-  return {
-    x: rect.width,
-    y: rect.height,
-  };
+  // Get fractional size when on DOM
+  if (domElement.getBoundingClientRect instanceof Function) {
+    const rect = domElement.getBoundingClientRect();
+    return {x: rect.width, y: rect.height};
+  }
+
+  // Fallback to rounded offsetWidth/offsetHeight
+  if (domElement.offsetWidth !== undefined) {
+    return {x: domElement.offsetWidth, y: domElement.offsetHeight};
+  }
+
+  return {x: 1, y: 1};
 }
--- a/packages/@haiku/core/src/renderers/dom/getElementSize.ts
+++ b/packages/@haiku/core/src/renderers/dom/getElementSize.ts
@@ -3,18 +3,9 @@
  */
 
 export default function getElementSize (domElement) {
-  let x;
-  let y;
-  if (domElement.offsetWidth === undefined) {
-    const rect = domElement.getBoundingClientRect();
-    x = rect.width;
-    y = rect.height;
-  } else {
-    x = domElement.offsetWidth;
-    y = domElement.offsetHeight;
-  }
+  const rect = domElement.getBoundingClientRect();
   return {
-    x,
-    y,
+    x: rect.width,
+    y: rect.height,
   };
 }
--- a/packages/@haiku/core/test/TestHelpers.ts
+++ b/packages/@haiku/core/test/TestHelpers.ts
@@ -42,18 +42,8 @@ const createDOM = (cb) => {
   mount.width = 800;
   mount.height = 600;
 
-  // Trick jsdom into giving us proper layout on mounts so we can write sizing tests.
-  Object.defineProperties(
-    win.HTMLElement.prototype,
-    {
-      offsetWidth: {
-        get: () => mount.width,
-      },
-      offsetHeight: {
-        get: () => mount.height,
-      },
-    },
-  );
+  // Trick jsdom into giving us proper getBoundingClientRect on mounts so we can write sizing tests.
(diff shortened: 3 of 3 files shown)


stristr:
@jonaias, wanted to specifically ask about your comment on Asana:

> Even using getBoundingClientRect() on `@haiku/core/src/renderers/dom/getElementSize.ts`didn't solve, but as it is more correct, I'll suggest we merge it anyway.

Can you please elaborate on this? If it does not solve, I do not think we should change this…except to rewrite with fewer allocations :smile:. Meaning:

```
if (domElement.offsetWidth === undefined) {
  const rect = domElement.getBoundingClientRect();
  return {x: rect.width, y: rect.height};
}

return {x: domElement.offsetWidth, y: domElement.offsetHeight};
```

This is for two reasons:
 - This change is likely crashing tests (does not work with virtual DOM).
 - `getBoundingClientRect()` is expensive and `offsetWidth`/`offsetHeight` is not, so if we really don't need to make this change, let's not.

Does the "after" video show that the issue isn't really fully fixed, only has some certain situations improved? And have you tested with various container widths and zoom levels on different Haiku? Have you considered the matrix calculation in `computeAndApplyPresetSizing`? It seems like the rounding we do here could explain why the values don't always perfectly match. In other words, even though we have "correct" `container.sizeX` and `element.sizeX` and `scaleX` needed to ensure `element.sizeX * scaleX == container.sizeX`, the rounding might be set up such that `container.sizeX` scales up but `element.sizeX` (as rounded when the attribute is applied as `style: <element.sizeX>px`) * `scaleX` (as applied and rounded when the layout matrix is calculated and applied via `transform: matrix3d(...)`) does not _round_ to the same pixel as `container.sizeX` (as rounded when the attribute is applied as `style: <container.sizeX>px`).

This probably really should be tractable as long as we understand the subpixel rendering behaviors of the browser—we're in full control of all the layout math and rounding in each of these cases and should be able to anticipate what will happen downstream.

jonaias:
At the time, `getBoundingClientRect()` didn't solve the issue alone, but with the other commit, it does. Without `getBoundingClientRect()`, we cannot get fractional width as explained by offsetWidth docs (from mozilla docs, also happens on chromium):
```
This property will round the value to an integer. If you need a fractional value, use element.getBoundingClientRect().
```

On master, resuming:
Problem 1: The matrix3d won't be calculated with fractional part (as it uses offsetWidth)
Problem 2: There are subpixel divergences on svg and div rendering.. On https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480077, you can see that the rect widths for both svg(the haiku itself) and div (the div for fork and haikudos) are the same size (on the browser page is showing up to the one hundredth pixel unit precision), but on the right on chrome layer debug tab, you can see that according to browser window width resizing, the svg (the haiku) width is increase or decreased in relation to div.

To solve this we allow getting fraction container width by using only `getBoundingClientRect()`, and we always ceil the value, so the svg is guaranted to be bigger (at max up to 1 pixel -see the video https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480079) than container div. IMHO, this is a browser rendering problem, that we might even consider reporting this bug.

>  It seems like the rounding we do here could explain why the values don't always perfectly match. In other words, even though we have "correct" container.sizeX and element.sizeX and scaleX needed to ensure element.sizeX * scaleX == container.sizeX, the rounding might be set up such that container.sizeX scales up but element.sizeX (as rounded when the attribute is applied as style: <element.sizeX>px) * scaleX (as applied and rounded when the layout matrix is calculated and applied via transform: matrix3d(...)) does not round to the same pixel as container.sizeX (as rounded when the attribute is applied as style: <container.sizeX>px).

What rounding are you talking, about the division to calculate scaleX/scaleY? As per floating point definition, they probably never be the same! And on master, they are guaranteed to not be the same for any fractional container width due use of offsetWidth.

I might being missing something, but on video https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480077 it clearly shows a difference in rendering svg and div using matrix3d.

> Does the "after" video show that the issue isn't really fully fixed, only has some certain situations improved?

I didn't like this solution either, spent yesterday trying to fix it the right way, but always end up with the browser rendering difference between svg and div.

stristr:
> What rounding are you talking, about the division to calculate scaleX/scaleY?

There's a decent amount of rounding that takes place through the entire stack (check out `appleCssLayout` for what happens with `width` and `height`. There are also browser behaviors—for example, when you feed Chrome a CSS transform `matrix3d(...)` with 12 digits after the decimal, it seems to always truncate to 6 digits at all costs—i.e. `.123456789` becomes `.123456`, and `123.456789` becomes `123.456`. There are other tricks like this lurking beneath the surface, and I do think it's worth exploring them to actually ensure the container and the SVG are the same size, (instead of the SVG always being at worst bigger). Seems like if we're going to round the container, we could also round the SVG, for example. If we simply don't allow subpixels, does the problem go away? There are methods in `HaikuElement` that maybe, if they always round (e.g. the base case of `computeSizeX()` and friends) might produce this effect.

jonaias:
> There are other tricks like this lurking beneath the surface, and I do think it's worth exploring them to actually ensure the container and the SVG are the same size,

I found a way to ensure it without any doubt (In the example I've applied only https://github.com/HaikuTeam/mono/pull/661/commits/414dbbfcaaa599fbbd8ceaf5e9001633122518ff patch, so we can have fractional width) : 
screenshot_20180808_234348

In the screenshot it is possible to see that container div, haiku div and background svg all have the same x position and width. Even so, is possible to see that the vertical white line from thumb background are kept showing, indicating again the problem displayed in the former video  https://app.asana.com/app/asana/-/get_asset?asset_id=772405297480079. We are sending the correct dimensions, but the browser isn't rendering correctly.

> If we simply don't allow subpixels, does the problem go away?

One hack would be calculating thumb width on demand and forcing them to be rounded (instead being dynamically calculated as of master) on share-page, but it would solve the problem only for our case. If a user tries 'stretch' or 'cover' with a fractional width/height the problem might appear. Do you prefer this path?

Hey @stristr, I might be missing something, if so, I'm sorry in advance for taking your time.

stristr:
Hey @jonaias, sorry for letting this dangle….

I think we can iterate on this as is (convinced myself there are no serious performance regressions and it does make the problem better, although not perfectly fixed).

We'll just need to fix `getElementSize` to be safe in headless mode so tests can pass.

Something like:

```
if (domElement.getBoundingClientRect instanceof Function) {
  const rect = domElement.getBoundingClientRect();
  return {x: rect.width, y: rect.height};
} else if (domElement.offsetWidth !== undefined) {
  return {x: domElement.offsetWidth, y: domElement.offsetHeight};
}

return {x: 1, y: 1};
```

jonaias:
Np @stristr.. Btw, I really liked the idea of using offsetWidth/offsetHeight as fallback.

jonaias:
I'll fix the test tomorrow morning..

stristr:
LGTM. Merging.

PR #663, opened 2018-08-08
fix: special case dragging for single selected tweens

roperzh:
OK to merge.

Short review.

Summary of changes:

[For single timeline segments, clicking & dragging from edge should change duration rather than position](https://app.asana.com/0/752360003454336/767130344203364)

Regressions to look for:

- Timeline dragging

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [x] Wrote an automated test for existing and modified functionality
- [ ] Added measurement instrumentation (Mixpanel, etc.)

PR #664, opened 2018-08-08
fix: update special casing of single segments dragging for left keyframes

roperzh:
OK to merge.

Short review.

Summary of changes:

A fix for an 'edge' case that I didn't contemplate in #663 : with this we allow the curve to be resized from both frames.

I also included a test for `Keyframe.groupIsSingleTween`

Completed checkin tasks:


stristr:
@roperzh, looks like tests are hung. Trying recheck….

stristr:
RECHECK!

stristr:
Oh, it looks like tests are in fact hung.

(You don't need `t.plan(...)` if you have `t.end()`, and in this case, `t.plan(...)` with the wrong number of assertions in the test is the problem.)

roperzh:
thank you very much :)

stristr:
LGTM, merging.

PR #666, opened 2018-08-09
feat: expand timeline properties when the row is changed on stage

roperzh:
OK to merge.

Short review.

Summary of changes:

As the title says, this fixes [UX tweak: when adjusting element on stage via controls affected properties should expand.](https://app.asana.com/0/752360003454336/695267281537668)

Regressions to look for:

- Not aware of any, but I included this feature behind a flag, just because stage -> timeline mutations always brings unexpected headaches.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-common/config/experiments.json
+++ b/packages/haiku-common/config/experiments.json
@@ -32,5 +32,6 @@
   "CopyPasteTweensWithAccelerators": false,
   "UserFacingDevTools": false,
   "ControlFlowIf": false,
-  "ControlFlowRepeat": false
+  "ControlFlowRepeat": false,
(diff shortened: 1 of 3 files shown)


tsortwehttam:
Going to merge this as-is. But, we should consider relocating this logic into Glass or even into Timeline. Something like this:

```
// if in Glass.js
ac.updateKeyframes(..., () => {
  expandTheRows()
})
```

Or:

```
// if in Timeline
ActiveComponent.on('update-keyframes', () => {
  expandTheRows()
})
```

Such that expanding rows doesn't become a baked-in concern of updating keyframes in all processes. (Note: I'm aware we employ similar sorts of hooks in many places :) )

PR #672, opened 2018-08-09
UI Tweaks

taylorpoe:
OK to merge.

This looks pretty clunkbanger to me. I'm not sure we really score more points listing out the path over putting a [REVEAL IN FINDER] button instead. Let's consider switching. I'm okay with this for now mostly because I feel like this whole UI is due for a revamp.

Summary of changes:
image

[Add proj path to publish modal](https://app.asana.com/0/752360003454336/771482871628185)

[Make Actions modal wider](https://app.asana.com/0/752360003454336/759100037963940)

[Clarify auto-save copywriting](https://app.asana.com/0/752360003454336/759100037963911)

### DIFF
--- a/packages/haiku-creator/src/react/components/EventHandlerEditor/EventSelector.js
+++ b/packages/haiku-creator/src/react/components/EventHandlerEditor/EventSelector.js
@@ -29,7 +29,7 @@ class EventSelector extends React.Component {
           {`
           .react-selectize.root-node,
           .dropdown-menu.tethered {
-            width: 457px;
+            width: 568px;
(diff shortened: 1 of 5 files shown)


roperzh (approved):
LGTM! the actions editor looks awesome with the new size.

One small nit: should we add `cursor: pointer` to the project path in the publish modal? Another alternative could be to use the `ExternalLink` component, which handles exactly this use case.

tsortwehttam:
Let's add @roperzh's suggestions as Asana tasks for a future iteration.

PR #673, opened 2018-08-09
fix: serialize event handler data when copying/pasting elements

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes the cause of ['CRASH: Null event handler written out to bytecode'](https://app.asana.com/0/752360003454336/755630850619782), the task also requests to prevent the app from crashing, but seems like this has already been done as part of something else because the app doesn't crashes anymore.

*edit*: I forgot to mention the cause, this was happening when copying/pasting elements with actions, since the handlers are functions, they need to be serialized on the copy action.

Completed checkin tasks:


tsortwehttam:
👍 good job figuring this one out.

PR #681, opened 2018-08-13
Fix auto-3D bugs with one crazy trick

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes [Weird z-index issues when editing](https://app.asana.com/0/768374296244701/776617977997479/f)
- Fixes [When scrubbing, animated things disappear and reappear](https://app.asana.com/0/768374296244701/776617977997480/f)

Regressions to look for:

- None—although by design, this PR breaks auto-3d when overflow-y is enabled.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/Migration.ts
+++ b/packages/@haiku/core/src/Migration.ts
@@ -172,7 +172,9 @@ export const runMigrationsPrePhase = (component: IHaikuComponent, options: Migra
   const referencesToUpdate = {};
 
   if (bytecode.template) {
-    const autoPreserve3d = component.config.preserve3d === 'auto';
+    // y-overflow + preserve-3d leads to various rendering bugs, so for now, disable when overflow is available.
+    // #FIXME
+    const autoPreserve3d = component.config.preserve3d === 'auto' && component.config.overflowY !== 'visible';
 
     visitManaTree(
       '0',


tsortwehttam:
> this PR breaks auto-3d when overflow-y is enabled

🤔 Doesn't this really mean "breaks auto-3d during editing," given https://github.com/HaikuTeam/mono/blob/master/packages/haiku-glass/src/react/Glass.js#L353 ?

stristr:
Hello from the skyyyyyy ✈️.

Indeed, this _intentionally_ breaks auto-3d during editing, for now, because the alternative is much much worse. I don't think the screenshots I was able to grab during slam really do justice to how borked stage gets without this "fix".

Definitely think it's worth identifying the specific browser bug here and actually fixing (if even possible), but this felt okay for now, based on (1) it's the "most correct" fix we know how to do right now and (2) "things work a little differently on stage" is already a thing.

tsortwehttam:
Makes sense. LGTM.

But, here's one question to ponder: Would it make sense to do the opposite, and, instead, _set overflowY to "hidden"_ in the case that auto-3d is "on" and the user has 3D-transformed elements on stage? I'm thinking to myself: If I am editing something 3D, would I rather see it in 3D at the expense of having overflow-y be hidden, or see it in 2D?

stristr:
(Here's a more convincing demonstration of the extent of the borkage: https://cl.ly/2M3W1e211C0W)

stristr:
My snap judgment was definitely that this is the best of only bad options—in a world without in-points and out-points, overflow is really essential to see the universe of animatable components.

Plus, I'd rather just fix the glaring bug, hope nobody notices the side effect, and find the real solution rather than spending any more time on it now (so going to merge).

PR #682, opened 2018-08-14
Ui fixes aug 13

jonaias:
OK to merge.

 - Added Cmd+Enter and Cmd+S shortcuts to action editor (700874348545634) 
 - Do not remove comments from event handlers (700874348545632) 
 - Shows do not need to save dialog and update it (775033840753851) 
 - Do not change to state inspector or library when exiting preview (776617977997478)

After merging, I'll update QA: 
```
When exiting preview mode:
- Pressing toggle preview button: restore creator widgets (open design + restore left panel) or (open code editor + state inspector)
- Pressing design button: go to desing and restore left panel
- Pressing code button: go to code editor and open state inspector on left panel
```

### DIFF
--- a/packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
+++ b/packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
@@ -1,3 +1,4 @@
+import {isMac} from 'haiku-common/lib/environments/os';
 import * as React from 'react';
 import Palette from './../../Palette';
 
@@ -16,7 +17,9 @@ const STYLES = {
 
 export interface ModalWrapperProps {
   style: React.CSSProperties;
-  onClose?: () => void;
+  onEsc?: () => void;
+  onCmdEnter?: () => void;
+  onCmdS?: () => void;
 }
 
 const stopPropagation: React.MouseEventHandler<HTMLDivElement> = (event) => event.stopPropagation();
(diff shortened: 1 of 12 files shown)


packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
```diff
@@ -35,9 +38,27 @@ export class ModalWrapper extends React.PureComponent<ModalWrapperProps> {
...
+      this.props.onCmdEnter();
+    }
+
+    // Command+s on MAC
+    // Ctrl+s on Windows/Linux
+    if (((isMac() && keyEvent.metaKey && keyEvent.keyCode === 83) ||
+         (!isMac() && keyEvent.ctrlKey && keyEvent.keyCode === 83)) &&
+         this.props.onCmdS instanceof Function) {
+      this.props.onCmdS();
+    }
```
stristr: Nit: could we use `Globals.isSpecialKeyDown()` instead of this logic? I think that's a function you wrote, @jonaias :smile:!

stristr (commented):
Looks very good! Just one thing I noticed…it feels weird to close the action editor modal _and_ get the "no need to save" pop-up when I Cmd+S from the action editor.

Rest LGTM!

stristr:
(And to clarify expectation—although it's not to spec of the ticket, probably just show the notice and don't close the modal. Since the other way around would be much harder because notice is from  Creator but editor is in Glass.)

jonaias:
> (And to clarify expectation—although it's not to spec of the ticket, probably just show the notice and don't close the modal. Since the other way around would be much harder because notice is from Creator but editor is in Glass.)

which modal? the notice "Haiku saves your work automatically"? or event handler editor?

packages/haiku-ui-common/src/react/Modal/ModalWrapper.tsx
jonaias: My bad! ehheehe

stristr:
Sorry, didn't really think of the notice as a modal (it's more like a toast).

So I was saying, Cmd+S with event handler editor should just show "Haiku saves your work automatically" and not close the editor.

jonaias:
Done as talked on slack!

stristr:
LGTM, merging.

PR #685, opened 2018-08-14
Virgin Atlantic

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes [Slice doesn't appear when instantiated (see asset)](https://app.asana.com/0/768374296244701/776617977997485/f)
- Fixes [Removing paths from slice caused bounding box problems](https://app.asana.com/0/768374296244701/776617977997488/f)
- Possibly controversial: suppresses logging args from `pasteThings` and `syncCode` update hooks. I found this to provide a noticeable perf boost.
- Possibly controversial: fixes [Bytecode shows version "0.0.0"](https://app.asana.com/0/768374296244701/759100037963946/f). Possibly controversial because of a now-defunct comment in `Master.js` that read: "[in Master, the websocket should be] a mock, since in theory no Master method will originate here and need to be sent outward to the other clients automatically". There were two problems here. One, a regression (possibly from `haiku generate` pre-work, possibly much older) which was relatively straightforward to fix. The second problem was that bytecode changes made directly on master were destined to never sync—specifically, `Master`'s `Project` model used a mock websocket and `ActiveComponent#writeMetadata` (called during publish) didn't even get written as a project update hook. By "fixing" this (using a real websocket on `Master` and syncing `writeMetadata` via update hook), we avoid the following family of conflicts:
  - `Master` makes direct changes to bytecode.
  - User switches to code editor (looking at stale version of bytecode), makes changes, and saves, clobbering whatever Master had written to disk previously.
- Fixes `offline` regression that prevented user from retrying a failed publish without reopening the project.

Regressions to look for:

- Any potential side effects from Master having a live WebSocket instance, although this seemed fine from some light testing.

Completed checkin tasks:


tsortwehttam:
LGTM.

PR #694, opened 2018-08-16
Trost Aug 16 Fixes I

tsortwehttam:
OK to merge.

Short review.

Summary of changes:

- See commits.

### DIFF
--- a/packages/@haiku/core/src/HaikuElement.ts
+++ b/packages/@haiku/core/src/HaikuElement.ts
@@ -8,7 +8,6 @@ import {
   BytecodeNodeMemoryObject,
   ClientRect,
   ComputedLayoutSpec,
-  HaikuBytecode,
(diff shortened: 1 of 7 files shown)


packages/@haiku/core/src/HaikuElement.ts
```diff
@@ -696,6 +729,10 @@ export default class HaikuElement extends HaikuBase implements IHaikuElement {
     return !!this.subcomponent;
   }
 
+  isSvgPrimitive (): boolean {
+    return !!SVG_PRIMITIVE_NAMES[this.tagName];
```
stristr, line 733: TypeScript will probably let you not do the `!!` (it groks falsiness) and you might want to elide it for microbial perf gains in this hot path.

stristr (approved):
LGTM, with ignorable perf nit noted below.

PR #698, opened 2018-08-16
More timeline!

roperzh:
OK to merge.

Short review.

Summary of changes:

- [Indicate max keyframe by making area beyond max slightly darker](https://app.asana.com/0/768374296244701/759916270411679)
- [The timeline hides the selection when using arrow keys to navigate it](https://app.asana.com/0/768374296244701/783184112396173)

Regressions to look for:

- Not aware of any

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/ScrollView.js
+++ b/packages/haiku-timeline/src/components/ScrollView.js
@@ -34,15 +34,13 @@ class ScrollView extends React.PureComponent {
     }
     if (
       what === 'timeline-frame-range' ||
-      what === 'timeline-max-frame-changed' ||
-      what === 'timeline-frame'
+      what === 'timeline-max-frame-changed'
     ) {
       this.forceUpdate();
     }
   }
 
   render () {
-    const frameInfo = this.props.timeline.getFrameInfo();
     return (
       <div
         ref="scrollview"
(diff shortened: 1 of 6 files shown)


packages/haiku-timeline/src/components/ScrollView.js
```diff
@@ -33,13 +34,15 @@ class ScrollView extends React.PureComponent {
     }
     if (
       what === 'timeline-frame-range' ||
-      what === 'timeline-max-frame-changed'
+      what === 'timeline-max-frame-changed' ||
+      what === 'timeline-frame'
```
stristr: @roperzh—please correct me if I'm wrong, but I believe the following are true:
 - `'timeline-frame'` is emitted every time the timeline frame changes (e.g. via timeline playback)
 - this component sits above all timeline rows
 - as a result this will call way more `forceUpdates()` on a component that lives above all timeline rows than before

Isn't this bad?

roperzh: I think you are 100% right and I was very confused, thank you

stristr: (Should I mark as WIP, or is this okay to just remove? Is this fixing "The timeline hides the selection when using arrow keys to navigate it"?)

roperzh: Sorry for not being clear, I will mark as `wip` and fix.

roperzh:
This is ready for another review, apologies in advance for the ugly hack

stristr:
LGTM, merging.

PR #705, opened 2018-08-19
Respect my integritah

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes [Integrity-related crashes easily triggered by Sketch mergeDesigns](https://app.asana.com/0/768374296244701/780886183120361/f)
- _Probably_ fixes [When logging into my "freeuserstaging" free account for a second I got the "incorrect data" vibration but it did log in.](https://app.asana.com/0/768374296244701/785020185816351/f)

Regressions to look for:

- Mostly a straightforward bugfix. `mergeDesigns` ticket was actually a misnomer…in that it was _always_ broken since putting a real WS in Master due to the hack previously used to emit mergeDesigns out to webviews. (@matthewtoast—are there any other hacks like this hiding in the weeds?)
- Small chance that the longer default Envoy timeout could hang the app when the Envoy server is down…but the Envoy server shouldn't go down without the app going down. The change is to accommodate network latency, which is likely what caused the issue @agustinafeijoo reported. Since `ECONNRESET` should return quickly when the Internet is actually down, I'm not seriously concerned about this.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -858,7 +858,7 @@ export default class Creator extends React.Component {
     );
 
     // Use a lengthy timeout for project channel, which does some CRUD.
-    this.envoyClient.get(PROJECT_CHANNEL, {timeout: 60000}).then(
+    this.envoyClient.get(PROJECT_CHANNEL).then(
 
       (project) => {
         this.envoyProject = project;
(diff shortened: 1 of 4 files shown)


tsortwehttam:
To my knowledge there aren't any other hacks like the one for `mergeDesigns`.

LGTM, merging.

PR #707, opened 2018-08-20
Calculate selection BB on demand (759100037963956)

jonaias:
WIP.

 - `Element.getComputedLayout` calculates bounding box on demand for svg objects to solve https://app.asana.com/0/777535458715338/759100037963956
 - I did some refactor so I could use `HaikuElement.computeLayout` code (`computeLayoutWithParentOffset` method )  on `Element.js`
 - The last commit https://github.com/HaikuTeam/mono/commit/d44aa6f2d96fb6f7c6f5dc3cec4f83105f21cf20 enables BB calculation on demand everywhere when `getComputedLayout` is used. Although I've done some benchmarks and didn't find any performance regression, we could revert it, so BB calculation on demand would work only for `ElementSelectionProxy.reinitializeLayout`
 - Some of refactor on `HaikuElement.computeLayout`, made the method cleaner and changed parameter type from `BytecodeNode` node to `LayoutSpec`, so we can have smaller gc for the parameter.
 - Probably this PR deprecate the BB calculation on ungroup (and it will make svg ungroup sync easier).

### Performance benchmark:
 - Master
master

  - BB calculation on demand
bb_on_demand

Due to hardware it is pretty difficult to compare due Turbo boost. (base 1.8ghz, max turbo 4ghz) But a good parameter is that for `Element.getComputedLayout` on both situations the majority of time is spend on `getLayoutSpec`

Regressions to look for:
- Performance issues

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -2410,7 +2410,7 @@ const computeAndApplyLayout = (node, parent) => {
   if (node.layout) {
     node.layout.computed = HaikuElement.computeLayout(
       node,
-      parent,
+      parent ? parent.layout : null,
     );
   }
 };
--- a/packages/@haiku/core/src/HaikuElement.ts
+++ b/packages/@haiku/core/src/HaikuElement.ts
@@ -848,7 +848,7 @@ export default class HaikuElement extends HaikuBase implements IHaikuElement {
       this.getRawBoundingBoxPoints(),
       HaikuElement.computeLayout(
         this.node as BytecodeNode,
-        null, // parentNode; none available here
+        null, // parentLayout; none available here
       ).matrix,
     );
   }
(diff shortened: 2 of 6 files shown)


tsortwehttam (changes requested):
Reviewed and tested locally. Feedback:

A. When I resize the `<ellipse>`, I expect the parent `<svg>` box to not clip the ellipse, but it still clips it: https://cl.ly/0N232J1m1W1L (We agreed to try `overflow: visible` IIRC.)

B. After I resize the `<ellipse>`, when I re-select the `<svg>`, I expect the transform control box to reflect the new size of the inner shapes, but it doesn't: https://cl.ly/0N232J1m1W1L (It only works the second time I click it. You may need to call `clearCaches`/`reinitializeLayout` to force the update.)

C. The refactor looks OK. But, before merging it, I'd like to make sure I fully understand the reason for it. Since this is a very critical code path I'm nervous about merging such a refactor now. Is it necessary for this fix?

jonaias:
Btw, the refactor made on core is pretty safe, the code path didn't changed, only it is called within a method (`HaikuElement.computeLayoutWithParentOffset`) extracted from `HaikuElement.computeLayout`.  But I agree that `Element.getComputedLayout` code change is too much having launch in mind.

This change was needed to make the selection BB grow according to svg contents (basically it overrides output of `HaikuElement.computeLayout` called inside `Element.getComputedLayout` with size/matrix from runtime BB, instead using  `sizeAbsolute` attributes one from `HaikuElement.computeLayout`. To do this override, `Element.getComputedLayout` needed `HaikuElement.computeLayoutWithParentOffset`, the reason  for core refactor ) e.g. 

peek 2018-08-21 04-05

stristr:
Going to close this in favor of #817 which is more promising and simpler.

PR #711, opened 2018-08-20
tweak: ensure critical-testing-path URLs are controllable via ~/.haiku/.env

stristr:
OK to merge.

Short review.

Summary of changes:

- As title implies, ensure `~/.haiku/.env` can be used to output correct URLs, both in the main process and the renderers.

Regressions to look for:

- None expected.

Completed checkin tasks:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,3 @@
 HAIKU_API=http://localhost:8080/
+HAIKU_WWW=http://localhost:8000/
+HAIKU_ACCOUNT=http://localhost:3001/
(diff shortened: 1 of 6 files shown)


stristr:
LGTJ, merging.

roperzh (approved):
LGTM! two tangentially related comments:

- There are other places in the app pointing to `share.haiku.ai` should we update those too?
- Would it be possible to add a similar mechanism to test the cli? (not a big deal since we have the `--api` flag)

stristr:
Re: share.haiku.ai—with the exception of the `TopMenu` stuff, which I skipped (not "critical path") we now get all share/GIF/MP4/index.embed.js/etc. URLs from inkstone. So having the correct inkstone API endpoint is all that's needed (which is nice 👍).

Re: CLI—good idea. Backlogging a task for this.

PR #713, opened 2018-08-20
Jonas' "overflow: visible" fix, sans core refactor and runtime bbox calc

tsortwehttam:
This is just the slice of Jonas' work from https://github.com/HaikuTeam/mono/pull/707 that we need for release, but omitting the core refactor (which may be risky) and the auto-bbox-calc fix (which may also be risky and which Jonas has identified some additional issues with). Should be OK to merge. Short review.

### DIFF
--- a/packages/haiku-serialization/src/bll/Element.js
+++ b/packages/haiku-serialization/src/bll/Element.js
@@ -1588,15 +1588,15 @@ class Element extends BaseModel {
     return HaikuElement.findOrCreateByNode(this.getLiveRenderedNode())
   }
 
-  getParentSvgElement() {
-    let currElem = this; 
-    while(currElem) {  
-      if (currElem.getNameString()=== "svg") { 
+  getParentSvgElement () {
+    let currElem = this
+    while (currElem) {
+      if (currElem.getNameString() === 'svg') {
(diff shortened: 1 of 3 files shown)


packages/haiku-serialization/src/bll/Element.js
```diff
@@ -1588,6 +1588,17 @@ class Element extends BaseModel {
...
+  getParentSvgElement () {
+    let currElem = this
+    while (currElem) {
+      if (currElem.getNameString() === 'svg') {
+        return currElem
+      }
+      currElem = currElem.parent
+    }
+    return null
+  }
```
stristr: …I think this is available as `Element.directlySelected.rootSVG` and didn't need to be re-implemented in the `Element` model.

tsortwehttam: Matt actually implemented `.rootSVG` on the `HaikuElement` model; this is for `Element`.

stristr: Yes, but `Element.directlySelected` uses the `HaikuElement` model.

stristr:
LGTM, merging.

PR #716, opened 2018-08-20
Fix monaco autocomplete

jonaias:
OK to merge.

I was able to repro in a consistent way probably the same issue as @stristr, described on https://app.asana.com/0/777535458715338/776617977997482.

By updating monaco version, I can no longer repro it. I took a look at breaking changes **(https://github.com/Microsoft/monaco-editor/blob/master/CHANGELOG.md#0140-10082018)** that we might use and didn't find any.

Completed checkin tasks:


stristr:
LGTM, merging.

PR #722, opened 2018-08-22
Maths fixes

stristr:
WIP—please do not merge.

Working through some demo regressions, e.g. in `butterblast`, that are actually pretty bad. In particular, looking into an upgrade path for corrected rotation ordering.

Short/mathy review.

Summary of changes:

- Simplify `TransformCache`. Should help reduce garbage while doing a lot of on-stage transforms.
- Ensure we explicitly compute composed transforms while scaling. This fixes part of, but not all of, [Scaling with z-rotation is broken](https://app.asana.com/0/777535458715338/788684413119404/f).
- Make `setCurrentActiveComponent` never time out, for long model hydrations. Fixes [User report: app consistently crashes upon loading a project](https://app.asana.com/0/777535458715338/788339780938810/f).
 - Rid our stack of buggy quaternion math and replace it with correct, direct calculation using Euler angles. This fixes [Scaling with z-rotation is broken](https://app.asana.com/0/777535458715338/788684413119404/f), [Z-rotated element gets wrong skew when negatively multi-scaled](https://app.asana.com/0/705068838027680/685895160103537/f), and [Element horizontally flipped in Sketch not rendered as such in Haiku](https://app.asana.com/0/705068838027680/759100037963936/f) (Froggerino). It is basically the holy grail of our math bugs and seems to trace down to an early bug in `mat4-recompose`, the analog to `mat4-decompose` which I suspect Famo.us math was based on. The bug can be reproduced by [this example](https://gist.github.com/stristr/62766d7507ab2db35fc650f2ac564df8)—note how the matrix out is _not_ the same as the matrix in! This may seem like a risky change, but please note the following:
   - Unit tests bear out that the matrices compose as one would expect, in particular with the vanilla use case of `rotation.z`.
   - I confirmed this fixes every weird layout bug with flipped/rotated/etc. layers in Sketch I have ever seen.
   - Rotation + scale is borked without this change (only alternative I'm aware of being is to roll back to `if (this.managesSingleElement()) { … }` code in `ElementSelectionProxy`, which pushes the problem back down to multi-scale where it sat before).
    - ~Although this does alter the composition behavior of X rotation and Y rotation, the end result is a correction where what we had before was subtly wrong. If you compare demo `comet-rotation` before and after d8bdb2e, then look at what the bytecode actually says, it should be clear that the now-behavior is correct. Therefore, even though this does constitute a "breaking change" of sorts, I think it's a net benefit and won't "break projects" like most breaking changes in the layout system would.~

Regressions to look for:

- Please see note above re: how d8bdb2e alters the composition behavior of X rotation and Y rotation. We need to make a judgment on whether this change should go in now for our stable build. ~My gut feeling is we should—what we had before was wrong and probably won't ever be upgradeable. With that said, please do check demos for other issues.~

Completed checkin tasks:

- [X] Did manual testing of interrelated functionality
- [X] Wrote an automated test for existing and modified functionality


packages/@haiku/core/test/unit/layout.parseCssTransformString.test.ts
```diff
@@ -125,13 +125,6 @@ tape(
           'scale.x': 0, // Because the horizon has vanished.
         },
       ],
-      [
-        'rotate3d(0,60deg,120deg);',
-        {
-          'rotation.y': 1.047, // = Math.PI / 3
-          'rotation.z': 2.094,  // = 2 * Math.PI / 3
-        },
-      ],
```
stristr, line 134: PS

I removed this test because:
 - `rotate3d` is not the thing we seem to think it is in our manafication library (see [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate3d)—it takes a vector and an angle, not three Euler angles).
 - The `css-mat4` library has the same bug as `mat4-decompose`…will go back in and clean up the remnants later.

tsortwehttam:
These changes LGTM so far.

One comment/question: Quaternions were included (inherited from Famous) for purposes of avoiding gimbal lock, which I believe is an issue with using Euler angles exclusively. I.e., if we only represent 3D rotation as Euler angles (with not even an internal quaternion), then 3D rotational tweens will often give unexpected results due to gimbal lock. Obviously, this is quite a minor concern at this juncture, considering that 3D is still a minority use case, but I wanted to mention it since I know we are looking ahead to the time when we have a proper "camera" and 3D is more common.

stristr:
To get out our build, I'm going to merge this.

a2b63b6 may need to be reworked/reverted if we can't recover legacy animations with 3D rotation sufficiently and deem this issue a blocker. I'm ambivalent at this point on the subject but will certainly spend some time today trying out our options.

PR #723, opened 2018-08-22
Fix GPU-Related Freezes/System Crashes

tsortwehttam:
OK to merge.

Short review.

Summary of changes:

After many hours of debugging and trial-and-error, I found one safe change that seems to fix the crash: **Don't render shadows/blurs on the on-stage control points. (See commit `39b80ec`.)** To be specific, I hid the shadow-rendering behind an experiment flag.

**Here are all of the other things I tried which did _not_ work:**

- Cache various hot `Element` methods such as `getBoxPointsTransformed` (and friends)
- Cache `Element` method `getComputedLayout`
- Remove `element.cache.clear()` from `ElementSelectionProxy#clearAllRelatedCaches`
- Avoid calling `ElementSelectionProxy#clearAllRelatedCaches` on all `ElementSelectionProxy#applyPropertyValue` and `ElementSelectionProxy#applyPropertyDelta` calls and instead call it surgically
- Comment out call to `updateKeyframes` that results from `ElementSelectionProxy#move`
- Completely disable/comment out all snapping-related logic (including snap-line drawing)
- Don't draw very long snap-lines that go outside of the viewport (the theory being that the long lines might be resulting in oversize GPU buffers) 
- Exclude the `onmouseenter`/`onmouseleave` listener attributes from the `controlPointMana`
- Exclude calls to `ActiveComponent#getContextSize` from `ElementSelectionProxy#getComputedLayout`
- Exclude calls to `HaikuElement#computeContentBounds()` from `ElementSelectionProxy#getComputedLayout`
- Exclude everything to do with "auto"-sizing from `ElementSelectionProxy#getComputedLayout`
- Comment out all "auto"-sizing related logic from `HaikuElement.computeLayout`
- Not include the `<Image>` component in the `primitives-misc-1` testbed project
- Not include any subcomponents in the `primitives-misc-1` testbed project
- Reduce calls to `HaikuElement.findOrCreateByNode`

I.e., the _only_ changes upstream of my `filter`-related fix that seem to _completely_ stop the crashes from occurring are changes that entirely prevent animation of [rapid re-rendering of] the control points on stage. But obviously those changes won't be viable since the result is that the user can move elements but the control box remains in the same place.

So, although we have rendered the control points with these shadows for a few months apparently without issue (?), my conclusion today is that we've now got the perfect storm of GPU-bombarding behavior with the inclusion of snap-lines, and perhaps some other stuff.

### DIFF
--- a/packages/haiku-common/config/experiments.json
+++ b/packages/haiku-common/config/experiments.json
@@ -33,5 +33,6 @@
   "UserFacingDevTools": false,
   "ControlFlowIf": false,
   "ControlFlowRepeat": false,
-  "ExpandTimelinePropertiesFromStageChanges": true
+  "ExpandTimelinePropertiesFromStageChanges": true,
+  "GlassControlPointShadows": false
 }
--- a/packages/haiku-common/src/experiments/index.ts
+++ b/packages/haiku-common/src/experiments/index.ts
@@ -42,6 +42,7 @@ export enum Experiment {
   ControlFlowIf = 'ControlFlowIf',
   ControlFlowRepeat = 'ControlFlowRepeat',
   ExpandTimelinePropertiesFromStageChanges = 'ExpandTimelinePropertiesFromStageChanges',
+  GlassControlPointShadows = 'GlassControlPointShadows',
 }
 
 /**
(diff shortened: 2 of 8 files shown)


tsortwehttam:
Also: Note that `cef1c81` doesn't affect this particular bug one way or another, but should probably be good for perf anyway.

stristr:
LGTM, merging.

PR #725, opened 2018-08-22
Trost Aug 22 Perf Fixes

tsortwehttam:
Note, this includes all commits from #722 (which I needed/wanted while doing these.)

OK to merge, I think. Short review. Summary of changes:

The `MotorCoverCheck` project is our largest functional project to date (bigger than `Metapoem`) and it's a good test case against a few of our internal project bootstrapping routines. Sasha observed two fairly severe problems with it:

1. Loading the project initially takes a long time
2. Switching to preview mode takes a long time (especially the Timeline UI)

Both are due to view-model hydration which run as a result of the particular method calls. In order to make these a lot faster, these commits...:

1. Don't hydrate any `Row` models in any process other than Timeline
2. Don't rehydrate models when a given method is `superficial` i.e. `setInteractionMode`

The result is that loading `MotorCoverCheck` now only takes about 5-10s (as opposed to like 60s), and switching to Preview Mode has more-or-less decent latency when round-robining to the webviews (the UI change is more-or-less immediate).

I believe both of these changes _should_ be safe.


tsortwehttam:
Oops, meant to say this also includes #723

stristr:
LGTM, merging.

PR #732, opened 2018-08-22
Sasha Has Something Against My Commits

tsortwehttam:
[I wasn't supposed to push to Sasha's branch.]


stristr:
LGTM despite what it looks like!

packages/haiku-creator/src/react/components/EventHandlerEditor/HandlerManager.js
```diff
@@ -68,7 +75,11 @@ class HandlerManager {
    * @param {String} oldEventName
    */
   replaceEvent ({id, event, handler, evaluator}, oldEventName) {
-    this.appliedEventHandlers.set(event, {id, handler, evaluator});
+    if (isHandlerEmpty(handler)) {
```
roperzh, line 78: note for future: there is a very similar check in line 49, we should remove or update that one too

PR #734, opened 2018-08-23
Code snippet insertion button outside monaco (759100037963950)

jonaias:
OK to merge after @taylorpoe green light.

@taylorpoe, do you have any input on design side?

peek 2018-08-23 13-08

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/components/EventHandlerEditor/Editor.js
+++ b/packages/haiku-creator/src/react/components/EventHandlerEditor/Editor.js
@@ -9,7 +9,6 @@ const STYLES = {
     backgroundColor: Palette.SPECIAL_COAL,
     fontFamily: 'Fira Mono',
     fontSize: '11px',
-    padding: '3px 13px',
   },
   preamble: {
     borderTopLeftRadius: 7,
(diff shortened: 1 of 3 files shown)


stristr:
I believe the suggestion in the ticket was to keep it inline, but add a gutter on the right side of the monaco textarea where it sits. Is that feasible?

taylorpoe:
I'm not a fan of having this outside of the visible editor. I think to solve the code overlap we should do this.

taylorpoe:
image

jonaias:
@stristr , I've miss-interpreted the asana description, my bad.

Following @taylorpoe  guidelines:

peek 2018-08-23 14-35

stristr:
So…does this solve the issue where I can't see what I'm typing when my cursor is at the end of the line?

It looks like maybe it doesn't, because text is still allowed to flow all the way to the end of the box. Is that right?

jonaias:
Atm is working like vscode. If you scroll to left on a really long line, the cursor will disappear on right, but as soon as you type any keystroke (including arrow keys) the current line position is made visible automatically by monaco. The '+' button to text overlap is gone.

What I think that might help is to make the width of text area smaller (Exactly 30 pixels, the width of the added div). Go for it?

stristr:
No; this is perfect as is (just tried it). I think the GIF just didn't do it justice :smile:

LGTM, merging.

PR #737, opened 2018-08-23
tweak: don't special-color the top-toolbar icons

zackbrown:
OK to merge.

Short review.


roperzh (approved):
LGTM

stristr:
Wow, that code path was probably also really thrashing React when the tools were open…so cool.

PR #739, opened 2018-08-23
More timeline fixes

roperzh:
OK to merge.

Short review.

Summary of changes:

- [Input selection cut at the bottom ](https://app.asana.com/0/777535458715338/792303516188613)
- [Zooming the timeline sometimes seems to break or freeze](https://app.asana.com/0/777535458715338/792303516188617)
- [If I zoom in/out of the timeline a lot, the scroller breaks.](https://app.asana.com/0/777535458715338/792303516188620)

Regressions to look for:

- Not aware of any

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/RowManager.js
+++ b/packages/haiku-timeline/src/components/RowManager.js
@@ -88,13 +88,11 @@ class RowManager extends React.PureComponent {
   }
 
   isCollapsedAndPreviousExpanded (curr, prev) {
-    if (curr && curr.rows && curr.rows.length && prev && prev.rows && prev.rows.length) {
-      if (!curr.rows[0].isExpanded() && prev.rows[0].isExpanded()) {
-        return true;
-      }
-    }
-
-    return false;
+    return (
+      curr && prev &&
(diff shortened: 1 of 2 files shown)


packages/haiku-timeline/src/components/RowManager.js
```diff
@@ -87,6 +87,16 @@ class RowManager extends React.PureComponent {
     return null;
   }
 
+  isCollapsedAndPreviousExpanded (curr, prev) {
+    if (curr && curr.rows && curr.rows.length && prev && prev.rows && prev.rows.length) {
```
stristr: LGTM, but a worthy nit for a hot path in React IMO…

Pretty sure this function body is equivalent to:

```
return (
  curr && prev &&
  prev.rows && prev.rows.length && prev.rows[0].isExpanded() &&
  curr.rows && curr.rows.length && !curr.rows[0].isExpanded()
);
```

(I tried to write that in the best order for evaluating to false early, assuming `isExpanded()` is false more often than it's true.)

PR #743, opened 2018-08-24
Reduce shellshock

tsortwehttam:
~WIP, don't merge yet.~
OK to merge.

For perf until we support proper expandable rows, filter out the "shells" in certain cases:
- Properties that are noisy and not-super-useful (still addable by JIT!)
- Elements that contain a bajillion "shell" rows


stristr:
LGTM, merging.

PR #718, opened 2018-08-21
Aria Tooltips, Button Titles, and Dash New Project Button

tsortwehttam:
OK to merge.

Short review.

Summary of changes:

- It's been bothering me for months that we don't have a 'New Project' thumbnail to lead off the set of project thumbnails. So I added it.

screen shot 2018-08-21 at 10 52 19 am

- A bunch of buttons/icons in the app didn't have any `title="..."` attribute set, so you wouldn't get the explanatory hover text. I added this text in a bunch of places. Makes UX feel slightly more polished and friendly.

- Made some minor copy tweaks to some labels and help text blocks.

PR #746, opened 2018-08-25
POC: use styled aria-label instead of titles

roperzh:
OK to merge.

Short review.

Summary of changes:

@taylorpoe suggested in our last meeting the idea of using tool-tips instead of the titles so we can style and control the delay until they appear.

This is a random idea to avoid using a `<Tooltip />` component, mainly because I have the feeling that will be so cumbersome to add a title in that way that nobody will do it.

So, the idea is to style elements with the attribute `aria-label` and add a tooltip with pseudo elements (this is not an idea of mine at all).

Where this approach falls short:

- A tooltip may have z-index problems, because we are tied to the stacking context of the target element.
- The position of the tooltip is not dynamically updated

If we decide that this is a good idea I can clean up and test this properly.

PS: @taylorpoe agreed to buy a new toy to Jiggs if we get to implement tooltips, so let's push this forward! (_note: he didn't_)

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/package.json
+++ b/packages/haiku-timeline/package.json
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@haiku/core": "3.5.2",
+    "balloon-css": "^0.5.0",
     "classnames": "2.2.5",
     "codemirror": "^5.28.0",
     "color": "^2.0.1",
--- a/packages/haiku-ui-common/src/react/TooltipStyles.tsx
+++ b/packages/haiku-ui-common/src/react/TooltipStyles.tsx
@@ -1,0 +1,206 @@
+import * as React from 'react';
+
+export default () => {
+  return (
+    <style>
+      {`
+        /**
+        * Tooltip Styles
+        */
+
+        /* Base styles for the element that has a tooltip */
+        [aria-label] {
+          position: relative;
+          cursor: pointer;
+        }
+
+        /* Base styles for the entire tooltip */
+        [aria-label]:before,
+        [aria-label]:after {
+          position: absolute;
+          visibility: hidden;
+          opacity: 0;
+          transition: opacity 0.2s ease-in-out,
+            visibility 0.2s ease-in-out,
+            transform 0.2s cubic-bezier(0.71, 1.7, 0.77, 1.24);
+          transition-delay: 0;
+          transform: translate3d(0, 0, 0);
+          pointer-events: none;
+          text-transform: none!important;
+          font-family: 'Fira Sans';
+          line-height: 15px;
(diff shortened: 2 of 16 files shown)


packages/haiku-timeline/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@haiku/core": "3.5.2",
+    "balloon-css": "^0.5.0",
```
stristr, line 25: Unused dep?

packages/haiku-ui-common/src/react/TooltipStyles.tsx
```diff
@@ -0,0 +1,206 @@
+import * as React from 'react';
```
stristr, line 1: If you want to just put this in a stylesheet, e.g. in `haiku-ui-common/assets/common.css`, you could probably just put this in (e.g.) `haiku-glass/index.html` like:

`<link rel="stylesheet" href="../../node_modules/haiku-ui-common/assets/common.css" />`

Should just work.

stristr (commented):
Nice! One note added with an alternative to using React to host the stylesheet.

taylorpoe:
Sweet!

packages/haiku-ui-common/src/react/TooltipStyles.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+        /* Base styles for the entire tooltip */
+        [aria-label]:before,
+        [aria-label]:after {
+          position: absolute;
+          visibility: hidden;
+          opacity: 0;
+          transition: opacity 0.2s ease-in-out,
+            visibility 0.2s ease-in-out,
+            transform 0.2s cubic-bezier(0.71, 1.7, 0.77, 1.24);
```
taylorpoe, line 25: This cu-bez's overshoot feels a bit off to me (in the gif at least). Could we knock this back to just an `ease`?

packages/haiku-timeline/package.json
roperzh: yes, apologies for that!

packages/haiku-ui-common/src/react/TooltipStyles.tsx
roperzh: This is gold, thank you!

packages/haiku-ui-common/src/react/TooltipStyles.tsx
roperzh: that makes sense, will do! thanks!

tsortwehttam:
Closing in favor of #718 , which now contains these commits.

PR #755, opened 2018-08-28
fix: disable distributing single elements

zackbrown:
OK to merge.

Short review.

Summary of changes:

- Disables distribution of single elements.  Is not 100% formally correct, in that distributing of a single element to stage should be equivalent to aligning that single element to stage, but I feel OK about simply disabling single-element distribution, since the user can simply use align to achieve that end.  That's what this commit does.

Completed checkin tasks:

- [x ] Did manual testing of interrelated functionality


stristr:
LGTM.

PR #756, opened 2018-08-28
UnXML

stristr:
OK to merge.

Short review (look at individual commits and health checks hopefully passing :octocat:).

Summary of changes:

- roll a new package, `haiku-vendor-legacy`, which exists to gradually retire packages from `@haiku/core/src/vendor`.
- hoist `xmlToMana`, `convertManaLayout` and friends out of `@haiku/core`, and deprecate support for XML template strings in bytecode.
- Also deprecates `HaikuHTMLRenderer`.
- fixes an unrelated, unfiled (AFAIK) bug where CSS IDs no longer worked in selectors. I noticed this while fixing demos and checking that our instructions in `README.md` actually work (as of 4.0.0 at least, they didn't).
- reduces our core footprint by all of 3KB gzipped and minified…which is still 4KB bigger than the last time we checked! (README.md is updated to say it's `~50KB` for this reason.)

Regressions to look for:

- Anything around these relatively hot code paths—specifically instantiation of massive SVGs with inner layout, ungroup, group, Lottie export. I tested these all and they seem fine.

Completed checkin tasks:


tsortwehttam:
Sad to see `HaikuHTMLRenderer` go - I think it is (will be? could be?) useful. 

But LGTM.

PR #767, opened 2018-08-30
fix: use 'data-tooltip' instead of 'aria-label' for tooltips

roperzh:
OK to merge.

Short review.

Summary of changes:

Instead of assuming that all elements with `aria-label` are intended to be tooltips, this forces elements to also have a `data-tooltip` attribute, preventing clashes with monaco, other libraries and ourselves if we put more efforts into a11y in the future.

Regressions to look for:

- None

Completed checkin tasks:


stristr:
LGTM.

PR #780, opened 2018-09-04
Refactor: Use state machine to switch between dashboard/loader/editor

jonaias:
OK to merge.

When doing https://app.asana.com/0/785272717735380/802753785269501, I spotted some state duplication/overlap that could be simplified. I went for it, as a fun exercise for holiday ahhahahahah

The refactoring uses a state machine (`userDisplay` state) instead of states `dashboardVisible, doShowProjectLoader, projectLaunching`. The state `isUserAuthenticated` was kind of redundant in relation to 'username' state, so it was also refactored and `INTRO` state was created. 

What do you think about this refactor? I would be happy anyway if you want to cherry pick only https://github.com/HaikuTeam/mono/pull/780/commits/051533ab353c959de87bb0f0cc978f5fc1084575 

Any suggestions for renaming `userDisplay`? I didn't like the name

Regressions to look for:

- Initialization breakage (e.g. not going from intro to dashboard)

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -80,6 +80,19 @@ const FORK_OPERATION_TIMEOUT = 2000;
 const MAX_FORK_ATTEMPTS = 15;
 const FIGMA_IMPORT_TIMEOUT = 1000 * 60 * 5; /* 5 minutes */
 
+// enum USER_DISPLAY {
+//   INTRO,
+//   DASHBOARD,
+//   PROJECT_LAUNCHING,
+//   PROJECT_LAUNCHING,
+//   PROJECT_EDITOR,
+// }
+
+const INTRO = 0;
+const DASHBOARD = 1;
+const PROJECT_LAUNCHING = 2;
+const PROJECT_EDITOR = 3;
+
 export default class Creator extends React.Component {
   constructor (props) {
     super(props);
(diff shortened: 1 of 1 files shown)


tsortwehttam:
@jonaias I like the refactor; I think this makes the dashboard logic easier to understand. But I think @stristr should take glance before merging.

I reviewed, locally tested, merged 051533a separately and merged it in with #781.

tsortwehttam:
(As far as a name, maybe `viewToDisplay` or `displayState` would be clearer.)

stristr (dismissed):
This PR introduces a regression in the critical path "new user flow" as follows:
 - With app shut down, `haiku logout`.
 - Open app and log in.
 - Observed: screen flashes white briefly before showing project browser.
 - Expected: no flash of white.

Weirdly, if you log out and log back in, no flash of white. But if you refresh Creator in a logged-out state, it repros every time.

jonaias:
@matthewtoast, nice suggestion

@stristr, I fixed it, https://github.com/HaikuTeam/mono/pull/780/commits/fd0ca9ea90e73b3eab5916e6642991d21a6751d7#diff-703cb6f028e8cd11de49d43d4e6b040bR2086. To investigate this issued I created setViewToDisplay method, which I utilized to add mixpanel hooks.

There are some states being called twice in a row, so I opted only to hook mixpanel on state changes. In the future we can cleanup them up.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -513,6 +518,15 @@ export default class Creator extends React.Component {
     shell.openExternal(getAccountUrl('checkout'));
   };
 
+  // Controls viewToDisplay state machine
+  setViewToDisplay (viewToDisplay, cb) {
+    if (this.state.viewToDisplay != viewToDisplay) {
```
stristr: `!==` should be used here.

stristr (commented):
LGTM except for note about `!==`. Please merge after fixing that.

PR #787, opened 2018-09-06
Restore directly selected elements on undo/redo (801712328420396)

jonaias:
OK to merge.

About the review given on another PR, I tested undo independently and didn't catch that mentioned case of deleting. Anyway, thanks @stristr for review, it is more robust now.

Summary of changes:

Regressions to look for:

- Any direct selection bug on undo/redo

Completed checkin tasks:

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -958,10 +958,21 @@ export class Glass extends React.Component {
   handleUndo (payload) {
     if (this.project) {
       mixpanel.haikuTrack('creator:glass:undo');
-      Element.unselectAllElements({component: this.getActiveComponent()}, {from: 'glass'});
+      const directlySelectedComponentId = Element.directlySelected && Element.directlySelected.componentId;
+      const component = this.getActiveComponent();
+      Element.unselectAllElements({component}, {from: 'glass'});
       this.project.undo({}, {from: 'glass'}, () => {
         // Important: purge the element selection proxy so that our box points can be reestablished.
         ElementSelectionProxy.purge();
+
+        // Restore directly selected element if it still exists
+        if (!directlySelectedComponentId) {
+          return;
+        }
+        const selectedElement = component.findElementByComponentId(directlySelectedComponentId);
+        if (selectedElement) {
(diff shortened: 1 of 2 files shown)


packages/haiku-glass/src/react/Glass.js
```diff
@@ -951,21 +951,35 @@ export class Glass extends React.Component {
   handleUndo (payload) {
     if (this.project) {
       mixpanel.haikuTrack('creator:glass:undo');
+      const directlySelectedComponentId = Element.directlySelected && Element.directlySelected.componentId;
       Element.unselectAllElements({component: this.getActiveComponent()}, {from: 'glass'});
       this.project.undo({}, {from: 'glass'}, () => {
         // Important: purge the element selection proxy so that our box points can be reestablished.
         ElementSelectionProxy.purge();
+
+        // Restore directly selected element if it still exists
```
stristr, line 968: LGTM, but would like to request some some perf tuning.

1. `undo`/`redo` are hot and potentially slow paths, and `this.getActiveComponent()` is not free, so we should cache the result of calling `this.getActiveComponent()` above if we use it below.
2. On a similar note, for both undo and redo, we should prefix the block where we try to restore the selection with 
```
if (!directlySelectedComponentId) {
  return
}
```

(because `findElementByComponentId` is also potentially expensive).

I also want to point out a leaky abstraction we should be more sensitive to as we refactor the BLL. **No action required** here, but something to keep in mind for the future: it is not great to embed this business logic inside our React app, and would be better for the action stack to emit on push/pop, and handle this sort of logic in `Element` model directly. (There are other problems here too, like the original implementation of `Element.directlySelected` as a static class property instead of an instance property, which would have required less hacking around in the long run.)

jonaias: Yep, In this line I've being wondering specifically about direct selection/edition.. Direct selection/edition would be nice in another file instead glass.js itself.

stristr: lol, yes, I was just talking to Zack about this yesterday. I plan to do this before long.

stristr:
LGTM, merging.

PR #793, opened 2018-09-07
fix: restore log view.

stristr:
OK to merge.

Short review.

Summary of changes:

- I'm not sure how this was done originally, but it looks like our last two builds have ripped the originating "view" of log messages out of user log files, which makes it much harder to debug user-reported issues based on logs/carbonite/etc. There may be a more elegant way of doing this, but what's here definitely works fine.

Regressions to look for:

- None expected.

Completed checkin tasks:

### DIFF
--- a/HaikuHelper.js
+++ b/HaikuHelper.js
@@ -32,6 +32,7 @@ function go () {
     haiku.mode = 'creator';
   }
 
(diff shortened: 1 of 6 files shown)


tsortwehttam:
LGTM.

PR #797, opened 2018-09-10
 Skip pointer events on locked elements (801712328420395) 

jonaias:
OK to merge.

The suggestion from Asana task to skip pointer events worked great.

During testing I found a strange behavior with lock. When you lock a selected element, it is possible to move/edit it, because it is already selected. I fixed it by unselecting element after locking it (which makes sense, because locked elements are unselectable)

About the changelog, can skipping pointer on locked elements be considered a feature?

Completed checkin tasks:


packages/haiku-glass/src/react/Glass.js
```diff
@@ -316,6 +316,13 @@ export class Glass extends React.Component {
         case 'mergeDesigns':
           Element.directlySelected = null;
           break;
+        case 'setLockedStatusForComponent':
+          // Unselect element after locking it
+          const lockedElement = this.getActiveComponent().findElementByComponentId(args[1]);
+          if (args[2] && lockedElement && lockedElement._isSelected) {
```
jonaias, line 322: It is possible to make args[2] early return, but as it would trigger only for half of setLockedStatusForComponent (only when locked=true), I didn't feel the need. Any other thoughts on this?

tsortwehttam:
@jonaias The pointer-events fix LGTM, so I'm going to merge this. But I noticed that the element on stage _was not_ deselected when I unlocked it. Repro:

1. $ yarn start blank
2. Select element on Stage
3. Click lock symbol in Timeline to lock the element
4. Observe that element on Stage remains selected

But this is lower-priority than the pointer-events fix.

PR #805, opened 2018-09-12
Roberto'n'Sasha timeline perf fun

stristr:
OK to merge after @matthewtoast reviews.

Short review.

Summary of changes:

- Removes perf-destroying call to `Timeline#play` on loop in `Timeline#update`. This seemed to significantly contribute to gradual slowdown of timeline scrubbing (and the app overall) while playback was active, because `play()` calls `update()` and `update()` calls `play()`.

Regressions to look for:

- None expected (we tested all over). But want @matthewtoast to review.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Timeline.js
+++ b/packages/haiku-serialization/src/bll/Timeline.js
@@ -258,6 +258,7 @@ class Timeline extends BaseModel {
 
         if (this.getRepeat()) {
           this.seek(0)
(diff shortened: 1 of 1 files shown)


tsortwehttam (approved):
Although I'm 60% sure the `play()` line was originally put there for a reason, I can't remember what it was any more. In any event, I tested locally and didn't see any issues, so LGTM.

PR #807, opened 2018-09-12
3-DPO

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes [Hoist overflow: hidden from root node to mount](https://app.asana.com/0/803720498732761/776617977997483/f). You can still break auto-3d by messing with `style.overflow` on the root node in bytecode, but I generally hid overflow from JIT menus along with this change (really couldn't think of a good reason we need users to be able to change it, but please correct me if I'm wrong).

Regressions to look for:

- Any unexpected side effects related to `overflow: hidden`.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuContext.ts
+++ b/packages/@haiku/core/src/HaikuContext.ts
@@ -283,21 +283,23 @@ export default class HaikuContext extends HaikuBase implements IHaikuContext {
       if (this.config.position && root.style.position !== this.config.position) {
         root.style.position = this.config.position;
       }
+    }
 
+    if (this.mount) {
       if (this.config.overflow) {
-        root.style.overflow = this.config.overflow;
+        this.mount.style.overflow = this.config.overflow;
       } else {
         if (
           this.config.overflowX &&
-          root.style.overflowX !== this.config.overflowX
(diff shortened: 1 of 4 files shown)


roperzh:
LGTM, but asking for a review from @matthewtoast since he knows this way better.

tsortwehttam (approved):
I'm thinking we should leave `Overflow` as JIT property — _or_ replace it with a higher-order "vanity" property that tells the subcomponent wrapper div to clip overflowing content while also abstracting away the CSS messiness and doing the right thing under the hood. However, I'm still saying LGTM to this since I believe this is a feature that very few individuals have discovered. And, they can still add these properties to via code if they wish.

The ability to control `Overflow` (or `Clip At Edges` or whatever we might call it), is mainly important to gives users the ability to position content outside of the subcomponent edges and then animate it in. For example, consider a list of tweets within a larger layout, where the tweets slide into view. There are workarounds but the ability to clip seems fairly fundamental to me.

I made a task for the overflow stuff here: https://app.asana.com/0/813447917062242/820311888215675

stristr:
Okay, will leave backlogged and merge as is.

PR #808, opened 2018-09-13
feat: allow timeline to handle pasting expressions if a row is selected

roperzh:
OK to merge.

Short review.

Summary of changes:

Implements [ux: pasting with property input selected, but not open to expression, should still paste in the clipboard contents.](https://app.asana.com/0/803720498732761/757672931655208), by allowing to paste in an expression the following kind of values:

- Valid values (`number`, `string`, etc)
- 'Single line' expressions, a la `= $user.mouse.x`
- Wrapped 'Multi line' expressions, a la:
```js
function ($user) {
  return $user.mouse.x
}
```
- 'Multi line' expresions, a la:

```js
const a = 1;
const b = 2;
return a + b;
```

There's a known gotcha related to the webviews losing focus: if the cursor is over the stage, this will simply not work, even if there is a row selected (ie. with the pink outline).

*edit*: one more comment, if the value the user is trying to paste is invalid for some reason, we will simply ignore it.

Regressions to look for:

- Copy/Paste is always messy. I think I implemented it in a safe way, but be on the watch.
- Are we properly handling all kind of trash that the user may paste?

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/ExpressionInput.js
+++ b/packages/haiku-timeline/src/components/ExpressionInput.js
@@ -244,11 +244,11 @@ export default class ExpressionInput extends React.Component {
     this.props.onNavigateRequested(direction, maybeDoFocus);
   }
 
-  getCommitableValue (valueDescriptor, originalDescriptor) {
+  getCommitableValue (valueDescriptor, originalDescriptor, editingMode = this.state.editingMode) {
     // If we are in multi-line mode then assume we want to create an expression as opposed to a string.
     // We get problems if we don't do this like a function that doesn't match our naive expression check
     // e.g. function () { if (foo) { ... } else { ... }} which doesn't begin with a return
-    if (this.state.editingMode === EDITOR_MODES.MULTI_LINE || doesValueImplyExpression(valueDescriptor.body)) {
+    if (editingMode === EDITOR_MODES.MULTI_LINE || doesValueImplyExpression(valueDescriptor.body)) {
       // Note that extra/cached fields are stripped off of the function, like '.summary'
       return {
         __function: {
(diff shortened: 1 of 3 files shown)


packages/haiku-timeline/src/components/ExpressionInput.js
```diff
@@ -678,6 +671,37 @@ export default class ExpressionInput extends React.Component {
...
+  willHandlePasteEvent () {
+    const selectedRow = this.props.component.getSelectedRow();
+    const text = clipboard.readText();
+
+    if (!selectedRow || !text) {
+      return false;
+    }
+
+    const parsedText = text.trim();
+    const mode = parsedText.startsWith('function') ? EDITOR_MODES.MULTI_LINE : EDITOR_MODES.SINGLE_LINE;
```
roperzh: dear reviewer: is this safe/smart enough to do?

packages/haiku-timeline/src/components/ExpressionInput.js
```diff
@@ -678,6 +671,46 @@ export default class ExpressionInput extends React.Component {
...
+    let parsedText = text.trim();
+    let mode;
+    if (parsedText.split('\n').length > 1) {
+      mode = EDITOR_MODES.MULTI_LINE;
+      if (!parsedText.startsWith('function')) {
+        parsedText = getRenderableValueMultiline({body: parsedText});
+      }
+    } else {
+      mode = EDITOR_MODES.SINGLE_LINE;
+    }
```
roperzh, line 691: dear reviewer: although I updated my first implementation, is this good enough to detect multi line cases? Do you think that this piece of code should live somewhere else? Thanks!

tsortwehttam:
LGTM.

PR #812, opened 2018-09-14
Do not crash timeline with out of spec bytecode transition

jonaias:
OK to merge.

The attached project ( https://share.haiku.ai/u/taylor/booptroop/fork ) had an out of spec transition:
``` javascript
        "translation.x": {
          "0": { value: 329.5, edited: true },
          "400": { value: 315.5, edited: true, curve: "linear" },
          "933": { value: 390, edited: true, curve: "linear" }
        },
```
When TransitionBody was created for the second transition, it didn't have next keyframe, so it was crashing timeline. I've tested (Hey @roperzh can you do some testing as you know how to achieve timeline edge cases?),  and the user can edit the element transition without crash after this commit. 

Question: Should we warn the user about the non conformant bytecode?

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -9,7 +9,8 @@
       "Snap lines now have a consistent thickness regardless of the artboard zoom level.",
       "Export options are now disabled outside of a project.",
       "User Menu -> Your Profile now points to the correct location.",
-      "Resolved a breaking change in Lottie Android 2.6.0 causing crashes."
+      "Resolved a breaking change in Lottie Android 2.6.0 causing crashes.",
+      "Fixed a crash caused by forking certain projects."
     ]
   }
 }
--- a/packages/haiku-timeline/src/components/TransitionBody.js
+++ b/packages/haiku-timeline/src/components/TransitionBody.js
@@ -101,9 +101,14 @@ export default class TransitionBody extends React.Component {
       this.teardownKeyframeUpdateReceiver = keyframe.registerUpdateReceiver(this.props.id, (what) => {
         this.handleUpdate(what);
       });
-      this.teardownNextKeyframeUpdateReceiver = keyframe.next().registerUpdateReceiver(this.props.id, (what) => {
-        this.handleUpdate(what);
-      });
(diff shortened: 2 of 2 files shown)


changelog/public/latest.json
```diff
@@ -9,7 +9,8 @@
       "Snap lines now have a consistent thickness regardless of the artboard zoom level.",
       "Export options are now disabled outside of a project.",
       "User Menu -> Your Profile now points to the correct location.",
-      "Resolved a breaking change in Lottie Android 2.6.0 causing crashes."
+      "Resolved a breaking change in Lottie Android 2.6.0 causing crashes.",
+      "Improve project forking robustness with out of spec transitions"
```
stristr: I might reword this as "Fixed a crash caused by forking certain projects." to avoid overwhelming users with TMI.

stristr (approved):
I know you asked for @roperzh, but LGTM :smile:

changelog/public/latest.json
jonaias: Much better :+1:

roperzh (approved):
LGTM! Just a note, some time a go we had the [same crash](https://the-haiku-team.slack.com/archives/C6XKR303S/p1535066067000100) and @stristr solved the underlying issue (which was related to publishing at that time) instead of doing this. Is this a symptom of something else going on or should we be doing this anyway?

stristr:
1. Definitely in favor of doing this anyway (should have done it the last time it was broken). Generally, unchecked deep property access through nullable/user-generated data fields is an accident waiting to happen.
2. This particular problem seems to be a consequence of assuming if a timeline keyframe has a "curve", it also has a "next keyframe". Not sure how this might come to _not_ be the case, but I can think of two ways. The first is putting three keyframes in one timeline row, connecting them with two tweens, and then deleting the final keyframe. The second is typing it in bytecode. Because the second always might happen even if we fix the first, I think we still want/need this fix.

roperzh:
✊ 🔥✊ merge, merge merge! ✊🔥 ✊

PR #816, opened 2018-09-15
fix: do not bind states that are already bound.

stristr:
OK to merge.

Short review.

Summary of changes:

- Fixes @taylorpoe blocker Launch blocker: [Haiku's states (instantiated haiku in React) not working after react-app is built. Does work in react dev locally.](https://app.asana.com/0/803720498732761/821764549032368/f).

Regressions to look for:

- I think this is fine, but worth a peek by @matthewtoast. Basically, this change ensures we don't clobber states downstream of `assignConfig` if the state is already defined, unless the state is coming to us _from_ the config.

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -864,9 +864,10 @@ export default class HaikuComponent extends HaikuElement implements IHaikuCompon
         continue;
       }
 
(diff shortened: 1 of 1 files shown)


stristr:
Gonna merge this to unblock @taylorpoe, but PTAL @matthewtoast.

PR #822, opened 2018-09-18
Multiline expression fixes

roperzh:
OK to merge.

Short review.

Summary of changes:

This PR contains two changes:

**fix: disable edition only in first/last lines of multiline expressions**

Instead of preventing any changes as long as the first and last lines
are selected, this marks those lines as non-editable content.

Before: https://gfycat.com/gifs/detail/LateCleanAnhinga
After: https://gfycat.com/gifs/detail/SlightPertinentAntarcticgiantpetrel

**fix: disable native scrollbars in multiline**

They were showing up only if the user has a mouse connected (not with
trackpad) and are not necessary since the multiline box expands
automatically after the first keystroke.

Regressions to look for:

- Multiline expressions behavior

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -10,7 +10,8 @@
       "Export options are now disabled outside of a project.",
       "User Menu -> Your Profile now points to the correct location.",
       "Fixed a crash caused by forking certain projects.",
-      "Improve code editor save error popup"
+      "Improve code editor save error popup",
+      "Minor fixes to the multiline Expressions editor"
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


tsortwehttam:
Question, what happens if the multi-line expression is like 50 lines long? (Or some number of lines giving it a greater height than the size of the timeline viewport?)

roperzh:
@matthewtoast now it looks exactly as it does with a trackpad, which is... ugly

screen shot 2018-09-18 at 1 44 58 pm

I will add the `hold` label and fix this issue (which I _believe_ exists prior this changes), but besides of that, should we display the scrollbars when the user has a mouse?

tsortwehttam:
No need to hold if the issue existed previously: I was just thinking back to the reason I chose to use the `'native'` setting for the scrollbar in this element originally, and I had some vague memory that it had to do with box height. But I'm probably imagining that.

This is LGTM as long as you have tested for regressions in scrolling. Please merge at your leisure.

roperzh:
Wohoo! In that case, I created https://app.asana.com/0/813447917062242/826648483359957 which I will try to fix this week.

PR #824, opened 2018-09-18
fix: call Globals.allKeysUp when Creator is blur

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes [Actions Editor randomly closing on 'Return' key](https://app.asana.com/0/813447917062242/827178890005381). This feels random but it's very easy to repro:

> 1- Keep the `cmd` key pressed
> 2- Press `tab` and switch to another window, but do it slowly, doing it very quickly does not produce a repro.
> 3- Go back to Haiku and observe how actions editor closes on enter because `Globals.isCommandKeyDown` has the wrong value

Regressions to look for:

- I did a mini-slam to check for unexpected behaviors but everything seems to be right. But: shortcuts and behavior involving `cmd` in general.

Completed checkin tasks:


stristr:
LGTM, merging.

(It is _always_ that. We should remember!)

PR #826, opened 2018-09-19
 Focus code editor when switching to code mode

jonaias:
OK to merge.

Use react refs to allow focusing to monaco editor when changing to code mode.

Completed checkin tasks:


packages/haiku-creator/src/react/components/CodeEditor/CodeEditor.js
```diff
@@ -180,6 +185,7 @@ class CodeEditor extends React.Component {
             closeBytecodeErrorPopup={this.hideBytecodeErrorPopup}
           />}
         <MonacoEditor
+          ref="monacoeditor"
```
roperzh, line 188: nit: string refs are deprecated and are known for causing some issues, if possible please switch to a [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs).

roperzh (commented):
Code LGTM and works fine, but I feel like there's a simpler way to do this, by just calling `this.editor.focus()` in `MonacoEditor.js` when the component is mounted/updated.

If the method I described causes any problems then this LGTM.

tsortwehttam:
Let's circle back later to see if the suggested simpler approach will work; for now, merging.

PR #827, opened 2018-09-19
hack: don't allow timeline draggable elements to be over top controls

roperzh:
OK to merge.

Short review.

Summary of changes:

> ⚠️ Note: this is a really hacky solution, if you think it's bad I can look for other solutions, or we can backlog this task.

This is a really ugly solution for ["Dragged rows in Timeline overlaps
top bar"](https://app.asana.com/0/813447917062242/813447917062247), I ended up implementing this way because:

- We can't play with `z-index` in order to overlap the dragged row,
simply because of how I refactored the timeline, remember that many
things that look like a single block in reality are two different HTML
elements.
- Our library doesn't allow us to set boundaries (it goes against their
[driving philosophy](https://github.com/atlassian/react-beautiful-dnd#driving-philosophy-physicality))
- I tried to play with `overflow: hidden` on the draggable container,
but this breaks our scrolling behavior.

While this is far from ideal, performance shouldn't be heavily affected
because the calculations are only performed on the row that is being
dragged when its position changes.

Apologies for how fragile the timeline structure is for this kind of
scenarios.

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -16,7 +16,8 @@
       "Fixed a bug causing Timeline inputs with numbers in exponential format to display 'NaN' instead of the correct value.",
       "Fixed a bug preventing the time/frame indicator of subcomponents to update.",
       "Fixed a bug causing the Timeline ticker to behave unexpectedly in certain situations.",
-      "Fixed a bug causing the Actions editor to unexpectedly close."
+      "Fixed a bug causing the Actions editor to unexpectedly close.",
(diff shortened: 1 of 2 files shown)


roperzh:
New behavior:

tsortwehttam:
Heh, it _is_ pretty hacky but I think it's worth it since the visual bug that it fixes is pretty ugly. Going to LGTM this; if a cleaner solution comes to mind we'll re-do it.

PR #830, opened 2018-09-20
Eventual consistency

stristr:
OK to merge.

Short review.

Summary of changes:

- Essentially, turns every IPC method into `METHODS_TO_AWAIT_FOREVER` and retires the entire concept of timing out waiting for IPC. This probably fixes a litany of unsolved mysteries, including the entire remaining family of integrity crashes, without any obvious downside that I see. Here's my explanation of why:
  - Every "important thing that should matter" is blocked/forced synchronous and ordered by `Lock.request(Lock.LOCKS.ProjectMethodHandler, ...)`. Although we're no longer doing "don't hang when important new messages are being received", the timeouts were already set up so long that realistically, the message we were abandoning to queue up ability to receive the next one was almost certainly important (i.e. integrity-affecting).
   - With respect to data loss, this should actually move the needle in a positive way by creating a system that is _guaranteed eventually consistent_ for a series of long-running tasks, so long as the user doesn't quit the app with Cmd+Q while master is catching up with what is/was on stage, which was already a possibility. This ultimately should cancel a major cause of "strategic crashes when we're worried about data loss". _In particular_, with `Plumbing#handleClientAction` now not blocking method message dispatch to Master when a webview is taking a long time to respond—or not responding at all, which would be the case when the webview has been destroyed by e.g. backing out to the dashboard—it will continue churning through a maybe backed-up method message queue even after `Master#halt` has been called.

Regressions to look for:

- None expected (see above). The worst thing that can happen is the app hangs for a really long time while IPC is catching up. This is already a thing that might happen, but this change should make it ultimately less destructive.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -789,24 +789,6 @@ export default class Creator extends React.Component {
       }
     });
 
-    this.props.websocket.on('method', (method, params, message, cb) => {
-      // Harness to enable cross-subview integration testing
-      if (method === 'executeFunctionSpecification') {
-        return Project.executeFunctionSpecification(
-          {creator: this},
-          'creator',
-          lodash.assign(
-            {
-              creator: this,
-              project: this.state.projectModel,
-            },
-            params[0],
-          ),
-          cb,
-        );
-      }
-    });
-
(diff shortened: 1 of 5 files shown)


tsortwehttam:
Tested locally and code LGTM.

PR #831, opened 2018-09-20
Avoid capturing mouse position on keyup if mouse was already captured (827578786675323)

jonaias:
OK to merge.

I've managed to end up with a simplified version. 

Another ctrl behavior that might be (or not) odd is:
 - ctrl down
 - mouse down, rotate
 - mouse up
 - try to rotate again without ctrl up (instead you can only scale)
(Basically ctrl detection for rotation in done on edge instead level). Anyway, it would be another task.

Regressions to look for:
 - I've done testing but there might be some edge cases, so It's a good idea for someone else to double check against regressions (Especially against direct selection/snapping).

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -20,7 +20,8 @@
       "Fixed a bug preventing Timeline marquee selection to start from within collapsed rows.",
       "Added better logic to calculate the size of the Timeline multiline expressions editor on different sizes.",
       "Do not allow Timeline rows to be dragged over the controls on the top.",
-      "Focus code editor when switching to code mode"
+      "Focus code editor when switching to code mode",
+      "Fixed wrong result when pressing ctrl during element mouse scaling"
     ]
   }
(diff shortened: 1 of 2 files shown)


tsortwehttam (dismissed):
I am seeing the following error when I drag an element in a specific way:

screen shot 2018-09-20 at 4 30 11 pm

To repro:

1. $ yarn start blank
2. Instantiate circle
3. Click an empty area so that nothing is selected
4. Drag the circle (do this immediately just as your mouse goes down (don't do a separate click-then-drag))
5. Observe toast and error message in Glass console

jonaias:
I didn't test dragging right away, thanks @matthewtoast.  When you start a drag right away, the `handleMouseDown` is executed from former ElementSelectionProxy instance and the `drag` method is executed from newer instance. I've transformed `_shouldCaptureMousePosition`  into static member :(

PR #836, opened 2018-09-24
fix: for Layout weirdness when creating nested groups (834345013882114)

stristr:
OK to merge after health checks pass.

Short review.

Summary of changes:

- Fixes [Layout weirdness when creating nested groups](https://app.asana.com/0/824697491061885/834345013882114/f). The problem was duplicate haiku IDs in the edge case where you do exactly the steps described by @roperzh in his repro! tl;dr, the first "shim" layout div resulting from grouping two elements on an empty artboard will always get the same haiku-id in a project. By splicing out the existing group _before_ instantiating the group mana, we created the possibility of a duplicate ID. By splicing out _after_ instantiation, no collisions are guaranteed.

Regressions to look for:

- (Run group QA checklist against a build with this fix.)

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -4036,6 +4036,7 @@ class ActiveComponent extends BaseModel {
       const timelineName = this.getInstantiationTimelineName()
       const timelineTime = this.getInstantiationTimelineTime()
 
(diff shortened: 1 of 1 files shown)


roperzh:
>  in the edge case where you do exactly the steps described

That's why repro steps are gold!

roperzh (approved):
LGTM, can confirm that this fixes the issue.

PR #839, opened 2018-09-25
Dashboard pagination (828702979228595)

jonaias:
OK to merge.

I'd like some early feedback, especially from @funnypapaya /@taylorpoe  on UX/design, so it's easier to fix the direction if any problem.

peek 2018-09-25 18-42

To devs: I've created a pagination component based on share page, so we could use it in other places in the future. I've opted to use first displayed element instead currentPageNum on `Paginator.tsx` component props, so we can keep the same component as first displayed on window resize.

Todo:
- Wrong width on non complete rows: Done
- Cleanups and rebasing to fix commit messages: Done
- Add transition: Done

Completed checkin tasks:


taylorpoe:
Lookin' very nice!

packages/haiku-creator/src/electron.js
```diff
@@ -125,6 +125,8 @@ function createWindow () {
     title: 'Haiku',
     show: false, // Don't show the window until we are ready-to-show (see below)
     titleBarStyle: 'hiddenInset',
+    minWidth: 700,
+    minHeight: 650,
```
jonaias, line 129: This forces at least 3 projects (+ new project box) to be show on dashboard per page. Also smaller sizes then this tends to makes the application itself stranger. I'm ok to remove this.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -307,6 +320,51 @@ class ProjectBrowser extends React.Component {
     );
   }
 
+  // Calculate how many projects can be displayed according to available dimensions
+  updateDimensions = () => {
```
jonaias, line 331: I didn't find a better way to calculate how many projects can be displayed on dashboard

jonaias:
There are other design details that we can refine later (eg. What should we with a lot of projects that page marks are needs more width then application? should we hide page marks after a certain number of pages?)

packages/haiku-ui-common/src/react/Paginator.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  renderPaginationDots () {
+    const pages = [];
+
+    for (let page = 0; page < this.state.numPages; page++) {
+      pages.push(
+        <a
+          key={`pagination-${page}`}
+          onClick={this.changeFirstItemToDisplay.bind(null, page * this.props.numItemsPerPage)}
```
roperzh: nit: doing `.bind` here is almost exactly the same performance wise as declaring an arrow function, keep in mind that `.bind` returns a _new_ function bound to the passed context every time is called.

If you really want to use `.bind` instead of an arrow function in the class you can do the binding in the constructor 💃  (same applies for all other functions here)

> note: this may very well be a technical debt introduced by me in the share page, apologies if that's the case!

roperzh (commented):
Code LGTM!

Two questions for @taylorpoe:

1- Should we be changing the amount of projects per page at different screen sizes? Seems like users with big screens will still have bad performance, right?
2- Could you please check if the design looks good to you and give your approval? Thanks!

packages/haiku-ui-common/src/react/Paginator.tsx
jonaias: I didn't like it either.. I did it to avoid ts warnings:
screenshot_20180930_200535
Moved back to arrow functions =)

jonaias:
About the @roperzh 1), the performance is better now (my resolution is 2560x1080, so there is a lot of projects per page here. Previously, every project was shown on dashboard, now only part if it is shown per page). caveat: I don't know how many is displayed in retina monitors.

taylorpoe:
@roperzh @jonaias 

> 1- Should we be changing the amount of projects per page at different screen sizes? Seems like users with big screens will still have bad performance, right?

I think so. I'm not sure why perf would be bad on larger screens since it won't ever involve scrolling — which was the only perf-horror I experienced. Was there something else I'm missing?

> 2- Could you please check if the design looks good to you and give your approval? Thanks!

Huge gap here on the bottom 1/3 feels a dramatically space-y on my 13" monitor. I'm sure on a 15" it fits 3 rows. Not sure what to do about it at the moment. I believe we could use a css max-height media query to change each card height to be just a bit shorter if the viewport height is < `900px`. In order to squeeze in 3 rows.

And because there's only 8 shown for me on my small screen, I feel a little resentful about that [new project] card taking up a full project position on every single page. I feel pretty strongly that we should limit that to just the first page.

Other than that the pagination buttons container needs a bit more space on the bottom. On quick devtools peek `margin` nor `bottom` worked but `transform: translateY(-20px)` looked about right.

image

jonaias:
I'm doing last taylor fix.. hold on

tsortwehttam:
LGTM.

tsortwehttam:
@taylorpoe is going to give this one last look.

tsortwehttam:
@taylorpoe gave his LGreatTM so merging.

PR #842, opened 2018-09-27
WIP: UI tweaks/fixes

taylorpoe:
WIP: more coming

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -2165,6 +2165,7 @@ export default class Creator extends React.Component {
             <SplitPanel split="horizontal" minSize={300} defaultSize={'62vh'}>
               <SplitPanel split="vertical" minSize={300} defaultSize={300}>
                 <SideBar
+                  isPro={this.state.privateProjectLimit == null}
                   projectModel={this.state.projectModel}
                   switchActiveNav={this.switchActiveNav}
                   onNavigateToDashboard={this.onNavigateToDashboard}
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -1,4 +1,3 @@
-import {FadingCircle} from 'better-react-spinkit';
 import {remote, shell, ipcRenderer, clipboard, webFrame} from 'electron';
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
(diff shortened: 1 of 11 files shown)


stristr:
FYI, there are some busted dependencies in here. Branch should be rebased to `master`.

packages/haiku-creator/src/react/Creator.js
```diff
@@ -2075,7 +2074,7 @@ export default class Creator extends React.Component {
       );
     }
 
-    if (!this.state.isUserAuthenticated || !this.state.username) {
+    if (!this.state.isUserAuthenticated || !this.state.username || this.state.areProjectsLoading) {
```
taylorpoe: For some reason, adding this conditional causes the error: 
image

I want to show this _same_ loader here for when the projects are loading (instead of the spin kit — which I removed). I don't want to call render of this same component in [that other spot where it was](https://github.com/HaikuTeam/mono/pull/842/files#diff-703cb6f028e8cd11de49d43d4e6b040bL2129) because it causes a hiccup in the playback (as it rerenders the same content). Hoping someone can help me with this piece. Looks like the error is due to the tour loading in the background at this phase, but I can't locate the issue

taylorpoe:
Also note that I did try to use the `<ProjectLoader />` component — which virtually looks the same, but due to the fact that it's in its own webview, its time to glass is too slow for the initial window loading.

packages/haiku-creator/src/react/Creator.js
stristr: Happy to help, but would you please rebase this branch so it has `4.3.0` dependencies first?

(`git fetch && git rebase origin/master` and resolve conflicts.)

stristr: Took a look…this needs a small refactor to work the way you want it to. Basically we need to get rid of `renderStartupDefaultScreen()` and replace it with explicitly putting the loading screen on top of everything while carefully not rendering the things that shouldn't render because their props aren't ready yet. This is a bit delicate, so just LMK when you have it rebased to a mergeable state and I can top it off.

stristr:
Rebased, now starting in on requested refactor.

stristr:
Heads up @taylorpoe, I updated this as requested.

Bad news is…it doesn't really perform well on my supercomputer and seems to achieve jank as the first thing you see when you enter the app. We might want to reconsider "due to the fact that it's in its own webview, its time to glass is too slow for the initial window loading"….

Off the top of my head, things we could do include:
 - fade in the loader when we launch and stick with the webview
 - do a totally different loader

Mocked up to show option 1 isn't totally terrible, although it would be less terrible with design love. https://cl.ly/1fd19bcfed0c

taylorpoe:
@stristr Okay, option one doesn't seem too bad at all. Let's go with that.

stristr:
This is ready for le merge.

stristr:
Note (cc: @taylorpoe), without design guidance to the contrary I refactored to support both styles of loader in the branch before I got involved (small version number near bottom of page for generic loader, big italic "Initializing project…" for project loader).

PR #844, opened 2018-09-28
fix: use [].reduce instead of [].map to avoid null values

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes [User-reported CRASH in Timeline](https://app.asana.com/0/824697491061885/842282329271511).

Given that on certain scenarios we have a race condition where our stack reports element that doesn't exist on stage, our current logic adds `null` items to the list of groups returned by `ActiveComponent#getDisplayableRowsGroupedByElementInZOrder` causing crashes in the timeline when trying to read values from this array.

This doesn't fix underlying issue (we shouldn't have this race in the first place), but since we are already ignoring the race, this makes the hack safer.

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -2948,25 +2948,29 @@ class ActiveComponent extends BaseModel {
 
     const rows = root.getHostedPropertyRows(false)
     const all = [].concat(rows)
-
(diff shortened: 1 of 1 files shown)


stristr:
LGTM, merging. (Nit: although checked, `changelog/public/latest.json` was not actually updated.)

PR #846, opened 2018-09-30
Set sync lock when editing deep properties on timeline (803324025609225)

jonaias:
OK to merge.

I've used the same idea of `options` parameter to set sync lock status as `updateKeyframes` and `updateKeyframesAndTypes` on `createKeyframe`. 

I choose to not try to add undo/redo lock/unlock sync, just like `updateKeyframes` works (`updateKeyframesAndTypes` has this ability due SNAPSHOTTED_UNDOABLES). To to so, would involve some deeper changes. IMHO the ideal option would be a segregated function specifically to lock sync eg. `ActiveComponent::lockSync` instead interlaced with `updateKeyframes`, `updateKeyframesAndTypes` and `createKeyframe`, but with our current architecture, it would introduce more latency when editing deep SVG attributes. What do you think about this point?

Completed checkin tasks:


tsortwehttam:
LGTM, merging.

PR #847, opened 2018-10-01
 Added debugger compound launch config for all haiku processes 

jonaias:
OK to merge.

I often debug on more then one process (e.g. glass, timeline, creator..) simultaneously. Restarting it tend to be very manual/tedious, so I've added a compound configuration to attach to all processes at the same time. Another good thing is that if you are debugging any process, you don't need to think/change the configuration, just set a breakpoint and "attach-all".

The other commit is a long standing drag and drop Linux exclusive bug. I left it printing to console, so it bothers people using Linux debug build, therefore increasing the chances of fixing it on electron upstream fix.

Completed checkin tasks:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,6 +7,16 @@
         "core-demo-server",
         "core-demo-client"
       ]
+    },
+    {
+        "name": "attach-haiku",
+        "configurations": [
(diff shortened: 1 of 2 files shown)


stristr:
LGTM, merging.

PR #854, opened 2018-10-03
Vanilla JS: provide a .zip option (597165063005888)

jonaias:
WIP.

Hey @stristr , I've tried to stick with what you suggested. I've left the `Embed JS` and  `Standalone JS` as development only export Menu options.

TODO: 
 - [x] Finish `bundledZipExporter.ts` 
 - Cleanup and add choose (question below) files to put on .zip
 - [x] (Probably on a separated PR in the future? not now?) use `StandaloneBundlerExporter` and `EmbedBundlerExporter` on publish steps and remove `haiku-plumbing/src/project-folder/createCDNBundle.ts`

Question @taylorpoe :
The zipped files should contain   `.embed.js` + `HaikuCore.min.js` + `index.html`,

``` html
<div id="mount-matthewtrost-BrushyBrush"></div>
<script src="./HaikuCore.4.1.0.min.js"></script>
<script src="./index.embed.js"></script>
<script>
  HaikuComponentEmbed_matthewtrost_BrushyBrush(
    document.getElementById('mount-matthewtrost-BrushyBrush'),
    {loop: true}
  );
</script>
```

or `standalone.js` + `index.html`?

``` html
<div id="mount-matthewtrost-BrushyBrush"></div>
<script src="./index.standalone.js"></script>
<script>
  HaikuComponentEmbed_matthewtrost_BrushyBrush(
    document.getElementById('mount-matthewtrost-BrushyBrush'),
    {loop: true}
  );
</script>
```

Completed checkin tasks:


stristr:
I support `index.embed.js`, which more explicitly accomplishes what @taylorpoe is after (showing you how you can use Haiku in a project, helps you understand what bytecode "is").

packages/haiku-common/src/electron/TopMenu.ts
```diff
@@ -269,6 +270,23 @@ export default class TopMenu {
...
+              click: () => this.emitExportRequest('embedjs', 60),
+              enabled: this.options.isProjectOpen,
+              visible: isDevelopment(),
+            },
+            {
+              label: 'Standalone JS',
+              click: () => this.emitExportRequest('standalonejs', 60),
+              enabled: this.options.isProjectOpen,
+              visible: isDevelopment(),
+            },
```
stristr: I don't think we ever will have occasion for these two items to be available from the export menu, so let's remove these two items.

packages/haiku-common/src/electron/TopMenu.ts
```diff
@@ -269,6 +270,23 @@ export default class TopMenu {
...
+            {
+              label: 'Zipped Standalone',
+              click: () => this.emitExportRequest('standalonezip', 60),
```
stristr: Let's just preserve the convention of `'zip'` (the extension) instead of `'standalonezip'`.

packages/haiku-formats/src/exporters/bundled/createBundle.ts
```diff
@@ -0,0 +1,64 @@
+
+// @ts-ignore
```
stristr, line 2: Let's look into adding typings as devDependencies when possible? I think you might get rollup typings for free if you just upgrade to a later version.

packages/haiku-formats/src/exporters/bundled/embedBundlerExporter.ts
```diff
@@ -0,0 +1,58 @@
...
+          const finalContent = `/** ${autoGeneratedNotice}\n${copyrightNotice}\n*/\n${bundledContents}`;
+          return resolve(finalContent);
+        },
+      );
+    });
+  }
+
+  writeToFile (filename: string, framerate: number): Promise<void> {
+
+    this.generateEmbedBundle().then((content) => {
```
stristr: This code always runs `fse.writeFile(filename, '{}');`. The only reason it might not actually output `{}` is it races the promise from `generateEmbedBundle()` and wins.

packages/haiku-formats/src/exporters/bundled/standaloneBundlerExporter.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  writeToFile (filename: string, framerate: number): Promise<void> {
+
+    this.generateStandAloneBundle().then((content) => {
+      return fse.writeFile(filename, content);
+    }).catch((error) => {
+      logger.error(`[formats] caught exception during bundled StandaloneBundlerExporter: ${error.toString()}`);
+    });
+
+    return fse.writeFile(filename, '{}');
```
stristr: See comment above about how this always runs `fse.writeFile(filename, '{}');`.

stristr (commented):
I actually think we should deprecate `createCDNBundle` in this pass.

It's a very small refactor which should help validate that the migrated bundler works the same as the legacy bundler, and has the added benefit of stripping more code/functionality out of `haiku-plumbing` (always a good thing :smile:). Currently, we produce assets from `haiku-formats` [here](https://github.com/HaikuTeam/mono/blob/master/packages/haiku-plumbing/src/Master.js#L939) and create the bundles [here](https://github.com/HaikuTeam/mono/blob/master/packages/haiku-plumbing/src/Master.js#L1026). Basically, we should just replace the chunk from L1026 with the chunk starting at L939 (actually move it down, since we want bundles to be built _after_ `bumpSemverAppropriately`, `writeMetadata`, and `createProjectFiles`).

jonaias:
The Haikus with image hosted on CDN is not being exported to the zip, otherwise the PR is RDY. Sorry for not being able to fully finish it.

PR #857, opened 2018-10-09
Hackathon fixes

stristr:
OK to merge.

Short review.

Summary of changes:

- Three unrelated changes from bugs encountered during last week's hackathon with OneGraph.
  - Fixes [null-valued opacity crashes core](https://app.asana.com/0/845286613726317/854040000213904/f).
  - Fixes (paid) user-reported issue related to missing raster images after duplicating a project. The new duplicate is simply copying the directory and removing `.git`. This is a battle-hardened technique I use all the time (project bootstrapping sets up the runtime critical configs correctly, and first publish fixes all the necessary metadata via `writeMetadata`).
  - Fixes obscure bug related to ungrouping something that has a defs node with layout on it.

Regressions to look for:

- None expected.

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -17,7 +17,8 @@
       "Improve the performance of the Timeline when many rows are open.",
       "Fixed an issue causing the component name to sometimes appear offset from the artboard.",
       "Set multi-line expression editor tab width to 2 spaces.",
-      "Corrected the position of mouse events when the artboard is zoomed."
+      "Corrected the position of mouse events when the artboard is zoomed.",
+      "Fixed issues with missing raster images after duplicating a project."
     ]
   }
 }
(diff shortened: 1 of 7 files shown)


tsortwehttam:
I think 5d19715 means that the following "shorthand" properties would be removed:

```
'timelines': {
  'Default': {
    'haiku:abc123': {
      'translation.x': 0,
      'someBoolState': false,
      'textContent': '',
      'somethingElse': null,
      'otherThing': undefined
    }
  }
}
```

Or am I misreading or missing something?

stristr:
You are correct—but the intention was to remove `null`, so I fixed this (I think) with 5638301. Or can you think of a reason that `null` is a value we should allow?

The problem is that `typeof null === 'object'` because JavaScript.

stristr:
Added a hotfix for issue reported by @taylorpoe in [www.haiku.ai #73](https://github.com/HaikuTeam/www.haiku.ai/pull/73).

tsortwehttam:
My only (weak) argument is that (in the idiomatic sense) `null` means "this box has nothing in it" whereas `undefined` means "there is no box". The update 5638301 is probably OK, as I'm not aware of any logic _offhand_ that this would break.

stristr:
Hmmm, agreed on semantics. However, my hunch is probably forbidding `null`-valued shorthand is the right thing for our product as it is today, since I know we don't have the correct guardrails in place to preserve number-required schemata like every layout property, which I'm pretty sure would result in lots of `NaN` and maybe exceptions downstream.

tsortwehttam:
LGTM.

PR #860, opened 2018-10-15
fix: reset the redoable stack when we perform a regular action

stristr:
OK to merge.

Short review.

Summary of changes:

- Reset the redoable stack when we perform an action that isn't an undo or redo. This fixes some edge case bugs that seemed to lead to crashes during QA last week.

Regressions to look for:

- None.

Completed checkin tasks:


roperzh (approved):
LGTM

PR #862, opened 2018-10-15
fix: check if a property row should be updated based on its pretty value

roperzh:
OK to merge.

Short review.

Summary of changes:

- Fixes ["Timeline property input shows wrong (cached) value"](https://app.asana.com/0/856556209422928/864034901419621) by checking if the react component should be updated based on its `prettyValue` instead of its `computedValue`.

Regressions to look for:

- none expected

Completed checkin tasks:

- [x] Did manual testing of interrelated functionality
- [ ] Wrote an automated test for existing and modified functionality
- [ ] Added measurement instrumentation (Mixpanel, etc.)
- [ ] Updated `changelog/public/latest.json`

PR #866, opened 2018-10-15
Shells rabbit

stristr:
OK to merge.

Short review.

Summary of changes:

- Small tweak/fix for #862—gives `prettyValue` a normalized structure so that we can always use it like an object.
 - Some minor rabbit/hat work for shells-ui—adds an `IF_ROOT_CHILD` check for property display/row hydration and applies it to layout properties. Also hides some sketchy 2.5D (z-related) properties.

Regressions to look for:

- Please test carefully, but this should simply not populate rows for properties we don't want to display.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Row.js
+++ b/packages/haiku-serialization/src/bll/Row.js
@@ -605,7 +605,9 @@ class Row extends BaseModel {
     let depth = 0
     let parent = this.parent
     while (parent) {
-      depth += 1
+      if (parent.element.hasAddressableProperties) {
+        depth += 1
+      }
       parent = parent.parent
     }
     return depth
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -3664,14 +3664,13 @@ class ActiveComponent extends BaseModel {
 
                 // Not all views necessarily have the same collection of elements
                 if (element) {
+                  element.rehydrateRows()
                   Row.where({ component: this, element }).forEach((row) => {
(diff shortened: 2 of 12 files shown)


stristr:
@roperzh, rebased to `master` and also realized I had to revert your last two commits in the original `shells-ui` merge since #867 makes it no longer needed.

packages/haiku-serialization/src/bll/Row.js
```diff
@@ -605,7 +605,9 @@ class Row extends BaseModel {
     let depth = 0
     let parent = this.parent
     while (parent) {
-      depth += 1
+      if (parent.element.hasAddressableProperties) {
+        depth += 1
+      }
```
stristr, line 610: @roperzh, please holler if you think this is a terrible idea or prone to breaking something important.

The point of this commit is to turn [this](https://cl.ly/02512030746c) into [this](https://cl.ly/1de3bd4589eb). The random-seeming indentation was a function of using "pure depth" to determine display depth—arguably, the better depth to use is "depth of things I can see".

stristr:
Correction re: 

> I had to revert your last two commits

I ended up needing it to prevent weird clearing issues in the timeline, so went with 47ec1d4de588dfa72344f2a8cfe63fedb3c61a39 instead.

packages/haiku-serialization/src/bll/Row.js
roperzh: This is way nicer, thank you.

roperzh (approved):
LGTM and works incredibly well.

stristr:
Okay, going to merge this with caveat that we basically need to QA entire timeline again 🐇 🎩 !

PR #870, opened 2018-10-16
Fix mergeDesigns-related integrity crash

stristr:
OK to merge after sanity check. Discussing with @zackbrown now.

Short review.

Summary of changes:

- Revise integrity hash to be grandchild-agnostic. A longstanding integrity crash we've been seeing is related to the simple fact that sometimes, `mergeDesigns` will change the internal structure of existing template nodes that are instantiated in the template—but in a way that is ultimately irrelevant to what I understand to be the primary purpose of the integrity hash, which is to async generate the same IDs when instantiating children in Haiku.

Regressions to look for:

- We should just SLAM this after getting tests to pass—too many hidden corners to accurately anticipate possible issues.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Template.js
+++ b/packages/haiku-serialization/src/bll/Template.js
@@ -155,7 +155,7 @@ Template.mirrorHaikuUids = (fromNode, toNode) => {
   }
 }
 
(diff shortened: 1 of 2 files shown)


stristr:
LGTJ, merging.

PR #871, opened 2018-10-16
The "trivial" JIT

stristr:
OK to merge.

Short review.

Summary of changes:

- Loosen the rules of what is allowed to display timeline rows, so that if we do have multiple keyframes for an `IF_CHANGED_FROM_PREPOPULATED_VALUE` property, they will always display.
- Meanwhile, greatly decrease the likelihood we will "accidentally" have multiple keyframes for an `IF_CHANGED_FROM_PREPOPULATED_VALUE` property, which (I think) is almost always a side effect of my laziness when implementing on-stage transforms. What we were doing before is creating all these verbose properties whenever a user scales—rotation was a little bit tamer, but the most basic scaling operation will create "trivial" JIT keyframes for the three `shear` properties, `translation.z`, the three `rotation` properties, and so on—despite the fact that a lot of the time, you're only updating `scale.x/y` and `translation.x/y`. This change was actually very simple and addresses the main culprit of the problem that led me to want stricter rules in the first place. Sadly, it does nothing for existing projects, which are going to be thrashed with ugly and extra keyframes. On the bright side, I'm pretty sure these were already there, so they're not necessarily making it worse.

Regressions to look for:

- None, but we'll SLAM! and find them if they're there.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ElementSelectionProxy.js
+++ b/packages/haiku-serialization/src/bll/ElementSelectionProxy.js
@@ -1,7 +1,7 @@
 const path = require('path')
 const logger = require('./../utils/LoggerInstance')
 const BaseModel = require('./BaseModel')
-const {rounded, transformFourVectorByMatrix} = require('./MathUtils')
+const {rounded, transformFourVectorByMatrix, basicallyEquals} = require('./MathUtils')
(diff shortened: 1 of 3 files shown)


roperzh (approved):
LGTM

PR #874, opened 2018-10-19
feat: auto tweening

roperzh:
~🛑 please hold until #858 is out~ OK to merge

Short review.

Summary of changes:

- Retire `Experiment.ExpandTimelinePropertiesFromStageChanges`, as this has been enabled for a while.
- Add auto tweening for newly created keyframes

Regressions to look for:

- none expected

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -3560,6 +3560,21 @@ class ActiveComponent extends BaseModel {
               propertyName,
               true // fallbackToInitialKeyframeIfProvided
             )
+
+            if (experimentIsEnabled(Experiment.AutoTweenNewKeyframes)) {
+              const ms = parseInt(Object.keys(keyframeMoves[timelineName][componentId][propertyName])[0], 10)
+              if (Number.isInteger(ms)) {
+                Bytecode.addDefaultCurveIfNecessary(
+                  bytecode,
+                  timelineName,
+                  Template.buildHaikuIdSelector(componentId),
+                  ms,
+                  propertyName,
+                  componentId,
+                  this.getElementNameOfComponentId(componentId)
+                )
+              }
+            }
           }
         }
       }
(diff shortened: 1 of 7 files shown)


packages/haiku-serialization/src/bll/ActiveComponent.js
```diff
@@ -3666,10 +3667,9 @@ class ActiveComponent extends BaseModel {
                 if (element) {
                   element.rehydrateRows()
                   Row.where({ component: this, element }).forEach((row) => {
-                    if (experimentIsEnabled(Experiment.ExpandTimelinePropertiesFromStageChanges)) {
-                      if (row.property && keyframeUpdates[timelineName][componentId][row.property.name]) {
-                        row.expand(metadata)
-                      }
+                    if (row.property && keyframeUpdates[timelineName][componentId][row.property.name]) {
+                      this.addDefaultCurveFromRawData(row, keyframeUpdates[timelineName][componentId], {from: 'timeline'})
```
stristr: For a couple of reasons, I believe this should go in `updateKeyframesActual` and not the hydrator. Rest LGTM.

roperzh: 👍 quick question: all the other calls to add the default curve are also in the hydrator of the different methods (`moveKeyframes`, `createKeyframe`) should I move those too or your comment is specific to `updateKeyframes`?

stristr: Sorry, I missed this comment. Yes, I did intend for the comment to apply to all examples where we're doing this in the rehydrator. (This is mostly based on the fact that the rehydration is going to run AFTER `softReload` [where caches get cleared] on a `hardReload`, which I expect will have weird playback consequences that are impossible to anticipate.)

I suspect the reason you chose the rehydrator is wanting to use the existing handle on the `Row` model entity rather than having to obtain one twice? That's a good instinct in terms of perf, but I could make a case for a totally static way of handling this that should address perf concerns:

`Bytecode.addDefaultCurveIfNecessary(bytecode, timelineName, timelineTime, propertyName)`

This would mean we _don't_ get to use `row.prev()` or `row.addCurve()`, and we'd use "manual mode" instead (see [here](https://github.com/HaikuTeam/mono/commit/cb6f1833c02c1a254ebdf4f69fa58452b3824ee7#diff-07e8365fb2fe8ddaffb3b2d33221b187R2390) for one reference implementation).

TBH, this is a tough call, but I am biased toward better perf/duplicative throwaway code which is readable and unambiguously correct. i.e. `Bytecode.addDefaultCurveIfNecessary`.

roperzh:
@stristr thanks for the comments, I think this is ready for another review.

stristr:
LGTM, merging.

PR #875, opened 2018-10-22
fix: safely check if a path exists before trying to read its properties

roperzh:
OK to merge.

Short review.

Summary of changes:

- Safely check if a path exists before trying to read its properties (fixes ["Direct selection bug: TypeError: Cannot read property 'closed' of undefined"](https://app.asana.com/0/856556209422928/875178733925060)). To me this seems like a classic JS mistake and the proposed solution is enough, but if you suspect this is the symptom of something deeper please disregard this PR 🐎 

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -4,7 +4,7 @@
       "FIRST!"
     ],
     "Fixes": [
-      "FIRST!"
+      "Fixes crashes related to direct selection on specific SVG assets."
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


stristr:
LGTM. While we're at it, let's also fix the outer loop in that method `buildPathLUT`. It's set up to crash on similar `TypeError` if NOT `points[0].moveTo`, so we should just start the loop from `let i = 1`, no?

roperzh:
That makes total sense, fix pushed.

stristr:
LGTM, merging.

PR #876, opened 2018-10-23
fix: do not always delete state values before modifying them

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes ["Undo/Redo + States bug: TypeError: Cannot set property 'edited' of undefined"](https://app.asana.com/0/856556209422928/876867697507733) by avoiding to delete state values when creating a new value.

**Why this fixes the issue?** 

The logic to update the name of a state is a bit weird, instead of updating the name of the state, we are:

1- Deleting the state with the old name
2- Creating a new state with the new name and the old value

This logic was also used to create new states, causing step `1` to add an invalid inversion method in the undo stack.

With this change, we are just doing `1` and `2` if the state name has changed AND we are not dealing with a new state

Regressions to look for:

- None expected, but a quick look at states CRUD will be helpful.

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -4,7 +4,8 @@
       "FIRST!"
     ],
     "Fixes": [
-      "Fixes crashes related to direct selection on specific SVG assets."
+      "Fixes crashes related to direct selection on specific SVG assets.",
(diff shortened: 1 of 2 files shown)


stristr:
LGTM, merging.

PR #878, opened 2018-10-23
fix: do not show double notices in Creator

stristr:
OK to merge.

Short review.

Summary of changes:

- Fix double notices in Creator. Holdover from recent refactoring. To repro issue before fix, cause a notice (e.g. with project open, instantiate a layer with unshapeified text content inside), then back out to project browser.
- Tear down notices when going from project -> project browser.

Regressions to look for:

- None.

Completed checkin tasks:


roperzh (approved):
LGTM

PR #879, opened 2018-10-24
fix: force actions editor to close in preview mode

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes crash [TypeError: Cannot read property 'contains' of null](https://app.asana.com/0/856556209422928/880556083108445).

Since we want to close the actions editor no matter what when switching
to preview mode, use `#forceHideEventHandlersEditor` instead of
`#safelyHideEventHandlersEditor`

Regressions to look for:

- None

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -6,7 +6,8 @@
     "Fixes": [
       "Fixes crashes related to direct selection on specific SVG assets.",
       "Fixes crashes in certain circumstances when creating states and undoing.",
-      "Fixes crashes when selecting autocompletion items with the mouse."
+      "Fixes crashes when selecting autocompletion items with the mouse.",
+      "Fixes crashes when switching to Preview Mode with the Actions Editor open."
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


stristr (commented):
Just an observation: I believe work _can be lost_ with this mechanism and I could imagine users getting tripped up by this (everything else auto-saves _except for_ code editing).

Could we at least _try_ to save first? I realize it's much harder to interrupt switching to preview mode if we can't save (because of IPC considerations), but it looks like `doSave()` is guarded, so we could try:

```
forceHideEventHandlersEditor (trySave = false) {
  if (trySave && this.editor && !this.editor.canBeClosedExternally()) {
    this.editor.doSave();
  }

  ...
}
```

roperzh:
Thanks for the review and for the great catch. I modified your example sightly to use "named parameters" just because of personal preference. We may want to define a general styleguide for mono2, what do you think?

stristr:
Certainly. I think with TypeScript, named parameters makes lots of sense. With JS, it's kind of icky. (This is fine, no need to debate it.)

stristr:
LGTM.

PR #882, opened 2018-10-25
fix: do not crash on invalid 'd' attributes

roperzh:
OK to merge.

Short review.

Summary of changes:

Avoid crashing on invalid 'd' attributes, preventing projects from becoming broken if the user enters an invalid value from the timeline.

This will, however show a toast (which I intentionally left since I think it's useful), but can be removed if we think it's noisy.

image

I also checked for polygon points, and the behavior is the same, ie: a toast

Regressions to look for:

- None

Completed checkin tasks:


stristr (changes requested):
Great starting point! I think there is still some work needed in the general realm of direct selection. Got a freeze from these steps:

1. New project, instantiate `blue-circle`.
2. Direct select, then convert to path.
3. Change path.d to `Mfoo,bar`. I get the toast, _but also_, the node does not disappear and I'm stuck in direct selection mode with no way to click out of it. I can't instantiate an element on stage, either—it seems like the entire app is broken at this point.

stristr:
I'm going to ticket the remainder of this and merge this as is for 4.3.2.

PR #888, opened 2018-10-29
tweak: ameliorate inadvertent panel resizings on ticker t0 drag attempts

taylorpoe:
I found myself accidentally resizing the timeline property panel quite often when attempting to scrub the ticker from a t0 position. I changed the size of the panel resizer to not overlap with the gauge to solve this misbehavior.


roperzh (approved):
LGTM, thanks for fixing this

PR #889, opened 2018-10-30
tweak: do not auto-curve when the value is unchanged.

stristr:
OK to merge if @roperzh and @taylorpoe agree.

Short review.

Summary of changes:

- Do not auto-curve when the value is unchanged ahead of/behind the curve in question. We will still auto curve if/when the value is changed. This fixes a workflow I frequently use when building an animation that I want to start at some frame > 0—e.g., creating a keyframe at t=0, then an identical keyframe at t=30 when I actually want the animation to start.

Regressions to look for:

- None expected—this seems to work nicely.

Completed checkin tasks:


roperzh:
LGTM

PR #891, opened 2018-10-30
fix: do not try to handle 'direct-selection-anchor' if there is not an Element.directlySelected

roperzh:
OK to merge.

Short review.

Summary of changes:

Very small change that fixes [TypeError: Cannot read property 'type' of null](https://app.asana.com/0/856556209422928/885913453927439), but I wanted to make a PR just in case I'm just sweeping a deeper problem (`Element.directlySelected` being `null`) under the rug.

However, we are doing very similar checks in other parts of Glass.js, so this seems ok.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging. I've been in the habit of sweeping things under the rug in similar fashion 🤠.

PR #896, opened 2018-11-01
fix: don't allow to create keyframes for size properties on root

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes [I can create keyframes for sizeX and sizeY on the root node in the timeline](https://app.asana.com/0/856556209422928/864034901419619), I tried to track down to where we broke this in order to revert that specific section of the code, but _seems like_ the app never behaved as expected: I tested all the way back to Haiku `v3.4.0` and the behavior is the same.

In order to fix this, I introduced a new kind of filter for properties in the `Property` model (please let me know if you think this belongs somewhere else): `keyframe` that determines if you can create new keyframes for a property or not.

The logic is flexible enough so we can add more specific cases in the future if we wish. The rest of the changes are related to special-casing the following paths:

- Keyframe creation
- Drag & Drop of keyframes
- Timeline context menu

Regressions to look for:

- None expected

Completed checkin tasks:


roperzh:
@stristr I removed some properties from `Property.DISPLAY_RULES` according to what we discussed in 310df2517655456c2e9217b397acec14f6e5ba45, and I think this is ready to be reviewed again 🙌

stristr:
Yes! This is a very satisfying PR, thank you. (@zackbrown, this fixes the "unexpected" issues you just mentioned to a user in Intercom, in which it was possible to accidentally create artboard size keyframes after t = 0.)

stristr:
LGTM, merging.

PR #904, opened 2018-11-20
Fixes 11 19

roperzh:
OK to merge.

Short review.

Summary of changes:

Fixes:

- [TypeError: Cannot read property 'value' of null](https://app.asana.com/0/856556209422928/913836623923855)
- [TypeError: e.match is not a function](https://app.asana.com/0/856556209422928/913835582290002) 
- [TypeError: Cannot read property 'getContextSize' of null](https://app.asana.com/0/856556209422928/898181831443936)

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/packages/haiku-glass/src/react/Glass.js
+++ b/packages/haiku-glass/src/react/Glass.js
@@ -583,17 +583,26 @@ export class Glass extends React.Component {
 
         case 'global-menu:zoom-in':
           mixpanel.haikuTrack('creator:glass:zoom-in');
-          this.getActiveComponent().getArtboard().zoomIn(1 + SHORTCUT_ZOOM_FACTOR);
+          const component = this.getActiveComponent()
+          if (component) {
+            component.getArtboard().zoomIn(1 + SHORTCUT_ZOOM_FACTOR);
+          }
           break;
 
         case 'global-menu:zoom-out':
           mixpanel.haikuTrack('creator:glass:zoom-out');
-          this.getActiveComponent().getArtboard().zoomOut(1 + SHORTCUT_ZOOM_FACTOR);
+          const component = this.getActiveComponent()
+          if (component) {
+            component.getArtboard().zoomOut(1 + SHORTCUT_ZOOM_FACTOR);
+          }
           break;
 
         case 'global-menu:reset-viewport':
           mixpanel.haikuTrack('creator:glass:reset-viewport');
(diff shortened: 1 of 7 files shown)


packages/haiku-glass/src/react/Glass.js
```diff
@@ -609,13 +609,15 @@ export class Glass extends React.Component {
...
+          const proxy = this.fetchProxyElementForSelection()
+          if (proxy && proxy.hasAnythingInSelectionButNotArtboard()) {
             this.handleCutDebounced();
           }
           break;
 
         case 'global-menu:copy':
-          if (this.fetchProxyElementForSelection().hasAnythingInSelectionButNotArtboard()) {
+          const proxy = this.fetchProxyElementForSelection()
+          if (proxy && proxy.hasAnythingInSelectionButNotArtboard()) {
```
stristr, line 629: Just to anticipate and correct any other crashes related to the issue fixed in this PR, we should check `Glass.js` for any and all unchecked calls to `this.fetchProxyElementForSelection()`. (For example, can you also trigger a new "trying to access property of `undefined`" crash with `Cmd+G` to group?) Any time it appears, we should check if the returned value is undefined and skip code that will crash if it is. From scanning through, there are a few cases where this _isn't_ as simple as early returning.

roperzh: understood! 👍

roperzh:
@stristr I added two related solutions in different commits so we can revert each of them individually if we need, we have:

- 64cac717be24edbe59a2c454e0d601bbe85237b8 which disables all the project-related top menu shortcuts by setting `isProjectOpen` to `false` until `awaitAllProjectModelsState` is called. This helps with the `cmd+enter` preview mode shortcut during init and seems overall like a good thing to do.
- ee75a6483953216b91a5f17614c1f3e529d4e373 which prevents crashes on zoom-in, zoom-out, zoom-reset, group, ungroup and paste during initialization.

stristr:
Both changes LGTM, but I think we _don't_ want the changes in `packages/haiku-glass/src/electron.js` and `packages/haiku-timeline/src/electron.js`—these are only used in standalone mode, and I think not having the default `true` there would break most keyboard shortcuts (e.g. can't zoom, group, etc. in standalone glass).

✅to merge after reverting only those changes from 64cac71.

PR #909, opened 2018-11-27
fix: check if moveTo is defined before trying to read from it

roperzh:
OK to merge.

Short review.

Summary of changes:

As the title says, `moveTo` can be undefined if we reach the `else` branch first, which can be the case for invalid paths.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #911, opened 2018-12-05
fix: edge case of long, allcaps proj names covering menu btn

taylorpoe:
OK to merge.

Fixed edge case where max-character-length, all-caps project names were covering the dashboard project card's options menu. A user encountered this and was not able to delete there project due to it. (Would have been able to if they resized their browser slightly.)


stristr:
Hehe, "edge" case.

LGTM.

stristr:
Health checks won't run because branch is based on a very old master. Merging, but will harass @taylorpoe on Slack.

PR #917, opened 2018-12-12
fix: calcule hashes differently on getInsertionPointInfo

roperzh:
OK to merge.

Short review.

Summary of changes:

@stristr this is a _very_ lazy way to solve the problem you described in [Multiple gradients clash when mergeDesigns happens](https://app.asana.com/0/922186784503552/941872991077444), is it good enough?

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -663,7 +663,7 @@ class ActiveComponent extends BaseModel {
 
     const index = (mana && mana.children && mana.children.length) || 0;
 
-    const template = mana && Template.manaWithOnlyMinimalProps(mana, () => ({}));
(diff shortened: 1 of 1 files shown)


roperzh:
Nevermind, I'll work on a better solution

stristr:
Yeah, let me clarify exactly what we want:
 - We want the hash from `getInsertionPointInfo` to be exactly as it is—this ensures stability during `mergeDesigns`.
 - We want the _randomizer_ used during `Template.fixFragmentIdentifierReferences` to be more strict while still being consistent across processes.

I was looking at the code and it seems like it's supposed to be doing this by passing an increasing `nonce` in `mergeMana`…so the regression might not actually be why I suspected. I'm going to walk through with a debugger quickly to see if I can't understand what's happening.

PR #922, opened 2018-12-20
fix: fixes crash when trying to use Object.keys on non-objects

roperzh:
OK to merge.

Short review.

Summary of changes:

In ES5, if the argument to this method is not an object (a primitive),
then it will cause a TypeError.

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/helpers/isMutableProperty.ts
+++ b/packages/@haiku/core/src/helpers/isMutableProperty.ts
@@ -6,15 +6,21 @@
  * @method isMutableProperty
  * @description Mechanism to determine if a property is mutated within the timeline.
  */
-export default function isMutableProperty (property, propertyName) {
+export default function isMutableProperty (property, propertyName: string) {
   // For now, we can use a "naive" set of sufficient (but not necessary) conditions for immutability of a property:
   //   1. Has exactly one keyframe.
   //   2. The keyframe is "0".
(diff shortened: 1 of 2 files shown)


stristr:
This fix shouldn't be necessary and is along a sufficiently hot path that I'm concerned about perf regressions…do we have an associated crash that should explain why what we're currently doing is not sufficient? `Migration.ts` is supposed to run before the method where this change was made: see [here](https://github.com/HaikuTeam/mono/blob/master/packages/%40haiku/core/src/Migration.ts#L152-L159).

stristr:
(Note: this is also broken in ES2017 :smile:)

roperzh:
@stristr I'm very sorry for not including the link to the task, I completely forgot! 

> do we have an associated crash that should explain why what we're currently doing is not sufficient?

We do have a crash: ["TypeError: Cannot convert undefined or null to object"](https://sentry.io/haiku-systems/nodejs-plumbing/issues/813802825/?referrer=asana_plugin), for some reason `undefined` or `null` made its way into `isMutableProperty`

First theory after a quick look is that in `Migration.ts` we are specifically checking for `null` but not for `undefined` 🤔. I'll dig further and ping you back with a better solution. Thanks!

stristr:
Yeah, that should not be possible! Note how we turn everything except null into an object.

roperzh:
Closing in favor of investigating the root problem causing this.

ref: https://app.asana.com/0/922186784503552/854040000213905

PR #923, opened 2018-12-20
fix: don't crash on null states

roperzh:
OK to merge.

Short review.

Summary of changes:

The title says it all, one small detail could be to set the type to `'null'` instead of `'object'`, let me know if you want to make that change and I'll amend shortly.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
Good enough!

PR #929, opened 2019-01-03
fix: actually solve the issue with ID re-refs during padding

stristr:
OK to merge.

Short review.

Summary of changes:

- [Fixes More ID shuffle madness can result in disappearing elements](https://app.asana.com/0/922186784503552/960107844835150/f). The prior fix in 497ec0fb00aa8248f25f9da397d8dd37f2415d65, reverted to solve gradient problems, didn't work because the remapped ID was re-read for the purpose of reassigning the ID…anyway, this should actually be good now.


stristr:
Tested and confirmed working for all problem areas. Merging.

PR #932, opened 2019-01-03
perf: do not enqueue ticks for serialization work

stristr:
OK to merge.

Short review.

Summary of changes:

- Other than adding some missing types, the two things happening are [here](https://github.com/HaikuTeam/mono/compare/cold-perf?expand=1#diff-47ce953ca75eceeeb727f4beece5ac77R91) and [here](https://github.com/HaikuTeam/mono/compare/cold-perf?expand=1#diff-7433d312b35db871f4e4161899095cfeR2535). Point being to not enqueue ticks for components that are effectively running headlessly. This should give us a nice perf gain in all processes where the serialization layer is running without affecting preview mode (since that mounts a new component each time). Although this change seems minor and safe, it's worth confirming nothing is broken in Glass and Timeline.

### DIFF
--- a/packages/@haiku/core/src/HaikuClock.ts
+++ b/packages/@haiku/core/src/HaikuClock.ts
@@ -2,6 +2,7 @@
  * Copyright (c) Haiku 2016-2018. All rights reserved.
  */
 
+import {ClockConfig, IHaikuClock} from './api';
 import HaikuBase from './HaikuBase';
 import HaikuGlobal from './HaikuGlobal';
 import assign from './vendor/assign';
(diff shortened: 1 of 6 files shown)


stristr:
Flagged WIP…needs some work on Glass where on-stage transforms are no longer working great due to dependence on manual `tick()` calls during reload.

stristr:
Unflagged WIP—should be ready for review.

roperzh (approved):
LGTM and works very nice! I tried to break it in every way I can imagine without success 💯

PR #933, opened 2019-01-04
fix: disable TopMenu items that require auth on login screen

roperzh:
OK to merge.

Short review.

Summary of changes:

This issue was identified as part of another familiy of issues, all
related to the tour and the behavior of the app for logged-out users.

As the time of this commit there where three top menu items that require
the user to be logged in in order to work:

- New Project
- Take Tour
- What's New

We introduced a new TopMenu option, `isUserAuthenticated`, which is set
during log in/out and projectList fetch (to aid for
already logged in users).

Asana issue: https://app.asana.com/0/922186784503552/963025133104712

Regressions to look for:

- None expected


stristr:
Tested all scenarios I could think of. LGTM, merging.

PR #935, opened 2019-01-07
fix: set Keyframes as TransitionSegments iif they have a supported curve

roperzh:
OK to merge.

Short review.

Summary of changes:

Since at the moment we don't support custom curves, our view logic
assumes that a keyframe with a curve value defined [is a supported
curve value][1], causing crashes if the user sets a custom curve in the
bytecode by hand.

We have at least two possible options to aid this:

1- Do a check in the view before trying to render the curve and render a default
if we don't recognize the curve.
2- Don't consider a Keyframe a transition segment if we don't recognize
the curve.

This changeset goes with `2` because it aligns better with what is actually
happening in core: we don't animate between keyframes with unrecognized curves.

Going with `2` also protect us against front-end logic that acts on transition
segments based on [events emitted from TransitionSegment components][2].

[1]: /HaikuTeam/mono/tree/e1e1bef96293f3554bd61a334a972732852fdfce/packages/@haiku/core/src/api/index.ts#L591
[2]: /HaikuTeam/mono/tree/e1e1bef96293f3554bd61a334a972732852fdfce/packages/haiku-timeline/src/components/TransitionBody.js#L196

Regressions to look for:

- Could perf be affected significantly by this?

### DIFF
--- a/packages/haiku-serialization/src/bll/Keyframe.js
+++ b/packages/haiku-serialization/src/bll/Keyframe.js
@@ -1,6 +1,7 @@
 const HaikuComponent = require('@haiku/core/lib/HaikuComponent').default;
 const expressionToRO = require('@haiku/core/lib/reflection/expressionToRO').default;
 const BaseModel = require('./BaseModel');
+const SupportedCurves = require('@haiku/core/lib/api/index').Curve;
 
 /**
  * @class Keyframe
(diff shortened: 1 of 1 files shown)


roperzh:
Please hold on this until we define if we want to support custom curves by manually editing the bytecode as of today.

ref: https://app.asana.com/0/922186784503552/966279543181585

roperzh:
Closing in favor of allowing custom curves.

PR #938, opened 2019-01-08
fix: cmd + shift + left now also changes time internally

roperzh:
OK to merge.

Short review.

Summary of changes:

The logic that moves the time when the shortcut is invoked was using
`Timeline.updateCurrentFrame` to set the frame to `0`, but this method only
updates the internal state of the `Timeline` model but not the `Envoy` channel.

Calling `Timeline.seek` does that and is enough to also update the the view.

Asana task: https://app.asana.com/0/922186784503552/967841045402195

Regressions to look for:

- None expected


roperzh:
RECHECK!

stristr:
LGTM, merging.

PR #939, opened 2019-01-08
fix: do not auto-tween on keyframe moves

roperzh:
OK to merge.

Short review.

Summary of changes:

The reason for this logic existing on "keyframe move" was because in
the first implementation we decided to add curves even between keyframes with the
same value, so this aided the special-case of creating a tween when the keyframe
at t=0 was moved.

Since dcaddc7843adffd611e61f7c9c83d3b101ba011d, we are no longer doing that, thus is safe to delete auto-tweening
from the keyframe move logic, which also aids a bug where a tween is created
every time a keyframe next to t=0 is moved.

Asana task: https://app.asana.com/0/922186784503552/967841045402196

Regressions to look for:

@taylorpoe @stristr can you think of a possible scenario that not autotweening on keyframe moves can cause?

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -3731,21 +3731,6 @@ class ActiveComponent extends BaseModel {
               propertyName,
               true, // fallbackToInitialKeyframeIfProvided
             );
-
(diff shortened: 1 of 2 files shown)


stristr:
RECHECK!

stristr:
LGTM, merging.

PR #940, opened 2019-01-08
fix: prevent empty event handlers from randomly appearing in bytecode

roperzh:
OK to merge.

Short review.

Summary of changes:

We have noticed empty event handlers written in the bytecode in the form of:

```js
eventHandlers: {
  "haiku:Main-03757d2ca1026e0a": {}
}
```

There are two "forms" of this bug: one transient and one permanent. Reproing the
transient form is easy: just switch between code/edit mode and watch the
`eventHandlers` key in the in-app editor. I call this the transient
version because the bytecode shown here is the in-memory bytecode of Creator, if you
make an edit to the file and save, you now have permanently saved the empty
event handler on disk.

This is happening because `getReifiedEventHandlers`, which is called many times
with the sole purpose of listing event handlers in objects, fetches and
modifies a reference of the in-memory bytecode, adding the empty object.

Given how our logic works, we still need to modify this reference in some
scenarios, but in order to solve this I introduced a new method,
`getReifiedUnreferencedEventHandlers`, which doesn't modify the in-memory
bytecode and can be used with the sole purpose of reading event handlers.

Asana task: https://app.asana.com/0/922186784503552/968639637587944

Regressions to look for:

- None expected, but watch out for regressions in Actions

### DIFF
--- a/packages/haiku-creator/src/react/Creator.js
+++ b/packages/haiku-creator/src/react/Creator.js
@@ -1918,7 +1918,7 @@ export default class Creator extends React.Component {
     );
   }
 
-  saveEventHandlers (targetElement, serializedEvents) {
+  saveEventHandlers = (targetElement, serializedEvents) => {
     const selectorName = 'haiku:' + targetElement.getComponentId();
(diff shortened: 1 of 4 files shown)


roperzh:
closing in favor of #941

PR #948, opened 2019-01-22
feat: curve editor & custom Bezier Curves

roperzh:
OK to merge.

Short review.

Summary of changes:

- Adds the ability to declare custom bezier curves in the bytecode via an array
  that specifies the control points of a Cubic Bezier Curve.
- Adds a custom bezier editor to the UI that can be enabled via right click >
  'Edit Bezier Curve', and it will be enabled iif:
  - All of the selected curves are the same curve
  - All of the selected curves can be decomposed into a single Bezier Curve.

Regressions to look for:

- None expected


packages/haiku-ui-common/src/react/Bezier/BezierDerivativeGraph.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+
+export interface BezierDerivativeGraphProps {
+  rightGradFill: string;
+  leftGradFill: string;
+  id: string;
+  value: number[];
+}
+
+export default class BezierDerivativeGraph extends React.PureComponent<BezierDerivativeGraphProps> {
```
roperzh, line 14: @stristr I'm pretty sure that this component doesn't behave as expected 100% of the time, it may be good to go and refine later, but I just wanted to point that up.

My understanding is that the graphs in the timeline are the rate of change => the derivative. Following [Wikipedia](https://en.wikipedia.org/wiki/B%C3%A9zier_curve#Cubic_B%C3%A9zier_curves) advise, I am able to get the derivative of the explicit form of the curve (which is a Quadratic Bezier Curve) and then I plot this new curve at regular intervals, but I still get weird results for some curves.

stristr:
Does this need to be rebased?

packages/haiku-serialization/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test": "yarn test:all",
-    "test:all": "cross-env NODE_ENV=test tape \"test/**/**/*.test.js\" | tap-spec || true",
+    "test:all": "cross-env NODE_ENV=test tape \"/Users/roperzh/projects/haiku-projects/mono/packages/haiku-serialization/test/bll/05_Keyframe.test.js\" | tap-spec || true",
```
stristr: (should be changed back)

roperzh: apologies 🙏

roperzh:
> Does this need to be rebased?

Indeed, rebased and pushed, sorry about that!

roperzh:
cc: @taylorpoe just in case you want to do another design pass :)

packages/haiku-ui-common/src/react/Bezier/BezierDerivativeGraph.tsx
stristr: That looks correct…but not every bezier curve is going to have an inflection point at x = `0.5` though, right? I would assume that's why it's weird (the bug is in `get d ()`, not here). What's an example curve that looks weird to you?

packages/haiku-ui-common/src/react/Bezier/BezierDerivativeGraph.tsx
```diff
@@ -0,0 +1,44 @@
...
+  private scale = 10;
+
+  derivateX (value: number[]): number[] {
+    const p = [0, value[0], value[2], 1];
+
+    return [
+      3 * (p[1] - p[0]),
+      3 * (p[2] - p[1]),
+      3 * (p[3] - p[2]),
+    ].map((v) => Math.abs(v * this.scale));
```
stristr: Also occurs to me that multiplying by scale doesn't really make sense…this doesn't normalize control points in the box (`[0, 0]` -> `[100, 25]`) which it seems is what we want.

roperzh: I pushed some changes to this logic in 7d13c5a95ae8491d9805ed45d98d22f393abf2c0, what do you think? :)

stristr:
@roperzh, compile error is: `src/components/BezierPopup.tsx(1,43): error TS2307: Cannot find module 'haiku-formats/lib/exporters/curves'.

Guessing we added a dependency `haiku-timeline`->`haiku-formats` but didn't declare it.

roperzh:
Thanks to @stristr for figuring out all the math and providing a [beautiful
codepen][1] with literally everything ready to copy/paste.

[1]: https://codepen.io/anon/pen/exGgeX

stristr:
LGTM.

roperzh:
Math is hard. Thanks Sasha!

PR #950, opened 2019-01-25
fix: avoid tweens from shrinking on certain drag motions

roperzh:
OK to merge.

Short review.

Summary of changes:

Due to how the drag logic is setup, the `onDrag` method can be called after the
`onStop` method on `TransitionBody`, causing the drag calculation to be
incorrect and shrink the components involved.

This adds a flag to prevent `onDrag` to be called after `onStop` is called.

Note that there's similar logic for other components like
`InvisibleKeyframeDragger` and `Scrubber`, but they shouldn't be affected by
this.

Asana Task: https://app.asana.com/0/922186784503552/1101796002992131

Regressions to look for:

- None expected, but just in case: drag and drop of tweens.

### DIFF
--- a/packages/haiku-serialization/src/bll/ActiveComponent.js
+++ b/packages/haiku-serialization/src/bll/ActiveComponent.js
@@ -2077,9 +2077,9 @@ class ActiveComponent extends BaseModel {
     }
   }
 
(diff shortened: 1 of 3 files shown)


stristr:
LGTM, merging.

PR #949, opened 2019-01-25
fix: prevent tween poles from overflowing of their row

roperzh:
OK to merge.

Short review.

Summary of changes:

Due to some CSS being misplaced, the keyframe poles of tweens were overflowing
from their rows, causing the wrong tween to be selected on certain
circumstances.

Asana Task: https://app.asana.com/0/922186784503552/1102067997618469

Regressions to look for:

- None expected

PR #951, opened 2019-01-28
fix: avoid marquee init from starting on TransitionBody keyframe poles

roperzh:
OK to merge.

Short review.

Summary of changes:

This will prevent issues with marquee selection causing a weird behavior where
marquee is active and some tweens are dragged at the same time.

This is also related to #949 because the overflow of the poles makes this issue
more feasible to happen.

Asana Task: https://app.asana.com/0/922186784503552/814729684196780

Regressions to look for:

- None expected


stristr:
LGTM, merging.

PR #952, opened 2019-01-28
fix: prevent crashes when the plus button ref is missing

roperzh:
OK to merge.

Short review.

Summary of changes:

According to the [React docs][1], it's completely normal and expected to get
`null` refs as the component is mounting/unmounting. Since the scroll event of
the Actions Editor can be triggered while the ref is null, this adds a simple
guard to prevent crashes.

Asana Task: https://app.asana.com/0/922186784503552/1104227124130859

[1]: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

Regressions to look for:

None expected


stristr:
LGTM, merging.

PR #954, opened 2019-02-04
fix: early return instead of crashing animating paths w/ different lengths

roperzh:
OK to merge.

Short review.

Summary of changes:

Early return instead of crashing animating paths w/ different lengths

Regressions to look for:

- None

### DIFF
--- a/packages/@haiku/core/src/helpers/PathUtils.ts
+++ b/packages/@haiku/core/src/helpers/PathUtils.ts
@@ -206,7 +206,7 @@ export const distributeTotalVertices = (path: CurveSpec[], totalVertices: number
 
 export const rotatePathForSmallestDistance = (source: CurveSpec[], dest: CurveSpec[]): void => {
   if (source.length !== dest.length) {
-    throw new Error('Mismatched source and destination length.');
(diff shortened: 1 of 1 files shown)


stristr:
I have some concerns about this approach.
 - Could this cause runtime Lottie crashes or Lottie export crashes? I suspect it could because Lottie is very particular.
 - As far as I can tell, `rotatePathForSmallestDistance()` is only invoked from `synchronizePathStructure()`, _after_ an invocation of `distributeTotalVertices()`, whose purpose seems to be preventing this error condition from ever being hit. Given that, it seems like a runtime error is the expected behavior, and we should probably actually be fixing a bug in `distributeTotalVertices()`. Do we have any more info about what might have caused this?

roperzh:
That makes sense, I'm going to close this to come back with a better approach. Thank you!!

stristr:
Thanks @roperzh! Do you happen to have an example of a project with this crash? I'm guessing it's a somewhat straightforward fix if we actually can see what broke. Might even be able to detect/fix just by reading/challenging the method signature.

PR #958, opened 2019-02-07
fix: prevent rows from unexpectedly expanding on certain situations

roperzh:
OK to merge.

Short review.

Summary of changes:

We are setting `_isExpanded` true to all nested heading rows on every render (ugly hack that I'm responsible of)
because otherwise elements under `<g>`s are not displayed due to Timeline internal "politics" causing [weird issues][1].

With that in mind, dragging rows triggers a re-render => that sets `_isExpanded = true` => which in  turn expands nested rows.

This fixes the problem by expanding only `<g>` elements, fixing all sort of weirdness around timeline rows and their expanded state.

Asana task: https://app.asana.com/0/922186784503552/1108525011175274

Regressions to look for:

- None expected

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -4,7 +4,7 @@
       "Double click on curves to open the Bezier Editor."
     ],
     "Fixes": [
-      "FIRST!"
+      "Prevent nested Timeline rows from expanding unexpectedly."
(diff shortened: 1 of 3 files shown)


stristr:
LGTM, merging.

PR #961, opened 2019-02-09
fix: filter invalid paths before trying to synchronizePathStructure

roperzh:
OK to merge.

Short review.

Summary of changes:

Another possible solution for #954, this time the idea is to filter invalid paths before trying to do anything with them. 

If we think that this is a solution worth perusing I can add some tests.

Regressions to look for:

- None expected


stristr:
LGTM, merging.

PR #962, opened 2019-02-09
fix: upgrade node-gyp version to prevent install errors on macOS Mojave

roperzh:
OK to merge.

Short review.

Summary of changes:

Version [3.6.3][1] only includes an specific fix to a regex that was causing
the installation of node-gyp to fail on macOS Mojave, so it should be safe to
update without problems.

[1]:https://github.com/nodejs/node-gyp/commit/4c387070872d8cc7224524852f03df4cd3b90a7a

Regressions to look for:

- I'm not sure, this shouldn't cause any problems but, node node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "lodash": "^4.17.4",
     "mime-types": "^2.1.17",
     "moment": "2.18.1",
-    "node-gyp": "^3.6.2",
+    "node-gyp": "^3.6.3",
     "semver": "^5.4.1",
     "semver-sort": "^0.0.4",
     "slack": "^8.3.1",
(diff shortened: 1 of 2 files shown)


stristr:
Would you please test that this builds on CI before we merge it by making a throwaway RC?

(Any PR from a branch beginning with "rc-" will give you an option to advance the build.)

Else LGTM.

roperzh:
yay! thanks for the review and the instructions. I created #963, which was [built successfully](https://ci.haiku.ai/blue/organizations/jenkins/Haiku/detail/Haiku/1488/pipeline), I also tested the [dmg file](https://s3.amazonaws.com/haiku-electron-releases-production/releases/production/master/mac/8450234216094/4.3.12/Haiku-4.3.12.dmg) and everything seems to be working fine.

Merging!

PR #965, opened 2019-02-11
hackfix: assign better boilerplate parameters to custom events

roperzh:
OK to merge.

Summary of changes:

A hackfix with several disadvantages:

- By looking at the code, seems like the user could send more than one data
parameter via custom events, however this covers the most common use case, which
is a single parameter.
- The way to detect if an event is a custom event is _ugly_, and will not update
the parameters of old custom events, which is good in a sense, since the user can
edit the parameters directly in the bytecode.

While this is not the ideal way to go, nor 100, it's better than showing 'element'
as the second parameter for all custom events.

Asana task: https://app.asana.com/0/922186784503552/1109117416798204

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -6,7 +6,8 @@
     ],
     "Fixes": [
       "Prevent Timeline rows from expanding unexpectedly.",
-      "Fix empty `lottie.json` files for certain animations made with path morphing."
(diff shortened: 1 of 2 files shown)


roperzh:
update: rebased and force pushed to resolve conflicts.

stristr:
LGTM.

PR #968, opened 2019-02-13
fix: check if we get a valid color string before trying to infer its model

roperzh:
OK to merge.

Summary of changes:

- This will prevent crashes when an invalid color string is entered in the input
or coded in the bytecode

Asana Task: https://app.asana.com/0/922186784503552/1109531996595639

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #973, opened 2019-02-19
chore: Print env.HAIKU_RELEASE_COUNTDOWN on Slack advancement messages

roperzh:
Summary of changes:

- Print `env.HAIKU_RELEASE_COUNTDOWN` on Slack advancement messages

Regressions to look for:

- None

Completed checkin tasks:


stristr:
LGTM, merging.

For future: it would be nice to expose this as an input in the final advancement step, defaulted to what Jenkins thinks it should be (similar to what we do with setting the RC version), and allow the user to change it to what it should be at this stage.

PR #975, opened 2019-02-20
fix: properly set internal state variables on `Timeline.seekAndPause`

roperzh:
Summary of changes:

Before this fix, `Timeline.seekAndPause` was only setting `currentFrame` and
`_isPlaying`, but turns out that `Timeline.seek` and `Timeline.pause` do
much more than that, for example, setting `_lastSeek` to `null`.

This fixes the issue by directly calling these methods instead of trying
to replicate what they are doing, but taking care of providing a parameter
to avoid notifying `Envoy` on every call.

Asana Task: https://app.asana.com/0/922186784503552/1110506015963658

Regressions to look for:

- None expected, but careful review is adviced.

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Timeline.js
+++ b/packages/haiku-serialization/src/bll/Timeline.js
@@ -169,7 +169,7 @@ class Timeline extends BaseModel {
     }
   }
 
-  pause (skipTransmit) {
+  pause (skipTransmit = false) {
     this._playing = false;
     this._lastSeek = null;
     if (!skipTransmit && !this.component.project.getEnvoyClient().isInMockMode()) {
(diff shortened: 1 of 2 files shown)


packages/haiku-serialization/src/bll/Timeline.js
```diff
@@ -169,10 +169,10 @@ class Timeline extends BaseModel {
     }
   }
 
-  pause () {
+  pause (skipTransmit) {
```
stristr: For maybe-future typing plans: should be `skipTransmit = false` since in most places it's called without a parameter. (If TS, would still prefer this to `skipTransmit?: boolean`.)

stristr:
LGTM (note changelog conflict).

stristr (approved):

packages/haiku-serialization/src/bll/Timeline.js
roperzh: good call! updated :)

PR #977, opened 2019-02-20
 fix: prevent the expression input from closing if the mose event orginated from itself

roperzh:
Summary of changes:

fix: prevent the expression input from closing if the mose event orginated from itself

Regressions to look for:

- None expected, but would be a good idea to double test timeline property inputs.

Completed checkin tasks:


stristr:
LGTM.

PR #978, opened 2019-02-20
feat: add basic schema declarations for injectables

roperzh:
Summary of changes:

- Add basic schema declarations for injectables, this will enable more rich autocompletion options on expressions:

- Fix? `$clock` injectable (see comment, I'm not sure if this was intentional).

Known issues:

Due to how the `ExpressionInput` logic works, parse errors are triggered when you have an incomplete expression, which is the case if the expression ends with a dot (example: `$user.mouse.`). When a parse error is triggered, all autocompletion entries are deleted and fixing this would imply some refactor. Do we want to?

Regressions to look for:

- `@haiku/core` weight, maybe this could live somewhere else?

Completed checkin tasks:

### DIFF
--- a/packages/@haiku/core/src/HaikuComponent.ts
+++ b/packages/@haiku/core/src/HaikuComponent.ts
@@ -17,7 +17,7 @@ import {
   ParsedValueClusterCollection,
   TwoPointFiveDimensionalLayoutProperty,
 } from './api';
-import Config from './Config';
+import Config, {DEFAULTS} from './Config';
 import HaikuBase, {GLOBAL_LISTENER_KEY} from './HaikuBase';
 import HaikuElement from './HaikuElement';
 import HaikuHelpers from './HaikuHelpers';
@@ -3619,11 +3619,38 @@ export const ATTRS_CAMEL_TO_HYPH = {
   xHeight: 'x-height',
 };
 
+const PositionSchema = {
+  x: 'number',
+  y: 'number',
+  z: 'number',
+};
+
 export const ATTRS_HYPH_TO_CAMEL = {};
 
 for (const camel in ATTRS_CAMEL_TO_HYPH) {
   ATTRS_HYPH_TO_CAMEL[ATTRS_CAMEL_TO_HYPH[camel]] = camel;
 }
+
+INJECTABLES.$element = {
+  schema: {
+    offset: PositionSchema,
+    attributes: {},
+    tagName: 'string',
+    matrix: [],
+    opacity: 'number',
+    origin: PositionSchema,
+    rotation: PositionSchema,
+    scale: PositionSchema,
+    shown: 'boolean',
+    size: {x: 'number', y: 'number'},
+    sizeAbsolute: PositionSchema,
+    sizeProportional: PositionSchema,
+    translation: PositionSchema,
+  },
+  summon (injectees, component: HaikuComponent, node) {
+    injectees.$element = HaikuElement.findOrCreateByNode(node);
+  },
+};
 
 INJECTABLES.$window = {
   schema: {},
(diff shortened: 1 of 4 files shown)


packages/@haiku/core/src/HaikuComponent.ts
```diff
@@ -3639,8 +3666,36 @@ INJECTABLES.$mount = {
...
+    getFrameDuration: 'function',
+    getTime: 'function',
+    setTime: 'function',
+    isRunning: 'function',
+    run: 'function',
+    start: 'function',
+    assignOptions: 'function',
+  },
+  summon (injectees, component: HaikuComponent) {
+    injectees.$clock = component.getClock();
```
roperzh, line 3689: It is not clear from the diff because I moved this block of code, but this [used to be](https://github.com/HaikuTeam/mono/pull/978/files#diff-9a87d84e8130e8b1ead934d5d9c902d9L3686):

```js
injectees.$timeline = component.getClock();
```

stristr:
LGTM (fixed conflicts).

stristr:
@roperzh, it's unclear to me if the autocomplete clear-on-unexpected-token is a regression or an existing issue. Can you clarify?

roperzh:
@stristr unfortunately it's an existing issue, and I couldn't find a fix that doesn't involve some refactor :(

stristr:
In that case, let's merge this and backlog the existing issue.

roperzh:
sounds good, my only concern is that the bug is more evident now that we have nested autocomplete :(

stristr:
Ah, I see. Perhaps it's worth taking a look before merging…I will add to my list.

stristr:
@roperzh 

What about this marginal improvement?

Inside `ExpressionInput.js` > `ExpressionInput` > `rawValueToOfficialValue` > single-line expression mode, do something like this:

```
        if (clean[clean.length - 1] === '.') {
          clean = clean.substr(0, clean.length - 1);
        }
```

Idea being that we evaluate an expression that ends with `.` as if it weren't there yet. There is not a correct JS statement that terminates with `.` as far as I know, so this should behave as expected:
 - The autocomplete should show the right options.
 - It won't result in a parse error.
 - If you try to commit it, it will be as if the `.` just isn't there. This is arguably the only problem with it…but it's not _really_ a problem.

stristr:
@roperzh, I sent up three commits as we discussed—one to fix up the behavior of `Enter` with an autocomplete item selected, and one to repress those spurious `.` errors.

This now needs _your_ review, but I hope you find the changes to work well. I did some LIGHT refactoring of the autocomplete system:
 - Moved autocomplete behavior out of the view and into the `parseExpression` module.
 - Modified the rule that throws out completions identical to the chain we're trying to autocomplete so it only fires if this is the _only_ completion we found.
 - Made a better strict weak ordering than we had before, which should ensure predictable enter-to-commit behavior for the edge cases noted in the Trello ticket you dug up. Rules are:
    - Exact matches come first.
    - Case-insensitive prefix matches come next.
    - Remainder comes lexographically sorted.

packages/haiku-serialization/src/ast/parseExpression.js
```diff
@@ -163,16 +162,55 @@ function populateCompletions (target, injectables, keywords, declarations) {
...
+    }
+
+    // Ensures typeahead matches come next after exact matches.
+    if (na.startsWith(nChain)) {
+      return -1;
+    }
+
+    if (nb.startsWith(nChain)) {
+      return 1;
+    }
```
stristr, line 200: Nit: this is technically _not_ a strict weak ordering if `na` and `nb` both start with `nChain`. But JS won't explode when you do this and I really don't think it matters.

roperzh:
@stristr you rock! this works way better than it did before, I'll create a task to also show completions when a trailing dot is present, for example in this case:

image

I'd expect suggestions, like this:

image

Merging.

stristr:
Weird, that was working for me! How'd you get there?

roperzh:
LOL, seems to be a matter of if there's a whitespace between `$` and `=`, so for me `=$user.` will show suggestions but `= $user.` will not.

PR #979, opened 2019-02-20
fix: avoid glass marquee controls from being affected by zoom

roperzh:
Summary of changes:

- Avoid glass marquee controls from being affected by zoom, h/t 🎩 to @stristr for pointing out the solution in the team meeting.

Regressions to look for:

- None

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -7,7 +7,8 @@
       "Added a notice when trying to edit Bounce/Elastic curves via double-click.",
       "Prevented property inputs on the Timeline from closing unexpectedly.",
       "Fixes a crash when trying to animate invalid paths.",
-      "Fixes issues with the Timeline playback controls."
+      "Fixes issues with the Timeline playback controls.",
+      "Prevented marquee controls on stage from being affected by zoom."
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


stristr:
LGTM! Tiny detail: in terms of Haiku-authored bytecode, I _think_ `{'vector-effect': ...}` is old syntax, and `{vectorEffect: ...}` _should_ work. (Anything in `HaikuComponent.ts:ATTRS_CAMEL_TO_HYPH` should work as camel case.)

And since I think all attributes pass through `const final = ATTRS_CAMEL_TO_HYPH[key] || key;`, we might as well shorten the lookup where we can.

stristr:
(That was not a suggestion that we refactor all similar code, which would be a length endeavor—just new code 😸!)

roperzh:
Thanks! for some reason doesn't work :( I also tried camelCasing other attributes present on the same object (`stroke-width`, `fill-opacity`) just to try and they also don't work.

Could it be because we are [dealing directly with `HaikuDOMRenderer`](https://github.com/HaikuTeam/mono/blob/task-1109269464313082/packages/haiku-glass/src/react/Glass.js#L2965)?

stristr:
Probably. Ignore me! Merge away!

roperzh:
Thank you! 🤗

PR #981, opened 2019-02-27
UI for Animator launch

taylorpoe:
Summary of changes:
- Ports 'free tier' to 'free trial'
- Adds countdown UI to in-editor and dashboard screens
- Adds lockout modal with upgrade btn if expired
- Adds header to dashboard prompting upgrade if expired
- Adds [VIEW ONLINE] option to proj thumbnail for trial/pro/expired-trial users
- Updates logos throughout application to animator
- Rips out HaiLoader in favor of Animator + css-only loader
- Rips out iPreview (couldn't get this working in Haiku-proper when attempting to change colors)
- Updates color theme with new branding

Regressions to look for:
- broken styles
- incorrect expired/not-expired/pro behavior

Completed checkin tasks:
- [ ] Updated `changelog/public/latest.json`.


packages/@haiku/sdk-inkstone/src/index.ts
```diff
@@ -325,6 +325,14 @@ export namespace inkstone {
     export interface Organization {
       Name: string;
       PlanExpirationDate?: number;
+      Role: Role;
+      TrialStartDate?: number;
+      TrialDurationDays?: number;
+    }
+
+    export enum Role {
```
roperzh, line 333: https://github.com/HaikuTeam/mono/blob/60bf4679765206c63fa2f9462a060d58cecbed75/packages/%40haiku/sdk-inkstone/src/index.ts#L333-L336

nit: use `const enum` for better runtime perf ([check out how it compiles](https://www.typescriptlang.org/play/#src=enum%20EnumOnly%20%7B%0D%0A%20%20%20%20Free%20%3D%200%2C%0D%0A%20%20%20%20Pro%20%3D%202%2C%0D%0A%7D%0D%0A%0D%0Aconsole.log(EnumOnly.Free)%0D%0A%0D%0A%0D%0Aconst%20enum%20ConstEnum%20%7B%0D%0A%20%20%20%20Free%2C%0D%0A%20%20%20%20Pro%0D%0A%7D%0D%0A%0D%0Aconsole.log(ConstEnum.Free))) and since TS enums are number based, you don't have to specify the numbers! 

```suggestion
    export enum Role {
      Free,
      Pro,
    }
```

packages/haiku-creator/public/stylesheets/global.css
```diff
@@ -253,6 +253,29 @@ input::-webkit-input-placeholder {
...
+    border-radius: 10px;
+    background-image: linear-gradient(-45deg, #04B6CB 25%, rgba(255, 154, 26, 0) 25%, rgba(255, 154, 26, 0) 50%, #04B6CB 50%, #04B6CB 75%, rgba(255, 154, 26, 0) 75%);
+    background-color: #0098AA;
+    background-size: 50px 50px;
+    border: 1px solid #04B6CB;
+    border-bottom-color: #0098AA;
+    animation: move 2s linear infinite;
+}
+
+@-webkit-keyframes move {
```
roperzh: nit: avoid vendor prefixes when possible.

```suggestion
@keyframes move {
```

packages/haiku-creator/src/react/Creator.js
```diff
@@ -2116,10 +2110,14 @@ export default class Creator extends React.Component {
           envoyClient={this.envoyClient}
           {...this.props} />
         }
-        <ProjectLoader show={this.shouldShowProjectLoader()}>
+        <ProjectLoader show={true} show={this.shouldShowProjectLoader()}>
```
roperzh: heads up: duplicated prop, I think the only the second one will be applied.

packages/haiku-creator/src/react/components/LockoutModal.js
```diff
@@ -0,0 +1,120 @@
...
+  },
+  strong : {
+    fontWeight: 'bold',
+  },
+};
+
+class LockoutModal extends React.PureComponent {
+  constructor (props) {
+    super();
+  }
```
roperzh: nit: If you don’t initialize state and you don’t bind methods, you don’t need to implement a constructor for your React component.

packages/haiku-creator/src/react/components/ProjectThumbnail.js
```diff
@@ -110,7 +131,10 @@ class ProjectThumbnail extends React.Component {
           </span>}
           {this.props.allowDelete && <span
             key="delete"
-            onClick={this.props.showDeleteModal}
+            onClick={(e) => {
```
roperzh: nit: please use a bound function instead of creating an arrow function on every render when possible.

roperzh (commented):
So far LGTM, if possible I'd like to get another review from @stristr before merging.

### "Haiku" in the app

Apologies if we decided to not do this in this iteration! I think this list at least will be useful for future reference, so here we go:

- I pushed more changes related to the tour 🚌 in this [commit](14e4a0680e13a948f74b3ff93a76e7fc2518218c) mostly for reference so you can update if needed (I didn't know if we are going to use "Animator" or "Haiku Animator") but there is still some wording related to pro that must be changed in the tour, for example:
> Publishing also exposes multiple export formats, like GIF, Video, and Lottie. If you go Pro, you can also export these formats offline without publishing.
- The top menu is still worded with "Haiku" (About Haiku, Minimize Haiku, etc)
https://github.com/HaikuTeam/mono/blob/aeae6f45bd07e46f04ec31cde88cfb7fce5bc806/packages/haiku-common/src/electron/TopMenu.ts#L9
- The fail whale is still worded with "Haiku" https://github.com/HaikuTeam/mono/blob/aeae6f45bd07e46f04ec31cde88cfb7fce5bc806/packages/haiku-creator/src/react/components/Popups/FailWhale.tsx#L69
- Proxy help screen is still worded with "Haiku"
https://github.com/HaikuTeam/mono/blob/aeae6f45bd07e46f04ec31cde88cfb7fce5bc806/packages/haiku-creator/src/react/components/ProxyHelpScreen.js#L46
- Many prompts in the app are still worded with "Haiku" 
https://github.com/HaikuTeam/mono/blob/60bf4679765206c63fa2f9462a060d58cecbed75/packages/haiku-creator/src/react/components/StageTitleBar.js#L244
- Prompt to move to `/Applications` is still worded with "Haiku"
https://github.com/HaikuTeam/mono/blob/aeae6f45bd07e46f04ec31cde88cfb7fce5bc806/index.js#L26

### Comments
- Please feel free to ignore my nit comments, I put them just for reference.
- New UI looks amazing

packages/@haiku/sdk-inkstone/src/index.ts
stristr: On phone so haven't checked use case, but please note that if this is used outside sdk-inkstone, it can't be both const enum and usable.

roperzh: es, it is used outside, my bad! only improvement then could be to remove the explicit number declarations. Thanks stristr!

packages/haiku-creator/src/react/components/ProjectThumbnail.js
taylorpoe: @roperzh I changed this because I needed access to the `e`. Is there a non-arrow function way to pass an event to a method call?

packages/haiku-creator/src/react/Creator.js
taylorpoe: Tk, removing.

packages/haiku-creator/src/react/components/LockoutModal.js
taylorpoe: Tk, removed it.

taylorpoe:
I believe we should change all references to "Haiku" in the app to just  "Animator." Does that sound right, @zackbrown?

packages/haiku-creator/src/react/components/ProjectThumbnail.js
stristr: Yeah, so you can implement this within the class like so:

```
// In the class body
showDeleteModal = (e) => {
  e.stopPropagation();
  this.props.showDeleteModal();
};

// In the render() method
<Thing
 onClick={this.showDeleteModal}
/>
```

taylorpoe: Oh okay, cool. Thanks for the tut.

taylorpoe: Updated

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -265,15 +284,15 @@ class ProjectBrowser extends React.Component {
         <div style={{width: 560, fontSize: 16, lineHeight: 1.3, textAlign: 'center'}}>
           <div style={{height: 160}}><NoCon loop={true} sizing="cover" /></div>
           <div style={DASH_STYLES.notice}>
-            <div style={DASH_STYLES.noticeTitle}>Unable to connect to Haiku hosting services.<br/>Are you connected to the Internet?</div>
-            <div>Upgrade to Haiku Pro for offline capabilities.</div>
+            <div style={DASH_STYLES.noticeTitle}>Are you connected to the Internet?</div>
+            <div>The Animator free trial requires an active intenet connection.</div>
+            <div>Upgrade to Animator Pro for offline capabilities.</div>
```
stristr, line 289: On my computer: this looks like this:
image

Not intentional, right?

packages/haiku-creator/src/react/components/SideBar.js
```diff
@@ -106,13 +105,20 @@ class SideBar extends React.Component {
...
-            <div style={[STYLES.proBadge, this.state.isFullscreen && {left: 34}]}>PRO</div>
+        <div style={[STYLES.bar, {paddingLeft: this.state.isFullscreen ? 9 : 74}]} className="frame">
+          <div style={STYLES.logo}><AnimatorSVG /></div>
+          {trialDaysRemaining > 0 &&
+            <div
+              style={[STYLES.proBadge, this.state.isFullscreen && {left: 34}]}
+              aria-label={trialDaysRemaining + ` day${trialDaysRemaining === 1 ? '' : 's'} remain${trialDaysRemaining === 1 ? 's' : ''} in your free trial`}
+              data-tooltip={true}
+              data-tooltip-bottom={true}>
+              {trialDaysRemaining + ` day${trialDaysRemaining === 1 ? '' : 's'} remain`}
```
stristr: This will say `1 day remain` at one day. It's correct everywhere else.

packages/@haiku/sdk-inkstone/src/index.ts
zackbrown: Note that the association of the numbers to their symbols is very important here — this is "duplicated truth" from Inkstone (where an enum declares Free = 0 and Pro = 1).  If we remove the explicit number assignment here, we're one re-ordering (Pro, Free) away from this logic being very broken.  I'm in favor of explicit value assignment here as a safe-guard.

roperzh: that makes a lot of sense to me :)

packages/haiku-creator/src/react/Creator.js
```diff
@@ -694,9 +695,25 @@ export default class Creator extends React.Component {
...
 
       this.checkOnlineStatus();
+
+      this.user.isTrialExpired().then((isTrialExpired) => {
+        this.setState({
+          expiredTrialNonPro: isTrialExpired,
+        });
+      });
+
+      this.user.getOrganization().then((org) => {
```
stristr, line 711: Note: the entire identify is resolved from the `user:load` event, so you can skip an extra async Envoy call and use the `organization` object (instead of `org`) already present in this handler.

stristr:
There's an awkward edge here where `Project > New` (`Cmd + N`) still brings up the new project dialog, in addition to the lockout modal. `Esc` key closes the new project dialog which is underneath the lockout modal, so you can't see it but the opacity of the overlay does change. If instead of pressing `esc` you go to the dashboard, you apparently are able to create project: 
image

until you get hit with this: 
image

Simplest fix is probably to intercept using the expected state variables inside the `showNewProjectModal()` which is currently unguarded. Note that this area is rife with races…you'll probably want to bail early unless `this.state.isUserAuthenticated && this.envoyProject && this.envoyUser && !this.state.expiredTrialNonPro`.

zackbrown:
> I believe we should change all references to "Haiku" in the app to just "Animator." Does that sound right, @zackbrown?

@taylorpoe Yes, that sounds right.

packages/haiku-creator/src/react/components/ProjectBrowser.js
```diff
@@ -55,6 +58,10 @@ class ProjectBrowser extends React.Component {
   }
 
   componentWillReceiveProps (nextProps) {
+    if (nextProps.expiredTrialNonPro) {
```
stristr: I suggest using `if (!this.props.expiredTrialNonPro && nextProps.expiredTrialNonPro)` instead. While testing the lockout dashboard, I found myself randomly getting the modal to appear whenever props change on the `ProjectBrowser`, which is…often, because there are a lot of props.

packages/haiku-creator/src/react/components/ProjectThumbnail.js
```diff
@@ -60,7 +74,9 @@ class ProjectThumbnail extends React.Component {
             (this.state.isMenuActive || this.state.isHovered) && {opacity: 1},
           ]}
           onClick={() => {
-            if (!this.state.isMenuActive) {
+            if (this.props.expiredTrialNonPro) {
+              this.openBrowserToProjectSharePage();
+            } else if (!this.state.isMenuActive) {
```
stristr: In lockout mode, I can still click the project title beneath the thumbnail and get into the editor.

There's another `launchProjectIfAllowed` attached to the wrapper div that needs to be handled.

packages/haiku-creator/src/react/components/ProjectThumbnail.js
```diff
@@ -84,9 +100,21 @@ class ProjectThumbnail extends React.Component {
...
+            key="view-online-alt"
+            onClick={this.openBrowserToProjectSharePage}
+            style={[
+              DASH_STYLES.menuOption,
+              DASH_STYLES.opt2,
+              !this.state.isMenuActive && DASH_STYLES.gone,
+            ]}
+          >
+            View Online
+          </span>}
```
stristr, line 130: Another awkward edge: not every project has a share page. Specifically, projects that have never been published will not have share pages.

Worse, we do have an unfinished/aborted initial snapshot created by the post-receive hook, so we find ourselves stuck in this situation indefinitely: https://share-staging.haiku.ai/32793a57-d48c-4cf7-8f2e-7c3f89fbf622

This is messy…I see some quick solutions as follows:
 - Attach HasSnapshots as a nullable virtual bool (not in DB but included in JSON), and do the work of attaching it (requires writing a semi-clever `LEFT JOIN`) inside `GET /v0/project`.
 - If up for more refactoring, just attach the entire `SnapshotUrl` as a nullable virtual string, kill the new project unique ID snapshot lookup, and pass that through instead of constructing it in Plumbing.

I actually prefer option 2 (we could also dump the new `HAIKU_SHARE` env in mono—inkstone already knows how to assemble share links against the correct domain), although it has a gnarlier query.

stristr: An alternative if we want to keep it way simpler that occurred to me:

 - The presence of `repositoryUrl` on the project object from plumbing is a _pretty_ solid indicator of if a share page should exist. The edge cases this fails to catch are:
   - Projects created before we launched lazy Git init on first publish.
   - Projects whose first publish failed for some reason.
   - Forked projects and the template project.

This seems like a much smaller set of cases to fail on.

packages/haiku-creator/src/react/components/SideBar.js
taylorpoe: Got it, fixed.

packages/haiku-creator/src/react/components/ProjectBrowser.js
taylorpoe: Updated.

packages/haiku-creator/src/react/components/ProjectThumbnail.js
taylorpoe: Nice catch, fixed.

packages/haiku-creator/package.json
```diff
@@ -22,8 +22,6 @@
   "dependencies": {
     "@babel/parser": "^7.0.0",
     "@haiku/core": "4.4.0",
-    "@haiku/taylor-creatorintro": "^0.0.5",
```
taylorpoe, line 25: Had to revert (re-add) these. Although they are unused, removing them causes the app not to start with no errors showing in the terminal.

packages/haiku-creator/src/react/components/ProjectBrowser.js
taylorpoe: Fixed.

packages/haiku-creator/package.json
stristr: That shouldn't be the case! I don't have that on my checkout. Did you see this after you ran `yarn`?

taylorpoe: Yeah, even with a `yarn` the application still loads frozen without the ability to open devtools.

stristr: Eh, it’s not hurting anything to have these there, but if you want to clean up…are you by any chance on a `yarn --version` that is **not** `1.9.4`? The `integrity` flags in `yarn.lock` you added here seem to indicate you might not be, and that could be related to what's actually happening here.

taylorpoe: Yeah, I'm on `1.12.3`

stristr: Let's have you downgrade and see if you still need those rando deps, maybe?

We shouldn't have `yarn.lock` being written by multiple versions of `yarn` anyway.

taylorpoe: Still no dice. Downgraded. Closed terminal. Opened. See correct yarn version. Yarn installed. Still can't get the app working after removing these from package.json.

I think we can just skip removing them for now. Not hurting too much having them in there.

stristr: Yeah, should be fine (don't think those integrity lines choke up the older `yarn` we build with on :jenkins:, but are just ignored). Just watch that `yarn.lock` and make sure weird versions of `@haiku/core` don't sneak in, which sometimes happen when `yarn` is being weird. Everything should resolve like `version: "4.4.0"`. Looks correct right now.

packages/haiku-creator/src/react/components/ProjectThumbnail.js
taylorpoe: @zackbrown 👀

stristr: We discussed yesterday…I'll put in this wee fix.

roperzh:
@zackbrown @taylorpoe how do you guys feel about this branch? Is it ready to rebase + merge? I'm asking in preparation for the build + QA on March 4.

taylorpoe:
@roperzh LGTM from my end

roperzh:
Thanks! Merging to create a RC.

PR #982, opened 2019-02-28
feat: Timeline dnd refactor

roperzh:
Summary of changes:

- Move a bunch of components from JS to TS, cleaning a lot of unused props in the process, this should improve dev experience and add a marginal perf improvement.
- Improved DnD on the Timeline:

Regressions to look for:

- Timeline DnD in general.

Completed checkin tasks:


changelog/public/latest.json
```diff
@@ -2,7 +2,8 @@
   "sections": {
     "What's new": [
       "Better auto-complete items in Expressions.",
-      "Added placeholder support for the Vue adapter."
+      "Added placeholder support for the Vue adapter.",
+      "Better drag an drop of Timeline rows."
```
stristr: Typo: drag **and** drop.

stristr:
There seems to be a major bug in this branch where element creation/deletion is no longer updating models. I haven't dug into this branch yet to see why, but it was one of the first things I noticed.

1. When I instantiate an element on stage, its row doesn't appear in the timeline.
2. When I delete an element on stage, its row doesn't disappear in the timeline.

@roperzh, can you reproduce this?

stristr:
Another weird phenomenon I'm seeing: DnD from asset library to stage is now using "ghost of Glass". Did we perhaps have some vestigial usage of `react-dnd` in that functionality which doesn't use obvious code?

https://cl.ly/606a5c83437b

roperzh:
@stristr thanks for the help testing! I pushed a fix for the rows not being updated when an element is instantiated / deleted (it was a product of using `React.PureComponent` when the view updates from data derived from props)

>  DnD from asset library to stage is now using "ghost of Glass". Did we perhaps have some vestigial usage of `react-dnd` in that functionality which doesn't use obvious code?

I can't repro this, but I'll investigate, do you know if it happens under a certain set of circumstances?

stristr:
Yep, that fixed the big problem!

You're right, it's not happening all the time. Here is my repro. Probably not only be an issue on this branch—I'll check in a second.
 - With cursor hovering over library, `Cmd+A`.
 - Then try to dnd.

stristr:
Yep…existing issue in `master`. (I encountered trying to quickly instantiate/delete elements.) Filing and can be ignored here.

roperzh:
very good find! thanks for taking care of double checking and finding a good repro 🤗. Heads up, I pushed another fix for an issue when you drop the timeline row outside of the timeline.

stristr (commented):
Let's be sure to QA extensively prior to release, but LGTM except for typo in changelog.

roperzh:
awesome, thanks! :) merging after health check pass.

PR #983, opened 2019-02-28
fix: don't show dots on top of expression editor

roperzh:
Summary of changes:

Don't show dots on top of expression editor, they were meant to show the calculated value of the expression a la Excel, but I have found some problems and according to the task prompt:

> if super easy, show the calculated value in the box — if there are any problems, don't waste cycles: just hide.

Regressions to look for:

- None expected

Completed checkin tasks:

Not worthy of a changelog entry IMO, but let me know if you don't think so.


stristr:
LGTM, merging.

PR #987, opened 2019-03-07
fix: use pre-compiled templates in Vue adapter

stristr:
Summary of changes:

- Fixes Vue adapter when used in the context of runtime-only mode. My bad—fixes breaking change when I deployed account just now.

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -4,7 +4,7 @@
       "FIRST!"
     ],
     "Fixes": [
-      "FIRST!"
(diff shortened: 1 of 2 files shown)


roperzh:
LGTM, sorry for not catching that, it looked better :(

Should we do a new release with this?

stristr:
I don't see any need to break the release cadence…don't think this feature is highly used at the moment.

stristr:
One bug report and I might change my mind :smile:

But I think it can wait.

roperzh:
sounds good!

PR #988, opened 2019-03-08
fix: prevent crashes when trying to drag the root row

roperzh:
Summary of changes:

- Prevent the main row from being dragged

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM.

PR #989, opened 2019-03-08
fix: prevent timeline marquee from initiating from the properties panel

roperzh:
Summary of changes:

- Prevent marquee selection from starting on the properties panel by doing a quick check based on mouse position:

Regressions to look for:

- None expected.

Completed checkin tasks:


stristr:
LGTM, merging.

PR #992, opened 2019-03-11
fix: use separate keyboard history for every Timeline input field

roperzh:
Summary of changes:

This prevents inputs on the timeline from sharing the same undo/redo
buffer causing values to cross over inputs if the user undoes/redoes
multiple times.

Asana Task: https://app.asana.com/0/922186784503552/1113303258027099

Regressions to look for:

- None expected, check undo/redo of timeline inputs?

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/src/components/ExpressionInput.js
+++ b/packages/haiku-timeline/src/components/ExpressionInput.js
@@ -144,6 +144,7 @@ export default class ExpressionInput extends React.Component {
     this._paramcache = null;
     this._parse = null; // Cache of last parse of the input field
     this._mouseDownStarted = false;
+    this._historyMap = new Map();
 
     this.codemirror = CodeMirror(document.createElement('div'), {
       theme: 'haiku',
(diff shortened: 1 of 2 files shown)


packages/haiku-timeline/src/components/ExpressionInput.js
```diff
@@ -299,6 +300,8 @@ export default class ExpressionInput extends React.Component {
       return void (0);
     }
 
+    this._historyMap.set(this.props.component.getFocusedRow().getUniqueKey(), cm.getHistory());
```
stristr, line 303: Wow, was really surprised to learn this is an ever-present component and not specific to the row, making these calls necessary.

stristr:
LGTM.

packages/haiku-timeline/src/components/ExpressionInput.js
roperzh: yes :( probably the proper fix would be to make the component specific to the row, but that's a can of worms on its own.

PR #993, opened 2019-03-12
fix: improve the logic to determine if a Keyframe is a transition

roperzh:
Summary of changes:

Don't assume that because a keyframe has a curve defined in the bytecode
is a valid curve, also check:

- If the curve is a string, check if has a valid value (we consider valid
a value in the Curve enum)
- If the curve is an array

Asana Task: https://app.asana.com/0/922186784503552/1113764845664744

Regressions to look for:

- Can't think of anything

Completed checkin tasks:

### DIFF
--- a/packages/haiku-serialization/src/bll/Keyframe.js
+++ b/packages/haiku-serialization/src/bll/Keyframe.js
@@ -1,5 +1,6 @@
 const HaikuComponent = require('@haiku/core/lib/HaikuComponent').default;
 const expressionToRO = require('@haiku/core/lib/reflection/expressionToRO').default;
+const Curve = require('@haiku/core/lib/api').Curve;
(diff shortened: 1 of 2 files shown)


packages/haiku-serialization/src/bll/Keyframe.js
```diff
@@ -1,5 +1,6 @@
 const HaikuComponent = require('@haiku/core/lib/HaikuComponent').default;
 const expressionToRO = require('@haiku/core/lib/reflection/expressionToRO').default;
+const Curve = require('@haiku/core/lib/api').Curve;
```
stristr, line 3: Nit: `const {Curve} = require('@haiku/core/lib/api');`

stristr:
LGTM, merging.

PR #994, opened 2019-03-12
fix: ensure we have a focused row before setting undo/redo history

roperzh:
⚠️ plz merge before release! 🙏 

Summary of changes:

Because we aren't always updating our Timeline components via `setState`
there are race conditions generated by calling `#handleEditorChange` on
a `setState` callback + our `handleUpdate` logic, making this guard
necessary.

Note that this guard is all over `ExpressionInput` so I should have known
better in the first implementation.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #995, opened 2019-03-12
Fixes and improvements for color pickers + timeline inputs

roperzh:
Summary of changes:

- fix: only display the color picker for color strings: This disables the color picker for `linear-gradients` and similar.
- fix: smarter setting of UI fallback value for color properties: Instead of doing gymnastics with codemirror, hack the value when engaging focus.

Regressions to look for: 

- Color popovers in general

Completed checkin tasks:


stristr:
LGTM, merging.

PR #996, opened 2019-03-13
 chore: remove repeat/if from JIT menu

roperzh:
Summary of changes:

Remove repeat/if from JIT menu, and since this is the only place where the experiment was enabled, also retire it.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #997, opened 2019-03-13
 fix: use 'Position' instead of 'Translation' in the JIT menu 

roperzh:
Summary of changes:

Use 'Position' instead of 'Translation' in the JIT menu, notes:

- Because of the way I solved this, any prefix in the `PREFIX_TO_CLUSTER_NAME` object will also be affected.
- We are careful about perf when dealing with this code because is a hot path, should we worry about this change on paper?

Regressions to look for:

- Maybe the 'notes' below are potential regressions?

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -4,7 +4,8 @@
       "FIRST!"
     ],
     "Fixes": [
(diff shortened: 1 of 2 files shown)


stristr:
LGTM, merging. Not concerned about perf here.

PR #999, opened 2019-03-14
fix: don't hardcode paths in autoupdate & use ditto

roperzh:
Summary of changes:

This implements a series of refactors to autoupdate logic by:

- Letting Electron's `app.relaunch();` figure out the execution path by
themselves instead of hardcoding one. We used to provide a path to cover
for users executing the app from outside Applications folder, but since
forking we are enforcing this.
- Using the `ditto` command (present in all macOS distributions) to sync
the downloaded file with Applications, so we don't have to worry about
the app name.

*How I tested this*

- Booted up a local squirrel server that always returns a hardcoded
download link.
- Compiled and repacked (using @sasha.com's `repack` bash script) a
production app living in Applications.
- Run the repacked app, let autoupdate do it's thing
- When the app is relaunched, verify that:
  - Everything works
  - The app is the provided by squirrel.

Regressions to look for:

- Autoupdate in general

Completed checkin tasks:

### DIFF
--- a/packages/haiku-creator/package.json
+++ b/packages/haiku-creator/package.json
@@ -22,8 +22,8 @@
   "dependencies": {
     "@babel/parser": "7.0.0",
     "@haiku/core": "4.5.2",
-    "@haiku/tina-haikuerror": "^0.0.6",
     "@haiku/taylor-nocon": "^0.0.2",
+    "@haiku/tina-haikuerror": "^0.0.6",
     "async": "^2.5.0",
     "better-react-spinkit": "2.0.0-4",
     "color": "^2.0.1",
@@ -57,9 +57,11 @@
     "react-onclickoutside": "^6.4.0",
     "react-popover": "^0.4.18",
     "react-selectize": "^3.0.1",
-    "react-transition-group": "^2.2.1"
+    "react-transition-group": "^2.2.1",
+    "uuid": "^3.2.1"
   },
   "devDependencies": {
+    "@types/uuid": "^3.2.1",
     "@types/fs-extra": "^4.0.8",
     "@types/node-fetch": "^2.1.1",
     "@types/qs": "^6.5.1",
--- a/packages/haiku-creator/src/utils/autoUpdate.ts
+++ b/packages/haiku-creator/src/utils/autoUpdate.ts
@@ -8,7 +8,7 @@ import nodeFetch from 'node-fetch';
 import * as os from 'os';
 import * as path from 'path';
 import * as qs from 'qs';
-import * as uuid from 'uuid';
+import {v4} from 'uuid';
 
 const DEFAULT_OPTIONS = {
   server: process.env.HAIKU_AUTOUPDATE_SERVER,
@@ -38,8 +38,8 @@ export default {
       }
 
       const tempPath = os.tmpdir();
-      const zipPath = path.join(tempPath, `${uuid.v4()}.zip`);
-      const extractPath = path.join(tempPath, uuid.v4());
+      const zipPath = path.join(tempPath, `${v4()}.zip`);
+      const extractPath = path.join(tempPath, v4());
       const appPath = path.resolve(electron.remote.app.getPath('exe'), '..', '..', '..');
       logger.info('[autoupdater] About to download an update:', options, url);
       await download(url, zipPath, progressCallback);
--- a/packages/haiku-creator/src/utils/autoUpdate.ts
+++ b/packages/haiku-creator/src/utils/autoUpdate.ts
@@ -1,11 +1,14 @@
 import * as electron from 'electron';
+import * as fs from 'fs';
 // @ts-ignore
-import {download, unzip} from 'haiku-serialization/src/utils/fileManipulation';
+import {ditto, download, unzip} from 'haiku-serialization/src/utils/fileManipulation';
 // @ts-ignore
 import * as logger from 'haiku-serialization/src/utils/LoggerInstance';
 import nodeFetch from 'node-fetch';
 import * as os from 'os';
+import * as path from 'path';
 import * as qs from 'qs';
+import * as uuid from 'uuid';
 
 const DEFAULT_OPTIONS = {
   server: process.env.HAIKU_AUTOUPDATE_SERVER,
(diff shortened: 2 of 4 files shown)


packages/haiku-creator/src/react/Creator.js
```diff
@@ -2081,12 +2081,12 @@ export default class Creator extends React.Component {
             startTourOnMount={true}
           />
         }
-        <AutoUpdater
+        {this.envoyClient && this.envoyProject && this.state.username && <AutoUpdater
```
roperzh: I would really appreciate a sanity check on this, are we leaving out non-authenticated users? The whole reason of doing this is because if squirrel is fast enough we can end with multiple downloads running in parallel as Creator re-renders itself multiple times on start.

packages/haiku-creator/package.json
```diff
@@ -22,8 +22,9 @@
   "dependencies": {
     "@babel/parser": "7.0.0",
     "@haiku/core": "4.5.2",
-    "@haiku/tina-haikuerror": "^0.0.6",
     "@haiku/taylor-nocon": "^0.0.2",
+    "@haiku/tina-haikuerror": "^0.0.6",
+    "@types/uuid": "^3.2.1",
```
stristr: `devDependencies`?

packages/haiku-creator/src/react/Creator.js
stristr: I was little confused by this message until I looked into how the AutoUpdater component is implemented…React render cycles shouldn't be capable of accidentally running multiple downloads in parallel and this problem arises because we are giving DAL responsibilities to a React component and binding things that should be singletons (like "is currently downloading an update") as React states. None of this should be the case. Really, this whole routine should just run from Envoy and the React component should at most query/receive its state from the server.

Is there actually a chance of the `<AutoUpdater />` component itself being created, uncreated, and created again in idle state? If not, I'm not sure we need to fix this. If we need to fix it, though, I think we should do it with a proper mutex and not at the mercy of the React props game. I would not assume this is a sufficient fix.

roperzh: This makes a lot of sense, I'll double check if we really need to do this and fix if so.

roperzh: I removed this and created an [Asana ticket](https://app.asana.com/0/922186784503552/1114550475268584) to refactor this appropriately.

roperzh:
This is ready for another review.

packages/haiku-creator/src/utils/autoUpdate.ts
```diff
@@ -22,42 +24,32 @@ export interface CheckResult {
...
+      const zipPath = join(tempPath, `${uuid.v4()}.zip`);
+      const extractPath = join(tempPath, uuid.v4());
+      logger.info('[autoupdater] About to download an update:', options, url);
+      await download(url, zipPath, progressCallback);
+      // `unzip` first, you can unzip in `ditto` by providing the `-xk` flags, but trying to target `/Applications`
+      // throws permission errors.
+      await unzip(zipPath , extractPath);
+      // `ditto` the contents of the extract path folder (the .app package) into `/Applications` by syncing
+      // folder-to-folder we don't have to mind about the .app name
+      await ditto(extractPath, '/Applications');
```
stristr: Ah, I think we do want to worry about the `.app` name.

For example, say you are running out of `/Applications/Haiku Animator.app` and you have unpacked the zipfile to `/tmp/…/Haiku Animator X.app`.

When you `ditto` from `/tmp/...` to `/Applications`, I believe you would have `/Applications/Haiku Animator.app` _and_ `/Applications/Haiku Animator X.app`, but the `relaunch()` will open `Haiku Animator.app`, and you'll get in an update loop.

Ideally we'd never have to hack this again if we want to change the name of the app in the future.

stristr: (ping @roperzh)

roperzh: sorry for missing this, sending a fix soon.

packages/haiku-creator/src/utils/autoUpdate.ts
```diff
@@ -1,11 +1,14 @@
...
 // @ts-ignore
-import {download, unzip} from 'haiku-serialization/src/utils/fileManipulation';
+import {ditto, download, unzip} from 'haiku-serialization/src/utils/fileManipulation';
 // @ts-ignore
 import * as logger from 'haiku-serialization/src/utils/LoggerInstance';
 import nodeFetch from 'node-fetch';
 import * as os from 'os';
+import * as path from 'path';
 import * as qs from 'qs';
+import * as uuid from 'uuid';
```
stristr: As a general pattern, just to get us in the habit, let's `import {v4} from 'uuid';` when our intent is to use one function.

packages/haiku-creator/src/utils/autoUpdate.ts
```diff
@@ -22,42 +25,40 @@ export interface CheckResult {
...
+  async update (url: string, progressCallback: (progress: number) => void, options = DEFAULT_OPTIONS) {
+    if (process.env.HAIKU_SKIP_AUTOUPDATE !== '1') {
+      if (
+        !options.server ||
+        !options.environment ||
+        !options.branch ||
+        !options.platform ||
+        !options.version
+      ) {
+        throw(Error('Missing release/autoupdate environment variables'));
```
stristr: I didn't know you could do this! (I checked and it's not in [the spec](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-throw-statement).)

For consistency with surrounding code, let's use `throw new Error(...)` here and below.

packages/haiku-serialization/src/utils/fileManipulation.js
```diff
@@ -52,6 +52,18 @@ module.exports = {
     });
   },
 
+  ditto (src, dest) {
+    const saneSrc = JSON.stringify(src);
+    const saneDest = JSON.stringify(dest);
+    const dittoComand = `ditto ${saneSrc} ${saneDest}`;
```
stristr: Recommended: `/usr/bin/ditto`.

Same should go for `unzip`, really. This is just as a defensive principal ("we mean the one that comes installed with macOS"), because we should in general beware weird `$PATH` corruption, permissions, and whatever Homebrew did to a given user's system.

roperzh:
@stristr you are a wise man, thanks! Tested E2E again and works, so this is ready for another review :)

stristr:
LGTM, merging.

PR #1001, opened 2019-03-15
fix: retire OutliningElementsOnStage experiment and stop trashing the…

roperzh:
… update loop

Summary of changes:

Possible fix to stop thrashing the update/sync loop, after investigation I arrived to the conclusion that this code was implemented only to send hover events to glass for the `OutliningElementsOnStage` experiment.

I also pushed #1000 which alternatively just disables the damaging code under the experiment flag. 

(I like this one more!)

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #1003, opened 2019-03-17
feat: improve position calc and add draggagability to Bezier Editor

roperzh:
Summary of changes:

This refactors the hierarchy used to display the `BezierEditor` popup,
almost all of the changes are made to components used to wrap the Editor
so I don't expect any changes in its behavior.

Things to note:

- We are attaching drag events to `window` instead of attaching them to
the wrapper of the Bezier Editor, this allows us to 1) be more flexible
about the component hierarcy and 2) fix possible edge cases.
- `Popover` in `ui-common` is removed in favor of consolidating all the
logic into `BezierPopup`, in this way we are removing unnecessary tension
product of both components sharing responsibilities.
- Use the `react-draggable` library already present everywhere on the
Timeline to add dnd support to the Editor
- Improve the logic to set the bounds of the Editor.

Regressions to look for:

- Bezier editor positioning

Completed checkin tasks:


packages/haiku-ui-common/src/react/Bezier/BezierEditor.tsx
```diff
@@ -217,14 +220,16 @@ export default class BezierEditor extends React.Component<BezierEditorProps> {
...
-        };
+    if (readOnly || !down) {
+      window.removeEventListener('mousemove', this.onDownMove);
+      window.removeEventListener('mouseup', this.onDownUp);
+      window.removeEventListener('mouseleave', this.onDownLeave);
+    } else {
+      window.addEventListener('mousemove', this.onDownMove);
+      window.addEventListener('mouseup', this.onDownUp);
+      window.addEventListener('mouseleave', this.onDownLeave);
+    }
```
stristr, line 245: qq @roperzh: do we also need to worry about `blur` events here given the webviewness? I am wary of creating stuck drags whenever I see code like this living inside React listeners…there are a few unconventional ways to eject from React and get stuck.

roperzh: @stristr I tried with `cmd+tab`, `cmd+tab+click on icon` (this gave us problems in the past) and by moving/releasing the mouse between webviews and outside of the app window and works fine! do you have other ideas/suggestions in mind?

stristr: Hmmm, probably, but if it's not easy to trigger an error while testing, then we're probably only dealing with edge cases anyway…

Can we at least be sure that we unmount these event listeners when the component unmounts?

stristr:
LGTM.

packages/haiku-ui-common/src/react/Bezier/BezierEditor.tsx
roperzh: that's an excellent idea, I pushed that along other improvements.

roperzh:
@stristr I pushed another set of changes, do you want to check again or should I just merge?

stristr:
Go ahead. LGTM still!

roperzh:
Thanks!

PR #1004, opened 2019-03-19
fix: prevent crashes when executing `cmd+key` during stage drag

roperzh:
Summary of changes:

This fixes a [crash][1] occasionated when executing a command during a
drag, for example by pressing `cmd + A` while an element is being dragged.

This involves two main things:

- Virtually reverting back PR #831 (2252e6a) which introduced part of the
faulty logic in order to solve a [bug][2].
- By taking adventage of the `Globals` object instead of setting an
internal value to track if the `cmd` key is down, which generated problems
because many of the functions are throttled.

Asana Task: https://app.asana.com/0/922186784503552/958155168927487

[1]: https://sentry.io/organizations/haiku-systems/issues/824060464/?project=226387&referrer=asana_plugin&statsPeriod=14d&utc=false
[2]: https://app.asana.com/0/813447917062242/827578786675323

Regressions to look for:

- Glass dnd, and selection

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -5,7 +5,8 @@
     ],
     "Fixes": [
       "Removed `Control Flow > If` and `Control Flow > Repeat` from the Timeline 'ADD +' menu.",
-      "Reworded 'Translation' to 'Position' in the Timeline 'ADD +' menu."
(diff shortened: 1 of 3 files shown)


stristr:
Seems like it should work (haven't tested). LGTM, merging.

PR #1005, opened 2019-03-19
 fix: improve the logic to calc if a Keyframe `hasCurveBody` 

roperzh:
Summary of changes:

This improves two methods:

- `hasCurveBody` to understand that not all curve strings are valid
- `hasDecomposableCurve` to first check if the keyframe `hasCurveBody`
before trying to infer if the curve is descomposable or not.

This will prevent crashes like the one in the associated task which can
be reproduced by setting a curve to an invalid string while the Bezier
Editor is open

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #1006, opened 2019-03-19
fix: disable `user-select` for child elements of RowManager

roperzh:
Summary of changes:

This fixes a drag ghost randomly appearing when draging keyframes, caused
because the text of the Rows was being selected (though not visibly)
triggering native drag and drop behaviors.

While the dragging issue is difficult to repro, you can notice this
behavior if you:

- Click anywhere on the Timeline.
- Check that `window.getSelection().toString()` is "".
- Drag a keyframe in a downward motion.
- Check that `window.getSelection().toString()` has contents, this
causes the drag ghost to randomly appear.

Asana Task: https://app.asana.com/0/922186784503552/1113905370581295

Regressions to look for:

- None expected

Completed checkin tasks:

### DIFF
--- a/packages/haiku-timeline/public/styles/Timeline.css
+++ b/packages/haiku-timeline/public/styles/Timeline.css
@@ -194,11 +194,6 @@ input:focus, button:focus {
   height: 100%;
 }
 
-#timeline .no-select {
-  -webkit-user-select: none;
-  cursor: default;
-}
-
 #timeline .no-incrementor input[type=number]::-webkit-inner-spin-button,
 input[type=number]::-webkit-outer-spin-button {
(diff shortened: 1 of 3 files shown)


packages/haiku-timeline/public/styles/Timeline.css
```diff
@@ -194,11 +194,6 @@ input:focus, button:focus {
   height: 100%;
 }
 
-#timeline .no-select {
-  -webkit-user-select: none;
-  cursor: default;
-}
-
```
roperzh, line 201: This was removed because it's duplicated down below (not visible on diff)

stristr:
LGTM, merging.

PR #1010, opened 2019-03-20
feat: remove tween before keyframes if a single tween is selected

roperzh:
Summary of changes:

Click a tween. Press 'backspace' (delete), tween is removed. [leave keyframes selected; delete those upon subsequent backspace]

Regressions to look for:

- None, really, but I noticed some weirdness on the Timeline while testing this, and I found that's because a very very old [underlying issue](https://app.asana.com/0/922186784503552/1115259939111443), which I'm working on #1009. My initial plan was to include this alongside the fix, but I think this is better.

Completed checkin tasks:

### DIFF
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -2,7 +2,8 @@
   "sections": {
     "What's new": [
       "Added the ability drag and reposition the Bezier Editor.",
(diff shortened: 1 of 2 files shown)


stristr:
LGTM, merging. Great surgical solution.

PR #1012, opened 2019-03-27
 fix: add a missing key in order to make Radium happy and prevent crashes

roperzh:
Summary of changes:

As the title says, this one took a while because of a combination of how Radium reports this error (by the name of the component) + uglyfication in prod without source maps, causing the error to be:

> Radium requires each element with interactive styles to have a unique key, set using either the ref or key prop. Multiple elements have no key specified. Component: "n". Element: "div".

Turns out that  the error is in the offline screen, this fixes it and adds a couple of extra small improvements here and there.

Regressions to look for:

- None expected

Completed checkin tasks:


stristr:
LGTM, merging.

PR #1016, opened 2019-04-08
fix: bad logic breaking main row normal collapsed behavior

roperzh:
Summary of changes:

When we refactored DnD of timeline rows, we introduced a new helper
prop `forceCollapse` in order to force the rows to be collapsed when
a drag is being performed, but since what we really want is to collapse
_all rows except Main_ I introduced some faulty logic to handle this
specific scenario.

This fixes this logic by not assuming that the Main row has to be always
expanded.

Asana Task: https://app.asana.com/0/1114158148688266/1117678624456373

Regressions to look for:

- DnD, collapsing & expanding rows.

Completed checkin tasks:


stristr:
LGTM.

PR #1022, opened 2020-04-07
Fix (Windows): Figma login in development

gobaldia:
Summary of changes:

In Windows, in development, there's not possible to login in Figma. To use Figma, it's needed to set an environment variable with the Figma token.

Completed checkin tasks:

- [ ] Updated `changelog/public/latest.json`.


roperzh (approved):
LGTM, merging.

PR #1023, opened 2020-04-08
Changed open in finder/file explorer label

gobaldia:


packages/haiku-common/src/electron/TopMenu.ts
```diff
@@ -90,7 +91,7 @@ export default class TopMenu {
 
     const developerMenuItems = [
       {
-        label: 'Open in Finder',
+        label: isWindows ? 'Open in File Explorer': 'Open in Finder',
```
roperzh: @gobaldia `isWindows` is a function, this is probably working in Windows because it's a truthy value

```suggestion
        label: isWindows() ? 'Open in File Explorer' : 'Open in Finder',
```

roperzh (approved):
LGTM.

PR #1025, opened 2020-04-08
Hid sketch references for Windows

gobaldia:


packages/haiku-creator/src/react/components/library/AssetItem.js
```diff
@@ -386,7 +386,7 @@ class AssetItem extends React.Component {
       );
     }
 
-    if (this.props.asset.isSketchFile()) {
+    if (isMac() && this.props.asset.isSketchFile()) {
```
roperzh: @gobaldia here and below, feels like we could modify `isSketchFile` to also check for `isMac`, or even better create a new method with a better name that performs both checks

gobaldia: @roperzh I decided to include `isMac()` in all those places instead of putting it in `isSketchFile()` for the sake of readability. Now, after your review, I decided to create a function that checks if it's needed to consider Sketch (this is that is Sketch and you're in Mac).
Tell me what's your feeling about it.

roperzh: yes, I think that a new function is perfect (we will probably use that in other places). We might end up having to abstract that function into `Asset`, but let's not do it until we need it.

roperzh (approved):
LGTM

PR #1028, opened 2020-04-09
Fixed problem with Windows paths

gobaldia:


roperzh (approved):
ahh we had a similar problem in Diez not so long ago. LGTM

PR #1035, opened 2020-04-15
fix: align post keyframes vertically

gobaldia:
This PR fixes a visual issue that was duplicating keyframes in the timeline. The following images show how it was before the fix and how it's now.

Before:
image

After:
image


roperzh (approved):
LGTM!

PR #1036, opened 2020-04-16
fix(windows): add autoupdate and fix crashes

roperzh:
Summary of changes:

- Build for 64 and 32 bits on Windows
- Fix crashes
- Add autoupdate logic
- Lint fixes


gobaldia (approved):
LGTM

PR #1037, opened 2020-05-05
tweak: remove intercom button/widget

roperzh:
This commit removes the intercom button in a single and revertible commit.


gobaldia (approved):
LGTM

PR #1040, opened 2020-09-08
Changed roles accepted as PRO

gobaldia:
Summary of changes:

- Fixed a bug that was preventing a Subsidiary PRO user to use the app

Regressions to look for:

- …

Completed checkin tasks:

### DIFF
--- a/changelog/public/5.1.0.json
+++ b/changelog/public/5.1.0.json
@@ -1,0 +1,8 @@
+{
+  "sections": {
+    "Fixes": [
+      "Fixed issues with Illustrator integration on Windows.",
+      "Fixed issues importing example design files in new projects."
+    ]
+  }
+}
--- a/changelog/public/latest.json
+++ b/changelog/public/latest.json
@@ -1,7 +1,5 @@
 {
   "sections": {
-    "What's new": [
-    ],
     "Fixes": [
       "Fixed issues with volume licensing."
     ]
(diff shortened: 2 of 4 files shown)


roperzh (commented):
Thanks for tracking this down! despite the simplicity of the changes I know that finding them must have been a great deal of work!

One question and one request:

- Question: have you tried to simulate in some way or maybe use a production user to test if this is all we need to do?
- Request: could you revert the changelog change and use [this script instead](https://github.com/HaikuTeam/mono/blob/5ecee212bead0eee96b33c3e259f317f030db0fe/scripts/rotate-changelog.js)? this way we don't lose the changes of the current version.

gobaldia:
> Question: have you tried to simulate in some way or maybe use a production user to test if this is all we need to do?

Yes, I simulated this situation with local users. I didn't with production ones.

> Request: could you revert the changelog change and use this script instead? this way we don't lose the changes of the current version.

Sure! It's already changed.

Could you please review it again @roperzh?

roperzh (approved):
LGTM!

gobaldia:
@roperzh I'll wait to merge this until I finish the implementation of https://app.asana.com/0/1175500332079249/1192850070200531.

Even though both tasks are not directly related, I'd like to test this task again when I have the another one implemented.

roperzh:
sonds good! :haikudos:

